PR #16, opened 2016-03-14
Add an option to specify a custom favicon

kcgthb:
Add an extra setting in `mkdocs.yml` to specify a custom favicon:
```
extra:
  favicon: favicon.ico
```

### DIFF
--- a/src/base.html
+++ b/src/base.html
@@ -58,7 +58,7 @@
     {% endif %}
 
     <!-- Favicon -->
-    {% set icon = icon | default("assets/images/favicon.ico") %}
+    {% set icon = config.extra.favicon | default("assets/images/favicon.ico") %}
     <link rel="shortcut icon" type="image/x-icon"
       href="{{ base_url }}/{{ icon }}" />
     <link rel="icon" type="image/x-icon"


squidfunk:
Basically, this functionality is already provided by MkDocs:
http://www.mkdocs.org/user-guide/configuration/#site_favicon

However, the template refers to `icon`, not `favicon` in the template, so correcting this would be the right way to go. I will close this pull request and provide a fix instantly.

squidfunk:
Fixed in master, will be part of 0.2.2

kcgthb:
Thanks.
Except it looks like the MkDocs devs defer to theme customization for this an plan to remove the `site_favicon` entirely: https://github.com/mkdocs/mkdocs/issues/725

d0ugal:
Yeah, it's weird having a setting for something so small and specific. We won't remove it for some time tho' - no idea when 1.0 will happen and we will want to have a clear migration path.

squidfunk:
Well, it depends. I think the favicon is something like the name of your project, so for me it makes sense to be part of the core MkDocs configuration. However, the favicon is only one of many customization options. You also would want to configure `og:image` and a web application icon for iOS, so maybe it makes sense to leave it up to the theme.

PR #17, opened 2016-03-15
Update drawer.html // Styling version number

sebastian-marinescu:
Separate version-number into span so styling it via CSS is possible

### DIFF
--- a/material/drawer.html
+++ b/material/drawer.html
@@ -8,7 +8,7 @@
         </div>
       {% endif %}
       <div class="name">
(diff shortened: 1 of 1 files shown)


squidfunk:
Oh, i saw you did your changes in `material/drawer.html`. However, this file is auto-generated from `src/drawer.html`, so it will be overwritten when the theme is built. I will add it to `src/drawer.html`.

sebastian-marinescu:
Okay, thanks. I know better now for the next time :+1:

PR #20, opened 2016-03-17
Fixing table border when not used as the first row but the first column

groud:
When the <th> is used as a column header, the first cell of each row appears rounded.
This fixes the problem.

However, as the table already has a "rounded corner attribute", I am no even sure that such rules are necessary. What do you think ?

### DIFF
--- a/src/assets/stylesheets/modules/article/_appearance.scss
+++ b/src/assets/stylesheets/modules/article/_appearance.scss
@@ -93,20 +93,6 @@
     th {
       background: mix($primary, $white, 75%);
       color: $white;
-
-      /*
-       * Round upper left border
-       */
-      &:first-child {
-        border-top-left-radius: 3px;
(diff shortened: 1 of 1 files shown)


squidfunk:
Can you provide some example code for which the theme breaks down? This makes it easier for me to verify the issue.

groud:
Here it is:
screenshot

squidfunk:
Thanks, but I meant the **markdown code** (or HTML) which generates the table - would be very helpful.

groud:
Ah sorry, I missed the word "code". :)
[bug.txt](https://github.com/squidfunk/mkdocs-material/files/178285/bug.txt)

squidfunk:
You we're right - the `border-radius` on the `th`s was not necessary - I added `overflow: hidden` to the table and now everything should work as expected.

PR #37, opened 2016-07-02
Fix inline code snippets in list

bborysenko:
A small fix to not wrap inline code snippets in list:


squidfunk:
Thank you, looks good to me. However, please also provide the final build located in the `material/` folder. Just execute `gulp build --production`, and add the resulting files (should be the `base.html`, `manifest.json`, and two css files).

squidfunk:
Closed due to inactivity.

PR #49, opened 2016-09-28
Adding social link: linkedin

urosjarc:
I have added linkedin social links to header and drawer, but @squidfunk , I need some help with linkedin icon. The font awesome has linkedin icon but I didn't want import whole library in header, I saw that you make icons with iconmoon... but when I have generate ttf,svg etc the zip file didn't contain json file representing icons so I have joust leave it as it is and set `icon-linkedin` content to twitter utf-8 code.

@squidfunk Can you pls include [linkedin icon](http://fontawesome.io/icon/linkedin-square/) in font bundle, and update [icon-linkedin content](https://github.com/squidfunk/mkdocs-material/compare/master...urosjarc:feature/linkedin?expand=1#diff-7dea3af3daf6b5638e0dffdb3ebc9284R108)
## Header section

Header section
## Drawer section

Drawer section

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ site
 # Distribution files
 dist
 mkdocs_material.egg-info
+
+# IDE files
+.idea
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -43,6 +43,7 @@ extra:
   author:
     github: squidfunk
     twitter: squidfunk
+    linkedin: squidfunk
 
 # Extensions
 markdown_extensions:
--- a/src/assets/stylesheets/fonts/_icon.scss
+++ b/src/assets/stylesheets/fonts/_icon.scss
@@ -102,6 +102,13 @@
 }
 
 /*
+ * Linkedin icon
+ */
+.icon-linkedin:before {
+  content: "\e606";
+}
+
+/*
  * Twitter icon
  */
 .icon-twitter:before {
--- a/src/assets/stylesheets/modules/base/_layout.scss
+++ b/src/assets/stylesheets/modules/base/_layout.scss
@@ -283,6 +283,7 @@ hr {
 /*
  * Hide source and twitter button
  */
+.button-linkedin,
 .button-github,
 .button-twitter {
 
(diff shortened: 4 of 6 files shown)


digitalcraftsman:
Would it be possible to add social icons for more popular social media sites in general?

urosjarc:
@digitalcraftsman Make a list of social sites that you want to have and I will make it happend.

squidfunk:
It's possible to add social icons, yes, but no priority for 0.2.x, because I'm near to finishing 1.0.0 which gets rid of the social icons within the header bar. I will leave it open for a successor version of 1.0.0. Personally I think it's better to include them in the footer.

urosjarc:
@squidfunk I joust want to say that this mkdocs template is awesome! This is thank you gift for you:

[](https://postimg.org/image/gfy2mczol/)

squidfunk:
Hahahaha thank you, this is really awesome ;-)

digitalcraftsman:
Here are a list of icons that could be related to the documentation of a (software) project:

- GitLab
- BitBucket
- StackOverflow (e.g. to link questions with a tag)
- Slack (e.g. for a support chat)
- Email

Perhaps we could also add

- YouTube
- Vimeo
- Facebook
- Google Plus

But at the end it's up to @squidfunk who decides what we wants to support in his awesome theme

makshh:
Maybe SVG something like this? 

[Codepen](http://codepen.io/ruandre/pen/howFi)

squidfunk:
@digitalcraftsman Good list! I will prepare something for 1.0.x, so we can add all those icons in the footer. Already did Bitbucket and GitLab integration, the rest also sounds legit.

squidfunk:
So I think the best idea would be to integrate this icon font I found:
http://www.socicon.com/

Seems to cover pretty everything. My idea would be to automatically integrate it, when there are any social links listed in the `extra` configuration, but leave it out otherwise to minimize the footprint.

squidfunk:
Fixed in #e08802aed16aa5e0c0cf11c499a48a9405af42d5. The theme now uses http://www.socicon.com/ for social icons. The syntax in the `mkdocs.yml` is:

``` yaml
extra:
  social:
    - type: github
      link: https://github.com/squidfunk
```

The value specified in `type` is appended to the `socicon` class name.

Issue #53, opened 2016-10-11
How Can I Remove the Footer "Docs – Documentation built with MkDocs using the Material theme."


ahmadawais:
## Description
I am looking into How Can I Remove the Footer "Docs – Documentation built with MkDocs using the Material theme." Can anyone let me know? A quick search didn't turnup anything

squidfunk:
It's hardcoded in the template. In order to remove it, you have to fork the theme and adjust the file base.html. Though the license says that you can do anything with it, it's probably fair to keep the note as a "thank you" ;-) But in the end it's completely up to you!

ahmadawais:
@squidfunk While creating documentation of simple users, it is hard for them to have so many distractions. Which is why it won't be a good idea to keep it there. Whereas, I have no problem with attribution in the source code. Also, SEO wise it is a death sentence to your products since Penguin doesn't like such links. Research on "widget link and penguin"

ahmadawais:
@squidfunk May I ask how will I install the forked theme?

squidfunk:
Oh okay, thank you for this clarification on widget links. I will think about it for 1.0.0.

You can fork the theme, make your adjustments, then clone it into your docs root and use the `docs_dir` attribute in your `mkdocs.yml`:

``` yaml
docs_dir: 'mkdocs-material/material'
```

ahmadawais:
@squidfunk Thanks for that.

I forked and changed the compiled theme in folder `material` base.html file. 
Then in `mkdocs.yml` I defined a custom theme.
`theme_dir: material/material`

That worked :)

ahmadawais:
@squidfunk Has the option to remove the footer attribution still not added? I thank you for your efforts but in some client projects, I can't use your link in the footer. Having a configurable option would help me keep the theme updated.

squidfunk:
Nope, there's no option, I'm sorry. However, overriding the footer is really simple:
https://github.com/squidfunk/mkdocs-material/issues/373

squidfunk:
Removing the generator notice is now supported in [Insiders](https://squidfunk.github.io/mkdocs-material/setup/setting-up-the-footer/#remove-generator) 🚀 

Just add this to `mkdocs.yml`:

``` yaml
extra:
  generator: false
```

PR #54, opened 2016-10-11
Better Attrtibution

ahmadawais:
Better attribution SEO wise is the one which is inside the code and doesn't link back. As discussed in #53 issue. Also, SEO wise it is a death sentence to your products since Penguin doesn't like such links. Research on "widget link and penguin"


squidfunk:
Won't accept this PR for now. However, I might make the note for mkdocs-material a little more decent.

Issue #46, opened 2016-09-11
Status of 1.0.0


squidfunk:
This issue is meant for updates and discussions on the progress of the rework of the theme. Please feel free to checkout and try the rework branch and comment on the current state of development.

**1.0.0-beta**
- [x] Implement search modal
- [x] Refactor JavaScript
- [x] Integrate Google Analytics from 0.2.x
- [x] Integrate palettes from 0.2.x
- [x] Refactor table styles from 0.2.x
- [x] Create specimen page and check for missing styles
- [x] Add `preventDefault` on overlay for iOS when menu is open
- [x] Refactor search result code in SCSS

**1.0.0-rc**
- [x] Put path into header (breadcrumbs)
- [x] Constrain image width to content area (like in 0.2.4)
- [x] Account for height change after parsing MathJax
- [x] Fix bug with non-stretching container when there is no content
- [x] Refactor footer to be variable-size for arbitrary footer content
- [x] Refactor link in footer (together with social icons)
- [x] Include social icons in footer (see #49)
- [x] Check all issues in original Material theme and ensure they are fixed in rework
- [x] Full integration with pymdown extensions
- [x] Write documentation on pymdown extensions
- [x] Fix anchor offsets for blurring
- [x] Add `-webkit-overflow-scrolling` via JavaScript on navigation
- [x] Add "Edit on GitHub" Link (dependent on mkdocs 0.16 release)
- [x] Make logo/icon configurable (webfont, svg)

**1.0.0**
- [x] Rewrite getting started guide
- [x] Write documentation on theme extension and customization
- [x] Write documentation on how to integrate theme from GitHub
- [x] Write CONTRIBUTING.md

**1.0.x**
- [x] Switch to PR-based development workflow after big merge
- [ ] Write unit tests with karma for components
- [ ] Refactor search algorithm
- [ ] Add a separate homepage/langing page layout
- [ ] Refactor 404 template
- [ ] Fix missing repaint on header/nav for tablet breakpoint

**DONE**
- [x] Introduced Webpack for more sophisticated JavaScript bundling
- [x] Introduced ESLint and SassLint for code style checks
- [x] Introduced more accurate Material Design colors and shadows
- [x] Introduced modular scales for harmonic font sizing
- [x] Introduced git-hooks for better development workflow
- [x] Rewrite of CSS using the BEM methodology and SassDoc guidelines
- [x] Rewrite of JavaScript using ES6 and Babel as a transpiler
[...]

PR #56, opened 2016-10-12
Minor documentation updates

brendo:
Hope you don't mind, but I thought this might pay forward in some small way to #46!

- Removed reference to `bower` as it's no longer part of the process
- Couple of newlines as per `.editorconfig`


squidfunk:
Thanks you for your contribution, however personally I think that prepending the $ breaks the ability to simply copy and paste the commands and removing the # breaks the syntax highlighting, so sorry, I'm not a fan.

Regarding the newlines: yes, those should be included in the files and where missing in my editor (clumsy me). I have a rewrite of the documentation on my list as a lot of stuff changed :-)

brendo:
No worries! Would you prefer I pair this PR back to just focus on the `bower` and newlines or simply close it?

squidfunk:
@brendo would be awesome, then I could merge it directly.

brendo:
Done :+1:

PR #58, opened 2016-10-23
Use `extra.version` if it is available when generating source partial

brendo:
A very small change remove the hardcoded version in `source.html`. It now uses `extra.version` set in the `mkdocs.yml` config.

In future, it may be possible to retrieve this information from the Github API along with the stargazers and fork counts.

### DIFF
--- a/material/partials/source.html
+++ b/material/partials/source.html
@@ -12,7 +12,9 @@
   <div class="md-source__repository">
     {{ repo_name }}
     <ul class="md-source__facts">
-      <li class="md-source__fact">v0.2.4</li>                             
+      {% if config.extra.version %}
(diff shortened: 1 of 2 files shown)


squidfunk:
I was thinking of making this entirely dynamic, because it makes no real sense when we are loading changes from the GitHub API anyway. There are already traces of some prototyping in application.js which I have to refactor.

brendo:
That's true, but for Gitlab and Bitbucket sources it will be difficult to retrieve that data as they don't expose public API's in the same way that Github does.

squidfunk:
Okay, I will merge it and check the GitLab and BitBucket APIs later. Thanks for your contribution!

PR #59, opened 2016-10-23
Abstracts fact fetching for Github sources

brendo:
This builds on #58, by introducing a `GithubSourceFacts` class which will use the Github API to retrieve the information. This is largely a refactor of existing code, with the additional of implementing `sessionStorage` caching.


brendo:
@squidfunk Sorry, this seems to have a couple more files than what I was expecting. What flags are you using with `gulp` prior to commit?

I noticed release fetching in your latest commit. If you're happy with the approach in this PR I could roll that in as well. Up to you!

squidfunk:
You should run the build before the commit. Only the scripts in `package.json` should be used:

```
npm run build
```

I will rework the documentation to reflect that and add a `CONTRIBUTING.md` However, as everything's in progress, it's doesn't really matter.

I'm sorry for all the mess in `application.js`, I'm still in prototyping-mode. Everything will be cleaned up or the release. Thanks for abstracting this stuff.

squidfunk:
BTW: good idea on using the session storage for the repository facts - was on my internal list. Did I mention it somewhere here or was that your own idea?

brendo:
That was your idea, I stumbled across the to-do while abstracting the logic. Thanks for the merge!

squidfunk:
Awesome! Thanks for going ahead and doing it.

PR #60, opened 2016-10-31
Add lint to Gulp, break it out as a seperate task

brendo:
This small PR exposes the existing lint tasks as a dedicated Gulp task so that we can easily check lint during development.

I've also fixed up the failing files so Travis should be happy.


squidfunk:
I did exactly the same last night - I put the scripts into `bin`, because they will get longer and then it will become unreadable very quickly. Also added a lint script which uses eslint and sass-lint directly. This is faster than invoking Gulp, because Gulp first needs to transpile all tasks. IMHO it's better we leave it like that.

brendo:
Nice!

Issue #63, opened 2016-11-20
"mkdocs serve" shows a bunch of deprecations


madrus:
## Description

Adding `theme : 'material'` instead of `'readthedocs'` to `mkdocs.yml` shows a bunch of warnings about deprecations.

### Expected behavior

No warnings

### Actual behavior

λ mkdocs serve                                                                                                                    
INFO    -  Building documentation...                                                                                              
INFO    -  Cleaning site directory                                                                                                
WARNING -  Your theme does not appear to contain a 'main.html' template. The 'base.html' template was used instead, which is deprecated. Update your theme so that the primary entry point is 'main.html'.                                                          
WARNING -  Template variable warning: 'site_name' is being deprecated and will not be available in a future version. Use 'config.site_name' instead.                                                                                                                
WARNING -  Template variable warning: 'favicon' is being deprecated and will not be available in a future version. Use '{{ base_url }}/img/favicon.ico' instead.                                                                                                    
WARNING -  Template variable warning: 'repo_url' is being deprecated and will not be available in a future version. Use 'config.repo_url' instead.                                                                                                                  
WARNING -  Template variable warning: 'page_title' is being deprecated and will not be available in a future version. Use 'page.title' instead.                                                                                                                     
WARNING -  Template variable warning: 'canonical_url' is being deprecated and will not be available in a future version. Use 'page.canonical_url' instead.                                                                                                          
WARNING -  Template variable warning: 'copyright' is being deprecated and will not be available in a future version. Use 'config.copyright' instead.                                                                                                                
WARNING -  Template variable warning: 'content' is being deprecated and will not be available in a future version. Use 'page.content' instead.                                                                                                                      
WARNING -  Template variable warning: 'google_analytics' is being deprecated and will not be available in a future version. Use 'config.google_analytics' instead.                                                                                                  
WARNING -  Template variable warning: 'repo_name' is being deprecated and will not be available in a future version. Use 'config.repo_name' instead.                                                                                                                
WARNING -  Template variable warning: 'page_description' is being deprecated and will not be available in a future version. Use 'config.site_description' instead.                                                                                                  
WARNING -  Template variable warning: 'site_author' is being deprecated and will not be available in a future version. Use 'config.site_author' instead.                                                                                                            
WARNING -  Template variable warning: 'current_page' is being deprecated and will not be available in a future version. Use 'page' instead.                                                                                                                         
[...]

PR #65, opened 2016-12-03
Update all templating references to avoid deprecated MkDocs values

brendo:
As described in #63, this PR replaces all the deprecated templating variables with the new recommendations. It also uses all the [new block names](http://www.mkdocs.org/user-guide/styling-your-docs/#overriding-template-blocks) for easy overriding. This should allow users of this theme to make little tweaks easily as required.

I also tweaked the README slightly to include MkDocs 0.16 as the requirement. While this will probably change before 1.0, I figured it'd help anyone hacking away on the `rework` branch.

Just for good measure, favicon and google analytics is supported again.


squidfunk:
Will check this when I'm back home in a few days

squidfunk:
Merged. Needed a few adjustments for the table of contents. Thanks for contributing!

PR #66, opened 2016-12-03
Specimen updates

brendo:
- Updates specimen to include a few additional checks and balances.
- Creates the inlinehilite content page.
- Updates the SASS to correct [inline code styling](https://www.impressivewebs.com/fixing-styles-on-code-tags-nested-inside-links/) as it appears in links.
- Updates the SASS to support alignment on tables

I have _not_ built the assets for this PR, largely because I didn't want any conflicts with #65.


squidfunk:
Will check this when I'm back home in a few days

PR #77, opened 2017-01-03
Remove hard-break in the copyright, do it with CSS

brendo:
A small change that removes the hardcoded `<br />` from the copyright template. The stacking looked awkward to me, which is was more pronounced when the `mkdocs` config omitted the copyright information.


squidfunk:
Replaced the `span` with a `div` - the block display has the same result but is the easiest CSS solution in my opinion. Now you can style it with CSS if you would like to have both on the same line.

PR #97, opened 2017-01-14
User overflow:auto instead of overflow:scroll

### LINKED ISSUE #96: Firefox Scroll bars
ffox
## Description

Ugly un-needed scrollbars are displayed in forefox

### Expected behavior

Scroll bars only appear when needed

### Actual behavior

Ugly scroll bar appears all the time

1. Open firefox

### Package versions

- Python: Python 2.7.11
- MkDocs: mkdocs, version 0.16.1
- Material: Version: 1.0.0

### System information

- OS: Exherbo Linux
- Browser: Firefox 50.1.0

roidelapluie:
Firefox does not support scrollbars customization like webkit does
because it is not part of the CSS standard. So it is better to hide the
scrollbars when they are not needed.

Fix #96

Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>

### DIFF
--- a/src/assets/stylesheets/layout/_nav.scss
+++ b/src/assets/stylesheets/layout/_nav.scss
@@ -260,7 +260,7 @@
       // List of items
       .md-nav__list {
         flex: 1;
-        overflow-y: scroll;
+        overflow-y: auto;
       }
 
       // List item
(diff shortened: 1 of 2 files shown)


squidfunk:
Fixed in https://github.com/squidfunk/mkdocs-material/pull/105

roidelapluie:
Sorry about the build..

Issue #159, opened 2017-02-16
Footer bug when scrolling in Chrome in large screen


ovasquez:
## Description
The footer appears and disappears when the contents of the page are small and don't require the page to expand. I created a GIF with the description of the bug http://imgur.com/a/P1fT9

You can also see this bug in the License page of the Material docs http://squidfunk.github.io/mkdocs-material/license/

**Note**:  I already found a fix for this (because just setting a min-height on .md-container or .md-main would result in the wrong height in Firefox), so let me know if it's OK to send a pull request.
### Expected behavior

The footer should stay still.

### Actual behavior

The footer "rebounds" and doesn't allow to scroll to the end of the page.

### Steps to reproduce the bug

1. Open Chrome and go to License page http://squidfunk.github.io/mkdocs-material/license/
2. Maximize the window
3. Try to scoll bottom

### Package versions

Latest version on 2017/02/15

### System information
* OS: Windows 8/10
* Browser: Chrome
* 1080p monitor

squidfunk:
You are encouraged to submit a PR. (However, I'm currently having some problems with flaky screenshot tests which I'm trying to fix, so the build may break.)

squidfunk:
Fixed in master with a more general solution (see comment in linked PR). Will do a release shortly.

PR #163, opened 2017-02-16
Fixed the scroll error in Chrome with small inner content

ovasquez:
This fixes the issue reported in #159 where the _min-height_ of the _md-main__inner_ calculated to 100% would result in undesired behavior in Chrome. This was the simplest solution I found because changing other containers resulted into Firefox _min-height_ bugs.

This was tested in Chrome 56, Firefox 51 and IE 11.

### DIFF
--- a/src/assets/stylesheets/layout/_base.scss
+++ b/src/assets/stylesheets/layout/_base.scss
@@ -109,7 +109,7 @@ hr {
 
     // If the browser supports calc(), no JavaScript is necessary
     .csscalc & {
-      min-height: calc(100% - #{5.6rem - 3rem});
+      min-height: calc(95% - #{5.6rem - 3rem});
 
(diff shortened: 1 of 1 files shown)


squidfunk:
First, thanks for sharing your solution. However, a few concerns:

1. When submitting a PR, please adhere to the standard described in the CONTRIBUTING.md. It's important to perform the build and check in the built files. Otherwise the PR cannot be accepted, as the source and build wouldn't be the same, thus somebody checking out the repository would not have your fixes present.
2. I don't consider this a good fix, as it's more of a fiddling solution. I will look into this and try to find a more generic solution. Your solution may and most probably is causing issues with some browsers or screen sizes (or content configurations).

ovasquez:
That's OK. I will close this one then, I hope you find a more general solution for this.

PR #206, opened 2017-03-11
Docs - steps for adding CSS within theme extension

### LINKED ISSUE #204: How to include extra CSS with theme extension
I'm extending Material for use on a site that will be hosting several dozen books. I have a directory within my custom publishing system that I'm using to store "theme extension" files (like partials) which override some of Material's settings when I run `mkdocs` with the `--theme-dir` option. This is great.

Now I want to add some custom CSS to my theme extension, but I'm having trouble. At first, I thought that if I put a CSS file like `custom.css` in my `theme-dir` directory, this CSS would be incorporated into the build. Wrong. It seems that the files in `theme_dir` only get built when they are overriding existing theme files, not when they are adding something new. Bummer.

I've gotten the custom CSS to work by automatically adding the `extra_css` setting within each book's `mkdocs.yml` file during publishing. But this approach is quite cumbersome for the following reason: the `extra_css` setting doesn't copy the css files and incorporate them into the build, it just adds a `<link>` element to the built html which loads the specified path. This is intuitive if the CSS is stored within the book, but not intuitive if the CSS is stored somewhere else (outside the book), and not even possible if that "somewhere else" is outside of the web-accessible directory of my app. I'm able to get this to work by storing my extra CSS separately from my theme extension, but this just feels like a hack. I wonder if there's a better way. 

Is there a way I can get new CSS within a "theme extension" of Material to show up in the built site?

seanmadsen:
This will close #204

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -159,6 +159,18 @@ The Material theme provides the following template blocks:
 For more on this topic refer to the [MkDocs documentation][4]
 
   [4]: http://www.mkdocs.org/user-guide/styling-your-docs/#overriding-template-blocks
+  
+### Adding custom CSS
+
+If you want to add additional CSS to your theme extension, you'll need to make sure MkDocs builds your HTML pages with a `<link>` tag to load your new CSS. To do this, you can override the `extrahead` block as follows:
+
+1. Add `extra.css` in `<theme-dir>/assets/stylesheets/` with custom style of your choosing.
+1. Add `main.html` in `<theme-dir>` with the following content. 
+    ```html
+    {% block extrahead %}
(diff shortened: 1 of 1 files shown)


squidfunk:
This is a very specific use case. The section on overriding template blocks links to the MkDocs documentation at the end for further information, in my opinion that should be enough:
http://squidfunk.github.io/mkdocs-material/customization/#overriding-template-blocks

Since you seem to be the only one with the problem for now and figured it out, I would say we leave it out of the documentation for now. It may confuse people that want to add `extra_css` the normal way. However, we may add it, if this problems pops up again.

Thanks anyway for your PR!

seanmadsen:
That makes sense. No worries. Feel free to close. :slightly_smiling_face:

PR #215, opened 2017-03-14
Do not using mkdocs force as entrypoint

netroby:
Signed-off-by: huzhifeng <huzhifeng@douyu.tv>

Do not using mkdocs as entrypoint, by default the default entrypoint will be ```sh```
that we can using the docker image to build or just get a shell into it . 

Please free entrypoint to us .do not force us using mkdocs as entrypoint

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,5 +48,4 @@ WORKDIR /docs
 EXPOSE 8000
 
 # Start development server by default
-ENTRYPOINT ["mkdocs"]
-CMD ["serve", "--dev-addr=0.0.0.0:8000"]
+CMD ["mkdocs", "serve", "--dev-addr=0.0.0.0:8000"]


squidfunk:
Why would you want to do that? In my opinion this makes no sense at all plus you can override the entrypoint when starting the image if necessary.

netroby:
I  am using docker image to intergrate with gitlab CI , to auto build document from source code . I have failed to using your original Dockerfile, and i have to modify your Dockerfile.

https://github.com/netroby/mkdocs-material/blob/master/Dockerfile

Your Docker image can not using to build document

netroby:
@squidfunk  using docker run  -ti squidfunk/mkdocs-material sh will just got error 

```bash
 docker run -ti squidfunk/mkdocs-material sh 
Usage: mkdocs [OPTIONS] COMMAND [ARGS]...

Error: No such command "sh".
```

squidfunk:
Did you read https://hub.docker.com/r/squidfunk/mkdocs-material/ and try:

``` sh
docker run --rm -it -v `pwd`:/docs squidfunk/mkdocs-material build
```

netroby:
It does not work as that way. the gitlab-ci won't work as that way.
Here is my new dockerfile https://github.com/netroby/docker-mkdocs-material
```dockerfile
FROM jfloff/alpine-python:2.7
LABEL author="ZhiFeng Hu <hufeng1987@gmail.com>"


RUN apk update; \
    apk add python2-dev ; \
    pip install mkdocs pygments pymdown-extensions  mkdocs-material 

```
and .gitlab-ci.yml
```yaml
image: netroby/docker-mkdocs-material

pages:
  script:
  - mkdocs build
  - mv site public
  artifacts:
    paths:
    - public
  only:
  - master

```

Anyway , thanks for you help.

squidfunk:
The thing is that this Docker image is intended to be self-contained so you can write your docs using the included server and build them. This is perfectly possible. If GitLab CI demands something else, you might need to build it yourself or use another image that contains mkdocs-material. There seems to be quite a few. Maybe there's an option to override the entrypoint in your configuration, as this is possible with `docker --entrypoint /bin/sh`.

PR #228, opened 2017-03-19
Set favicon type when using png

seanmadsen:
When I upgraded from 1.3.0 to 1.4.1 the favicon stopped working when building sites. I think this is why.


squidfunk:
I will close it for now, as you say. Feel free to reopen or create a new one if you find out what's causing the issue.

PR #245, opened 2017-04-03
Multilanguage search support

automation-stack:
## Extending search

Due to the fact that `Lunr` Javascript library currently only supports English, we have not support other languages out-of-box. But you can extend default search with `lunr-languages` package and add support for other languages.

The following languages are available: `German`, `French`, `Spanish`, `Italian`, `Japanese`, `Dutch`,  `Danish`, `Portuguese`, `Finnish`, `Romanian`, `Hungarian`, `Russian`, `Norwegian`.

- Add as submodule into your project `lunr-languages` repository.
```bash
> git submodule add git@github.com:MihaiValentin/lunr-languages.git
```

- Add languages and multilanguage support as `extra_javascript`
For convinience just create symlink into `docs` directory to `lunr-languages` directory
```yaml
extra_javascript:
  - 'lunr-languages/min/lunr.stemmer.support.min.js'
  - 'lunr-languages/min/lunr.de.min.js'
  ...
  - 'lunr-languages/min/lunr.ru.min.js'
  - 'lunr-languages/min/lunr.multi.min.js'
  - 'js/lunr.handler.js'
```

- Now you can handle and extend search with default index (`lunr.handler.js`)
```js
lunr.events.addListener(
  'index.created',
  (index) => {
    index.use(lunr.multiLanguage('en', 'de', 'ru'));
  }
);
```


src/assets/javascripts/components/Material/Search/Result.jsx
```diff
@@ -23,6 +23,9 @@
 import escape from "escape-string-regexp"
 import lunr from "lunr"
 
+window.lunr = lunr
```
squidfunk: This breaks with all the nice encapsulation and exposes `lunr` to the window. Pleae find another way to do this that doesn't involve putting objects into the global object. There should at any time only be a single entrypoint and that is `app` in `application.js`

docs/customization.md
```diff
@@ -165,6 +165,43 @@ For more on this topic refer to the [MkDocs documentation][4]
 
   [4]: http://www.mkdocs.org/user-guide/styling-your-docs/#overriding-template-blocks
 
+## Extending search
```
squidfunk: There are some grammatical mistakes in this documentation, e.g. articles missing etc.

docs/customization.md
```diff
@@ -165,6 +165,43 @@ For more on this topic refer to the [MkDocs documentation][4]
...
+  - 'lunr-languages/min/lunr.ru.min.js'
+  - 'lunr-languages/min/lunr.multi.min.js'
+  - 'js/lunr.handler.js'
+```
+
+- Now you can handle and extend search with default index (`lunr.handler.js`)
+```js
+lunr.events.addListener(
+  'index.created',
+  (index) => {
```
squidfunk: Demo JavaScript should be written in ES5 so it runs everywhere without compilation.

squidfunk:
Thanks for your PR. I understand the problem, however the way it is implemented is not ideal. If you could improve your implementation, I'm happy to merge your PR.

squidfunk:
Closing due to inactivity.

I will look into this when migrating to lunr 2, maybe there's an easy way of integration. I would also want this to be integrated with Material directly, so that no configuration regarding stemming etc. is required. it could be coupled with #274.

squidfunk:
With commit https://github.com/squidfunk/mkdocs-material/commit/6336e5055c2c3f6c0bb29b1b170633ce3cdae091, solving this became even easier

1. The language is set via localization, default is English
2. The search component reads the document's language attribute, loads the corresponding libraries/stemmers automatically from a CDN and sets the language in lunr.js

Will implement this when I find the time.

PR #275, opened 2017-04-14
override def. Home w site_name, show site_name in header on 404 page

mjames-upc:
Without specifying a page name in mkdocs.yml mkdocs will default to "Home", which is useful to have in the side-menu but prefer config.site_name be displayed by default in the header for index.md pages. 

This change also allows config.site_name to display on the 404 page.

### DIFF
--- a/material/partials/header.html
+++ b/material/partials/header.html
@@ -17,14 +17,20 @@
       <div class="md-flex__cell md-flex__cell--stretch">
         <span class="md-flex__ellipsis md-header-nav__title">
           {% block site_name %}
-            {% if page %}
-              {% for parent in page.ancestors %}
-                <span class="md-header-nav__parent">
-                  {{ parent.title }}
-                </span>
-              {% endfor %}
-            {% endif %}
-            {{ page.title | default(config.site_name, true) }}
+              {% if page %}
+                  {% for parent in page.ancestors %}
+                      <span class="md-header-nav__parent">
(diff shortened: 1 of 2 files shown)


squidfunk:
Please read the contributing guide before proposing a PR - your changes are not persistent and would be overwritten on the next build.

squidfunk:
However, this is rather a hack then a solution.

mjames-upc:
> Please read the contributing guide before proposing a PR - your changes are not persistent and would be overwritten on the next build.

You say this as if the contributing guide explains how changes would be or not be persisted.  I'm left to guess that the changes would have to be applied to src/ rather than material/ (or both) and then run through your build process, which I'm not using.  So I'll leave it for you to consider these changes outside of this PR (which I'll close), even if they aren't up to your standards (ahem), they do address these small design annoyances.

squidfunk:
Well, I didn't want to sound harsh, if that's what you got. I was just referring to the fact that I cannot accept this PR and wanted to discuss it, as I rather view it as a hack, but I will look into it.

In my opinion the [contributing guide](http://squidfunk.github.io/mkdocs-material/contributing/) explains the necessary steps pretty well, as it links to the guide for [setting up the development environment](http://squidfunk.github.io/mkdocs-material/customization/#theme-development) which in turn contains an admonition warning block that addresses this problem.

Do you have some suggestions on further guidance?

PR #281, opened 2017-04-17
Add site_name to example mkdocs.yml

mattbostock:
`site_name` is a required field, so include it in the example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ pip install mkdocs-material
 Add the following line to your `mkdocs.yml`:
 
 ``` yaml
+site_name: Test site
 theme: 'material'
 ```
 


squidfunk:
The Material documentation states that you already have a working MkDocs project. It is not intended to fix you up with MkDocs. The docs clearly state:

> Add the following line to your `mkdocs.yml`

`site_name` might end up being redundant in `mkdocs.yml`. We could add a notice in the Getting started guide, but no one ever complained about this, so I would favor to keep it short.

mattbostock:
I misread that line to mean 'create an `mkdocs.yml` file with the following content', I guess because the pip package for Material [pulls in mkdocs itself](https://github.com/squidfunk/mkdocs-material/blob/8ac6712dc3031ac2e7b2b82fd96f7c0d8df282d7/requirements.txt#L21) and I was following the README as a new user of Mkdocs (finding a good theme played an important part in my decision to use Mkdocs).

So feasibly you might follow the README as a new Mkdocs user and be confused why it complains about `site_name` (as I did).

That said I can understand that you don't wish to duplicate the documentation for Mkdocs.

mattbostock:
Closed in favour of a smaller change:
https://github.com/squidfunk/mkdocs-material/pull/289

PR #289, opened 2017-04-21
Make it clearer mkdocs.yml should exist

mattbostock:
Make it clearer that `mkdocs.yml` should exist already and the `theme` line should be added to it.

Otherwise you might follow the README as a new Mkdocs user and be confused why it complains about a missing mandatory field such as `site_name`.


squidfunk:
Thanks for your contribution!

PR #290, opened 2017-04-22
Hero teaser feature

squidfunk:
**Work in progress** - already prototyped for pre-1.0.0, it will finally make it into Material.

bildschirmfoto 2017-04-22 um 12 43 07
bildschirmfoto 2017-04-22 um 12 43 18
bildschirmfoto 2017-04-22 um 12 43 23


carlosperate:
Will this be a configurable option? It'd be really nice if there is background or product image in the header, but when used only for the page title like in the first screenshot I'm not sure how practical it is to have so much vertical spaced used.

squidfunk:
It will be a feature, silently enabled at first but I intend to switch it to the default look. However, it will be configurable and fully downward compatible.

carlosperate:
That's awesome, thanks @squidfunk!

material/base.html
```diff
@@ -86,8 +86,19 @@
       {% include "partials/header.html" %}
     {% endblock %}
     <div class="md-container">
-      {% set feature = config.extra.get("feature", {}) %}
-      {% if feature.tabs %}
+      <div class="md-banner"> 
+        <div class="md-banner__inner md-grid">
```
brendo: What do you think about abstracting this to a partial? I'd really like to customise this section of the site and with the banner feature I can see a lot of additional opportunities!

squidfunk: Sure, up to now it's just a quick and dirty implementation which needs some cleanup but it will definitely be encapsulated in a partial.

coliff:
this looks really great. would love to be able to use this on our Mkdocs-material-powered site. Is there any news on when we can expect to see this finished/released?

squidfunk:
@coliff not really. I'm about to integrate some tests before I incorporate this into the theme, so this will have to wait for some time.

squidfunk:
Closing this for now, as there's no timeline on integrating this. Some higher priority stuff to be done before, like visual tests etc. The branch is already stale, but I might flesh it out later.

squidfunk:
Implemented in #604. Will be merged in to master - would love to hear some feedback before the feature makes it into the release.

squidfunk:
It is already deployed here:
https://squidfunk.github.io/mkdocs-material/extensions/metadata/

coliff:
Looks great! I noticed when I scroll to the top the 'Yes, this is set via Metadata' indents a little bit - looks like a bug. Not investigated fully though... I'm using Chrome 62 on Windows 10.

squidfunk:
You mean that it fades in and moves a little from the bottom? that's actually intended, same behavior as the tabs.

coliff:
The 'Material for MkDocs' / 'Metadata' at the very top is nicely animated and moves when scrolling to the top. The 'Yes, this is set via Metadata' isn't animated though so when it moves suddenly its slightly jarring. would be nice if it was animated too.

carlosperate:
In firefox, during the scroll-up resize, the top area momentarily shows scrollbars:

hero

martinbira:
Can confirm that behaviour in Google Chrome 62.0.3202.94

squidfunk:
Oh, that's interesting and unintended. Tanks for reporting. Can you open an issue?

carlosperate:
Issue https://github.com/squidfunk/mkdocs-material/issues/606 opened.

squidfunk:
Fixed in #608

PR #302, opened 2017-04-24
Use type="search" - more semantically correct

coliff:
type="search"- more semantically correct and has accessibility benefits. Very old browsers which don't recognise type=search will fallback to type=text so there are no compatibility issues. I've tested this locally and works the same as before.

### DIFF
--- a/material/partials/search.html
+++ b/material/partials/search.html
@@ -3,7 +3,7 @@
   <label class="md-search__overlay" for="search"></label>
   <div class="md-search__inner">
     <form class="md-search__form" name="search">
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for your PR. However, you made your changes in the `material` folder which is compiled from `src`. See
http://squidfunk.github.io/mkdocs-material/contributing/#submitting-a-pull-request-pr and
http://squidfunk.github.io/mkdocs-material/customization/#theme-development

Did you also test in all major browsers? The user agent styles may make it look different and break the layout.

coliff:
Thanks for the quick reply and sorry for opening the PR on the wrong file. I've noticed an issue in Chrome actually (an extra close 'x' button appears in search results), so this may need a CSS change to hide. I'll close this PR for now and investigate some more.

squidfunk:
Yeah, browsers are pretty inconsistent in styling search fields. Especially the Microsoft products are pretty ugly and hard to get right, that's I was asking.

Issue #22, opened 2016-03-20
Favicon option is ignored


mnapp:
This is true for 0.2.1 and the current master version as well (in master only the variable name is changed.)

In base.html we have this portion

```html
{% set favicon = favicon | default("assets/images/favicon-e565ddfa3b.ico") %}
    <link rel="shortcut icon" type="image/x-icon" href="{{ base_url }}/{{ favicon }}">
    <link rel="icon" type="image/x-icon" href="{{ base_url }}/{{ favicon }}">
```

I use mkdocs 0.15.3 and material 0.2.1 and I can't get the favicon to stick for the life of me. If I remove that "set favicon" portion and write something like:

```html
{% if site_favicon %}
      <link rel="shortcut icon" type="image/x-icon" href="{{ base_url }}/{{ site_favicon }}">
      <link rel="icon" type="image/x-icon" href="{{ base_url }}/{{ site_favicon }}">
    {% else %}
    {% set icon = icon | default("assets/images/favicon-e565ddfa3b.ico") %}
    {% endif %}
```

It works. I am very much a noob and don't understand how that Jinja2 filter is supposed to work but it doesn't work how it's described in the documentation (e.g. just set site_favicon and be done).

squidfunk:
This is related to #16 and fixed in 0.2.2. I wanted to wait with the release until some more bugs appear, but I think best is if I release it now.

squidfunk:
Just released 0.2.2. Please update and try setting the favicon via `site_favicon` again.

mnapp:
Merged this into my custom modded version and it works now. Thanks.

PR #319, opened 2017-05-11
Added missing documentation for the site_favicon option

ovasquez:
I noticed this setting is the way to set the favicon in mkdocs-material (different from mkdocs's way), I found this setting in #22 and I wanted to add documentation for anyone else looking to change the favicon.


squidfunk:
Thanks for the PR!

Issue #315, opened 2017-05-08
Integrate clipboard.js for docs


coliff:
Feature request: Integrate clipboard.js for docs - maybe as an optional plugin?
https://clipboardjs.com/

squidfunk:
Are you speaking of integration for the code sections or what in particular?

coliff:
Yes, for code sections (sorry I didn't make that clearer)

facelessuser:
This seems like something that could be done with some custom JavaScript without requiring Material to include this.  I feel like you could just have some JS that targets the code blocks, adds a unique id, and then injects a clipboard button linked to the that unique id.

Having it built-in is fine though if that is the direction that is wanted, but I feel like you could get that functionality now without Material having to intervene.  That's just the impression I get from reading `cliboard.js`'s documentation.  I haven't tried it yet.

facelessuser:
This is just a real quick proof of concept.  This is hunting for code highlighted under `<div class="codehilite"><pre>`.

Some CSS:

```css
.md-typeset div.codehilite {
  position: relative;
  overflow: visible;
  padding: 0;
}
.md-typeset div.codehilite pre{
  overflow-x: auto;
  display: block;
  padding: 1rem 1.2rem .8rem;
}
.md-typeset .btn {
  float: right;
  display: inline-block;
  position: absolute;
  top: 3px;
  right: 3px;
  opacity: 0;
  transition: opacity 0.3s ease-in-out;
  background-color: #eee;
  background-image: linear-gradient(#fcfcfc,#eee);
  border: 1px solid #d5d5d5;
  border-radius: 3px;
  padding: 6px 12px;
  cursor: pointer;
}
.md-typeset .btn:hover {
  opacity: 1;
  transition: opacity 0.3s ease-in-out;
}
```

Some JavaScript to convert these code blocks into clipboard ready blocks. `doc_theme/extra.js`.

```js
(function (document) {
    function clipboard_init() {
        var charts = document.querySelectorAll("div.codehilite"),
            arr = [],
            i, j, maxItem, pre, btn, el;

        // Make sure we are dealing with an array
        for(i = 0, maxItem = charts.length; i < maxItem; i++) arr.push(charts[i]);

        // Find the UML source element and get the text
        for (i = 0, maxItem = arr.length; i < maxItem; i++) {
            el = arr[i];
            pre = el.childNodes[0];

            pre.id = "hl_code" + i.toString();
            btn = document.createElement('button');
[...]

PR #330, opened 2017-05-16
Clipboard.js Support

facelessuser:
This adds Clipboard.js support.  I'm going ahead and creating a pull so we can discuss this in a review.

There are a few things that need to be addressed, particularly the initializing.  Currently I rely on the user including clipboard.js and check if the `Clipboard` is available before initializing the clipboard overlays. Due to the approach it requires me to to add `$FlowFixMe` and `eslint-disable-line no-undef`.  I realize these exceptions are most likely not desired, but I want to get direction on the preferred way to handle this before I go further.

I imagine if we pass the `Clipboard` object in as a parameter, this would solve the lint issues, but that requires us to inject Clipboard.js before application.js.  Do we vendor Clipboard.js? Do we scan scan `extra_javascript` for clipboard.js and inject it ourselves and remove it from the array?

I also realize it may be desired to not have the Clipboard.js setup directly in application.js.  Once I know how we want to handle Clipboard.js, I can push it down into a submodule  if desired.

I have tested this in Chrome, Firefox, Mobile (Android), Edge, and IE.

References #315


squidfunk:
Thanks for the PR! I'm so sorry I haven't given feedback until now.

What about including clipboard.js as a default like lunr.js? I think it makes total sense and provides good value for only 3kb+ gzipped. I would add it to the build pipeline and it gets built with the whole bundle, so no need to integrate. I would also enable it by default without a possibility to disable, until someone complains about it, because I can think of no scenario where this wouldn't make sense.

facelessuser:
> sorry I haven't given feedback until now.

I just figured you were busy, which I completely understand.  I decided to issue the pull as I think I've tested most everything I can at this point.

> What about including clipboard.js as a default like lunr.js? [...] I would also enable it by default without a possibility to disable, until someone complains about it, because I can think of no scenario where this wouldn't make sense.

Sounds good to me.  I didn't want to presume, so I shied away from it until I could get an official response, but I generally agree.  I don't really see a downside of including it by default.

squidfunk:
Will provide a review tonight.

src/assets/javascripts/application.js
```diff
@@ -81,6 +81,47 @@ function initialize(config) { // eslint-disable-line func-style
         })
       })
     }
+
+    /* If Clipboard.js is available, inject "copy to clipboard" overlays
+       and attach Clipboard to copy the code.
+       Handle both div.codehilite>pre and pre>code. */
+    // $FlowFixMe
+    if (typeof Clipboard !== "undefined" &&
```
squidfunk: This should be encapsulated in a component, like the link blurring for example.

src/assets/javascripts/application.js
```diff
@@ -81,6 +81,47 @@ function initialize(config) { // eslint-disable-line func-style
...
+        Clipboard.isSupported()) { // eslint-disable-line no-undef
+      const blocks = document.querySelectorAll(
+        "div.codehilite>pre,pre.codehilite>code")
+      let i = 0
+      Array.prototype.forEach.call(blocks, code => {
+        const parent = code.parentNode
+        const codeId = `hl_code${i}`
+        const btn = document.createElement("button")
```
squidfunk: We should use JSX here, as it's more convenient

src/assets/stylesheets/application.scss
```diff
@@ -56,6 +56,8 @@
 @import "extensions/footnotes";
 @import "extensions/permalinks";
 
+@import "clipboardjs/clipboardjs";
```
squidfunk: Create a folder called `integrations`, so the clipboard integration is located in `integrations/clipboard`.

src/assets/stylesheets/clipboardjs/_clipboardjs.scss
```diff
@@ -0,0 +1,104 @@
...
+  // Tooltips for messages when a copy is successful or fails.
+  // Render it to the left as somtimes it is in a code table which
+  // cuts off the right, or on the edge of the screen with mobile.
+  .clip-tip {
+    &::after {
+      position: absolute;
+      top: 0.3rem;
+      left: 0;
+      padding: 0.6rem 1rem;
+      transform: translateX(-112%) translateY(0);
```
squidfunk: `112%` looks quite arbitrary, where does this value come from? What is `translateY(0)` for?

src/assets/stylesheets/extensions/_codehilite.scss
```diff
@@ -216,15 +216,23 @@ $codehilite-whitespace: transparent;
...
   .codehilite {
+    position: relative;
     margin: 1em 0;
-    padding: 1rem 1.2rem 0.8rem;
+    padding: 0;
     border-radius: 0.2rem;
     background-color: $md-code-background;
     color: $md-code-color;
     line-height: 1.4;
-    overflow: auto;
```
squidfunk: I need to validate, that everything works as expected here. `overflow` is a critical property for Material, as it induces margin collapse which may break other features.

src/assets/javascripts/application.js
facelessuser: I agree.

src/assets/javascripts/application.js
facelessuser: I'll have to look into JSX as I've never used it before :slightly_smiling_face:.

src/assets/stylesheets/application.scss
facelessuser: Cool, this is what I was looking for.  I wasn't sure what to call it, but I figured you wanted to shove it down somewhere.

src/assets/stylesheets/clipboardjs/_clipboardjs.scss
facelessuser: Yeah, it kind of is.  I was doing some research to do a tooltip, so I imagine I this can get cleaned up more.  I will get a better grasp on what going on and provide something that makes more sense.

src/assets/javascripts/application.js
squidfunk: No worries, I will handle this.

src/assets/stylesheets/clipboardjs/_clipboardjs.scss
squidfunk: I will also have a look.

squidfunk:
Okay, so most of the changes involve architectural stuff, so it's fine. I will go from here, makes most sense until you waste your time and I change it eventually anyway.

src/assets/stylesheets/extensions/_codehilite.scss
facelessuser: Understood, I believe this should check out okay.  In this case we are just flipping the the attributes with the wrapper and code element.  I'm actually using this over in my current pymdown-extensions docs.  But maybe their is a corner case I'm not aware of.

facelessuser:
> Okay, so most of the changes involve architectural stuff, so it's fine. I will take over from here, makes most sense until you waste your time and I change it eventually anyway.

Glad to hear. At least the bulk of the work is done.  I kind of figured you were going to tweak it anyways.  My frontend skills are serviceable but far from professional :slightly_smiling_face: .

squidfunk:
Frontend development is as worse as development can get. Every day a new framework, every month a new browser or update. Kind of fractal :D

PR #331, opened 2017-05-16
Fixed undefined source stats for private repositories

squidfunk:


notpushkin:
Sorry for bumping this, but I was wondering why do we have to iterate over user's repos here? Wouldn't just handling the 404 error at `api.github.com/repos/...` be enough?

squidfunk:
The change was made a long time ago but as I remember it's implemented that way as now also organizations can simply be linked. It's the same logic. However, I understand that it's suboptimal and have it on my list of fixing things as part of the rewrite I'm currently doing.

PR #333, opened 2017-05-16
Clipboard.js integration

squidfunk:
Thanks @facelessuser, I will continue polishing this from here.


facelessuser:
Look forward to seeing it finished :).

squidfunk:
Actually, when I remove all your changes from the `.codehilite` class, except `position: relative`, it keeps working. Why did you had to change the padding/margin of the nested tags? This breaks the scroll bar position and some other stuff, but seems to work without, at least in Chrome.

facelessuser:
Couple of reasons.

The overlay is positioned in the upper right, but when your text overflows, an you scroll to the right, the overlay doesn't stay in the upper right, but scrolls to the left.

This fixes an issue where the wrapper div gets a block inserted and causes a large gap at the top of code blocks, particularly those who highlight the first line.

```
.md-typeset .codehilite[id]:before {
  display: inline;
}
``

**Comment by facelessuser** [comment]
Like always, there may be a better way to implement the overlay.  I used Clipboard.js's page as an example, but noticed the way code blocks were done in Material were not compatible.  Unfortunately, what I didn't notice is the scroll bar position issue.

**Comment by squidfunk** [comment]
Hmm, but as stated, it changes/breaks the layout of the code blocks and I don't quite like it.

bildschirmfoto 2017-05-16 um 21 28 46

I'll try to find another way to do it.

**Comment by squidfunk** [comment]
Okay, so it seems possible to place the element right before the `codehilite` div (on the top level), but the hover effect (only show when mouse is inside the container) doesn't work them. However, I wouldn't include that effect anyway, because you won't see the button on an iPad. Thus, the button must always be shown. I will make it a little less obtrusive and think it should work out.

**Comment by facelessuser** [comment]
Does iPad work different than Android?  Touching a code block in Android shows the faint button, and then pressing it copies.

**Comment by facelessuser** [comment]
At least in Chrome on Android.

**Comment by squidfunk** [comment]
No, it will show the button, but it's considered really bad practice from a UX point of view to hide actions behind hover/click actions on touch devices. Material does this nowhere on purpose.

**Comment by facelessuser** [comment]
> No, it will show the button, but it's considered really bad practice from a UX point of view to hide actions behind hover/click actions on touch devices. Material does this nowhere on purpose.

Ahh, yeah, I'm no UX guy :slightly_smiling_face:.

What browser was that image from?  I don't see that.  It looks like margin is applied to the wrapper and and the code block.

**Comment by squidfunk** [comment]
Just Chrome. This is the padding you set on the outer element which renders a scrollbar for overflow.

**Comment by facelessuser** [comment]
Do you see that here? http://facelessuser.github.io/pymdown-extensions/extensions/superfences/.  I don't.  Maybe I got something wrong?  I don't think I saw it on Mac either, but my Mac is at my house.  I work on Windows.

**Comment by facelessuser** [comment]
I have a couple of overflows on that page.

**Comment by squidfunk** [comment]
Actually, I don't. Very strange.

**Comment by facelessuser** [comment]
I must have messed up somewhere.  I'm using a mock up there of what I was going for.  I can hunt it down this evening.

**Comment by facelessuser** [comment]
If it helps, my doc CSS is here: https://github.com/facelessuser/pymdown-extensions/blob/master/doc_theme/extra.css.  It should mirror what's in my pull (minus custom stuff).

**Comment by squidfunk** [comment]
No worries, I'll figure it out. It will probably work with some tweaks.

**Comment by facelessuser** [comment]
Feel free to ask about anything you need.  Happy to help.

**Comment by squidfunk** [comment]
Just discovered that your changes will only work if the codehilite extension is enabled, so if somebody wants to use another, e.g. JS-based syntax highlighter, the button is not shown. The markup is pretty different when codehilite is not enabled. This is everything but easy, may take some time.

**Comment by facelessuser** [comment]
Can't they just set `use_pygments` to false in CodeHilite to use a JavaScript highlighter?

That's why I have code for `div.codehilighte pre` and `pre.codehilite code` .  If you don't do CodeHilite, they you also have to provide style for `pre code`, but it is likely that the JS highlighter will obliterate whatever is there and possibly write in something like `div.someotherclass pre` etc.

**Comment by squidfunk** [comment]
I need to think about this a little. Probably putting it before the code block will be easier to implement, less error prone and more compatible.

**Comment by facelessuser** [comment]
Understood, there appears to be more flexibility considerations than my narrow use case.

**Comment by facelessuser** [comment]
Thinking about this more, just having it always muted and transparent except on hover is probably quite reasonable.

Since rewriting those block won't destroy the button with the button before the block, I think in pymdown-extensions, I would need to switch UML output to a div to keep them from getting targeted, but that support has been in for a bit now. I just need to start using it.

**Comment by facelessuser** [comment]
So I dug into this a little more tonight.  Here is a couple things I found.  It seems most modern JS highlighters can target HTML5 style which Codehilite (with `use_pygments=False`) outputs:

 ```
<pre><code class='language-mylang'>...</code></pre>
```

 I tried `Prism.js` and `Rainbow.js` . `Highlight.js` doesn't look for HTML5 style, so I let it guess syntax, but a custom load is easy to code up for it.  All left `pre` intact and allowed clipboard to work.  So I don't think there is any risk of JS highlighters not being able to use the current format.  There are probably a couple old school highlighters that won't play nice as they destroy the pre and code blocks; they would also destroy the id assigned to the pre, so clipboard wouldn't be able to find them anyways if you inserted the button before them.

The reason I bring this up is that I also tried to overlay the button by injecting it before the code blocks.  I found that to be a pain to overlay as positioning was affecting as code blocks are nested under other objects.  Also, targeting non JS highlighting (Pygments) with line numbers is troublesome as you have to find the code, but then step all the way out of the table.  To be honest, this seems like more work, and it won't protect you against destructive JS highlighters that obliterate the pre and code with your nicely crafted ID for clipboard.

I honestly think by just changing the `pre code` CSS to be similar to `div.codehilite pre` and maybe relaxing Clipboard targeting to `pre>code` and `div.codehilite>pre`, so that even if CodeHilite is not used, clipboard will still work, would be much easier.  And obviously dropping reveal on code hover for compliance for good UX on touch devices.  It just seems like more pain than it's worth to go the other route, and I tried.

Anyways, just thought I'd share the findings.  Maybe it will save you some pain.  Or maybe you have some good tricks that makes going the other way easy.  Anyways, good luck!

facelessuser:
I was able to fix scroll offset issue and get clipboard working in CodeHilite (with and without Pygments) and without CodeHilite.  Also changed how the button operates.  Always faintly visible and becomes fully visible when hover on button: https://github.com/facelessuser/mkdocs-material/commit/e062e7109d4d5b3d48d5a981e74c30c626b631a1.

I realize you may not be going in this direction anymore, but I wanted to go ahead and at least provide the experimental changes for reference.

squidfunk:
Awesome work, thanks! When implementing the codehilite stuff I wanted it to be as much as compatible with other syntax highlighters and this really was a huge pain in the ass. I think after inquiring into it you know what I'm talking about. I mean, it's just a container that should scroll, what could possibly go wrong? Much.

> The reason I bring this up is that I also tried to overlay the button by injecting it before the code blocks. I found that to be a pain to overlay as positioning was affecting as code blocks are nested under other objects. Also, targeting non JS highlighting (Pygments) with line numbers is troublesome as you have to find the code, but then step all the way out of the table. To be honest, this seems like more work, and it won't protect you against destructive JS highlighters that obliterate the pre and code with your nicely crafted ID for clipboard.

That's a really interesting insight. I hope to find some more time this week to get into this but thanks to your work I will be able to finish it faster.

facelessuser:
You asked earlier about the weird translate settings on the tooltip.  Yeah, I understand better now.  So yes, `translateY(0)` was unnecessary.  And second `translateX(-112%)` was also non-nonsense.  I was trying to position it left of the button with a little space between, but better would have been `translateX(-100%)` and then create a buffer by shifting left a constant amount with `left: 0.5rem`.  `-112%` was shifting it proportionally over to get it left with a gap, but the gap differs based on text length of the element since the transform is proportional.  All that is needed is `-100%` and then just offset the tip with `left`.

facelessuser:
One last small fix that I noticed when playing with `details` tags.  There was a slight overflow issue with code under `details` with padding.  We no longer need to specify `min-width: 100%` with the rework of code blocks. https://github.com/facelessuser/mkdocs-material/commit/f693a8ada3b18c0a0d1afd953ac87ae4070702a8

squidfunk:
When I add the `superfences` extension but omit `codehilite`, some blocks get wrapped inside a `div` with the class `highlight`, some don't. Those which do get wrapped have no `code` inside the `pre`, destroying the whole layout. This was one of the reasons why I chose to style only the  `pre` tags. This also deviates from [fenced_code](https://pythonhosted.org/Markdown/extensions/fenced_code_blocks.html).

Would you consider this a bug or is this intended behavior? I would say that `superfences` should try to imitate the behavior of `fenced_code` in regard to nesting. Could you add a `code` block below the `div > pre` structure? Then I think we have all cases covered.

squidfunk:
Also, is the `overflow: visible` on the outer container (`div` or `pre`) intentional? I think it is not necessary, or why did you set it?

facelessuser:
> Also, is the overflow: visible on the outer container (div or pre) intentional? I think it is not necessary, or why did you set it?

Probably  because my initial work was based on overriding default Material CSS.  As long as they haven't been set to something else previously, it is not needed.  I know when switching from `pre>code` to `div>pre` you have to override `pre` from it's earlier configuration, but that may be the only case.

facelessuser:
When you get it all ready, I can run your branch through known scenarios I've encountered to make sure there is parity with the output I originally configured it for.

facelessuser:
Cool, I figured there would be some unnecessary stuff, but I was hoping I got it close 🙂 .

squidfunk:
Jep, good job. BTW, I made the design a little less intrusive and tried to blend more with the code block style:

bildschirmfoto 2017-05-31 um 15 16 05

facelessuser:
Cool.  Yeah, I wasn't sure how close to "Material" you were aiming. I read over the Material spec about tooltips and buttons and such, and made a guess.  I like the triangle on the tips, I almost put it on, but then the Material spec was all "don't add triangles", so I am kind of glad you deviated there.  Can't wait to try out the final product.  Is the icon always full transparency? Or does it fade a little to not obscure text under it?

squidfunk:
No, it has 20% opacity by default, 100% opacity on code block hover and turns blue (accent color) on direct hover.

facelessuser:
Perfect!

squidfunk:
Okay, I think I got everything in this PR - can you check if I included all of your additional patches and test the scenarios you mentioned?

squidfunk:
FYI: GitHub seems to have problems. The Travis hooks is somehow broken. Hope they fix it asap.

squidfunk:
Ah, regarding the docs - I would remove the section, because Clipboard is now directly integrated with Material, so nothing to comment I think.

facelessuser:
> Ah, regarding the docs - I would remove the section, because Clipboard is now directly integrated with Material, so nothing to comment I think.

Makes sense.

> FYI: GitHub seems to have problems. The Travis hooks is somehow broken. Hope they fix it asap.

Yeah, GitHub had some outages this morning (my time).

> Okay, I think I got everything in this PR - can you check if I included all of your additional patches and test the scenarios you mentioned?

In general, it is looking good. I'll finish up those scenarios by this evening, I need to setup those scenarios, and I just don't have time before heading into the office this morning.

squidfunk:
No hurry, thanks for your support!

facelessuser:
Looks good to me.

facelessuser:
Everything checked out.

squidfunk:
Merged, just finishing the lunr search localization, so both can be released with 1.7.0

coliff:
And a very big thanks to the both of you from me for doing all that work in integrating my suggestion.

<3 Open Source

I just updated to latest version of 'mkdocs-material' and it works great.

PR #339, opened 2017-05-20
Updated the documentation for Discus Integration

sivapcu:
'site_url' field is mandatory in the 'mkdocs.yml' file for the Disqus Integration to work, else it will display the error message as "We were unable to load Disqus. If you are a moderator please see our troubleshooting guide."


squidfunk:
Thanks!

PR #349, opened 2017-06-01
Allow site logo to be any of the material icons

brendo:
Internally we've had some requests about changing the site logo to something other than the "hat". While I know it's possible to specify an image as the logo, it also dawned on me that we're already using the [Material Icons](https://material.io/icons/) font in this theme, so why not allow authors to select from any of the existing 900 icons?!

This PR adds a new config option, `logo_icon`, which if specified, will replace the default 'hat' logo with anything the author desires.


squidfunk:
Well, this is something I already thought about, but it collides with my plan to drop the Material icon font and use SVGs, as they can be embedded which doesn't cripple the layout with ligatures until the site is loaded. I'm still investigating switching to SVG but it is a very, very low priority issue.

For this reason, I think it makes no sense to add a feature that would be dropped anyway if I decide to switch to SVG. In my eyes, better support for browsers and faster loading outweighs some pains with the setup. Let's leave this PR open, but think about this properly before integrating it.

brendo:
That's fair enough, if the plan is to remove that font and replace with SVG's then I agree it's a bit awkward to draw further attention to it. I'm currently maintaining a thin customisation of the theme to add this functionality so I'm not attached to this PR. Feel free to close if required!

squidfunk:
Change of plan: the icon font will not be dropped, as it's not really convenient converting everything to SVG. Furthermore we loose the ability to set icons via CSS. For this reason this functionality will make it into master, but I have to check how exactly.

squidfunk:
Just released as part of 1.10.0:

``` yaml
extra:
  logo:
    icon: 'cloud'
```

The `icon` value can now be changed to any of the Material icons.

PR #350, opened 2017-06-01
Prepare 1.7.0 release

squidfunk:


facelessuser:
This release broke UML rendering for me.  Flowchart.js always throws an error now, and then rendered svgs for sequence diagrams are of size 20px.  I'm going to need to unravel all of this 😦 .

squidfunk:
Hmm okay, too bad. The only thing that is included in this release besides the Clipboard.js integration is multilingual search, but that should not impact anything inside `md-typeset`. Have you been using 1.6.4 before or some earlier release?

facelessuser:
Yes 1.6.4 renders things fine.  I'll reverse engineer what's going wrong.  I tried rendering the UML all the way up in body and then copying then back into place, and I'm just getting tiny avg.

As for why flowchart breaks, that's with too.  Are you using the clipboard from node_modules/clipboard/dist or from the root clipboard?  I wonder if that makes a difference...

facelessuser:
I guess when I experimented with webpack and clipboard, I pulled the browser release from dist.

squidfunk:
Hmm, but that shouldn't interfere with anything. Actually, the only exported variable is `lunr`, due to the way how the languages package works. I'm using Clipboard from the default export, which according to their `package.json` is `lib/clipboard.js`. It seems that the `lib` version has UMD-style exports, while the `dist` version exports to `window`. However, the UMD-version should be less intrusive, as it's handled completely inside Webpack.

squidfunk:
Can you post the error you get?

facelessuser:
> However, the UMD-version should be less intrusive, as it's handled completely inside Webpack.

Cool. I'm still learning about a lot of this stuff.  I've learned a lot just from doing a pull request on this repo.

> Can you post the error you get?

```
flowchart.min.js:6 Uncaught TypeError: Cannot read property 'key' of null
    at e (flowchart.min.js:6)
    at f (flowchart.min.js:6)
    at Object.drawSVG (flowchart.min.js:6)
    at exports.default (extra.js:208)
    at HTMLDocument.<anonymous> (extra.js:254)
```
Switching back to 1.6.4 clears it up for me.  I tired pulling out the extra stuff in in base.html in regards to luna, but that doesn't help.

squidfunk:
> Cool. I'm still learning about a lot of this stuff. I've learned a lot just from doing a pull request on this repo.

That's nice to hear.

> Switching back to 1.6.4 clears it up for me. I tried pulling out the extra stuff in in base.html in regards to luna, but that doesn't help.

Hmm, here are all changes between 1.6.4 and 1.7.0:
https://github.com/squidfunk/mkdocs-material/compare/1.6.4...1.7.0

squidfunk:
The only thing I can think of is the name of the code blocks. I changed them from `hl_code...` in your PR to `__code_...`

facelessuser:
Aha, I got it! I think it's a load order thing.  As you are now controlling clipboard, your were getting a hold of my uml blocks before my UML libraries.  Switching UML blocks to the div output format avoids getting them getting targeted for clipboard overlays.  I was the problem all along 😄 .

squidfunk:
Urgh. Should we do anything to mitigate this?

facelessuser:
Not sure.  It was probably my fault for defaulting UML to code format from the beginning.  I thought the code format was a nice fallback if the UML libraries didn't load.  People will probably have UML disappear on if they upgrade Material, but if people add this:

```
- pymdownx.superfences:
      custom_fences:
        - name: flow
          class: uml-flowchart
          format: !!python/name:pymdownx.superfences.fence_div_format
        - name: sequence
          class: uml-sequence-diagram
          format: !!python/name:pymdownx.superfences.fence_div_format
```

The problem will go away.

facelessuser:
Basically in recent versions of SuperFences I abstracted special blocks so people can add anything they want beyond UML.  Graphs, Charts, whatever.

squidfunk:
So the button element is the problem? Would inserting it after the codeblock help? It's positioned absolute anyway, so that would make no difference in execution.

facelessuser:
> So the button element is the problem?

In this case, it appears so.  I had thought I was executing clipboard before UML and it was working, but maybe not.  I haven't dug deep enough to be sure.  

> It's positioned absolute anyway, so that would make no difference in execution.

Maybe, but will that work with table code blocks?  Like with Pygments with line numbers?  Now I'm curious what happens when running a JavaScript syntax highlighter on this.  In my tests it worked, but maybe load order was different here too.

facelessuser:
Prism.js works.  Hmm..I wonder if this just affects UML.

facelessuser:
I don't think this is a problem for Material.  You don't mention directly supporting UML.  I think as people run into it, it should be a support issue for PyMdown Extensions.  It's not unexpected that there will be special cases that will require workarounds.

I wonder if I should default UML to div format moving forward and just field issues when people complaint about breakage.  My original UML js script was naive as there are cases (like in details) where the SVG can't figure out there height and width because their parent is hidden.  My new script injects them closer to the root to render them, and then inject them back where they need to be (I believe MathJax does the same).

facelessuser:
Yeah, it's not even that UML is incompatible, it's just my script naively assumes the first child will be the code block.  Even in my documents I mention that what I provide is just an example and it may require modifications.  I just need to give people a more polished script.

squidfunk:
Okay, so it's easily fixable :-)

I must say: I'm very pleased working with you! Pymdownx and Material have become really good companions fixing a lot what is wrong with Python Markdown and usable and compatible markup in general. You do the Python stuff, I fight with the frontend. I think this greatly shows the true value of Open Source.

facelessuser:
> Okay, so it's easily fixable :-)

Yes, in order to use this feature, users have to load their own snippet of JavaScript.  The example I provide assumes too much.  It needs to be more explicit.

> I must say: I'm very pleased working with you! Pymdownx and Material have become really good companions fixing a lot what is wrong with Python Markdown and usable and compatible markup in general. You do the Python stuff, I fight with the frontend. I think this greatly shows the true value of Open Source.

Thanks! I agree.

PR #358, opened 2017-06-04
Render repo in header only with URL

crackcomm:
It is much wider when repo box is not wanted.

### DIFF
--- a/material/partials/header.html
+++ b/material/partials/header.html
@@ -34,13 +34,13 @@
           {% include "partials/search.html" %}
         {% endblock %}
       </div>
+      {% if config.repo_url %}
       <div class="md-flex__cell md-flex__cell--shrink">
         <div class="md-header-nav__source">
-          {% if config.repo_url %}
-            {% include "partials/source.html" %}
-          {% endif %}
+          {% include "partials/source.html" %}
         </div>
       </div>
+      {% endif %}
     </div>
   </nav>
 </header>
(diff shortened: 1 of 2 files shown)


brendo:
Have noticed this as well, essentially pre this PR, there'd be a gap between the search box and the right side of the browser where the repo "should" exist.

I'm curious though, does this cause JS errors from trying to update an element that doesn't exist?

crackcomm:
There is no error.

This is how it was before:

crackcomm:
See [example](https://ipfs.io/ipfs/QmUzFWChxFgaHE5Hde25VNh6sCHsJWu9YUvVHkzcgfTmFS/).

squidfunk:
So if I understand your question correctly: why is the repo container rendered thus offseting the search to the levt, even when there is no repository linked to the docs? Well, this is done on purpose, so that the open search aligns perfectly with the content:

bildschirmfoto 2017-06-06 um 09 37 54

This will not be the case if your PR is merged. Personally, I think that this will impede a good user experience. For this reason, I won't include this PR, sorry.

squidfunk:
Hmmm.... testing this actually probably kind of works for me.... I need to think about it.

crackcomm:
In fact, I would like to have a feature that would disable the repository URL in the header but sources still can be linked. `repo_hide` set to `true` may hide it, what do you think?

squidfunk:
You can hide it with CSS or adjust the templates. I don't want to have too many hide/show features because there will be no end to this.

crackcomm:
@squidfunk hiding with CSS seems like a reasonable idea, thanks for the tip.

PR #363, opened 2017-06-07
Fixed support for Japanese search

miyajan:
When I set ```jp``` to ```search.languages```, the following JS error is output to a browser console.

```
Uncaught TypeError: e.TinySegmenter is not a constructor
```

The ```lunr.jp.js``` requires ```tinyseg.js```. (https://github.com/MihaiValentin/lunr-languages/pull/6)

So I fixed to load ```tinyseg.js``` when ```jp``` is in ```search.languages```.

### DIFF
--- a/material/base.html
+++ b/material/base.html
@@ -156,6 +156,9 @@
         <script src="{{ path }}/lunr.stemmer.support.js"></script>
         {% for language in languages | map("trim") %}
           {% if language != "en" %}
+            {% if language == "jp" %}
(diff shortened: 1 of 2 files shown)


squidfunk:
Cool, thanks!

squidfunk:
Will do a release for this immediately, because it's an annoying error.

miyajan:
@squidfunk Thanks! You are the best!

PR #401, opened 2017-07-21
Update EmojiOne demo url

bguyl:
The link http://emojione.com/demo does not exist anymore and it is replace by https://emoji.codes/


squidfunk:
Thanks!

Issue #400, opened 2017-07-20
Search bar takes tab focus


paulgibbs:
## Description

When I press tab, I expect to be able to navigation around the page with keyboard navigation. Instead, the search bar comes into focus when I press tab, regardless of how many times I try pressing tab again.

Is there a way to disable the JS capturing tab (and other keys) apart from not loading the entire JS file? This breaks keyboard accessibility for users on my site.

### Expected behavior

Use the tab key to navigate through interactive elements on a web page.

### Actual behavior

Search bar comes into focus.

### Steps to reproduce the bug

1. Go to http://squidfunk.github.io/mkdocs-material/
2. Press tab
3. Observe you are in the Search bar, as opposed to any other content on the page.
4. Press tab more.
5. You're still in the Search bar.

### Package versions

* Python: `2.7.10`
* MkDocs: `0.16.3`
* Material: `1.7.4`

### Project configuration

https://github.com/squidfunk/mkdocs-material/blob/master/mkdocs.yml :)

squidfunk:
Keyboard accessibility is a long standing thing I wanted to optimize, but I never had the time. If you come up with a good concept, I'm happy to integrate it. The problem is that the `blur` event cannot be used to close the search bar because that would not allow clicking or navigating through the results.

Secondly, when there are search results, <kbd>Tab</kbd> is mapped to <kbd>Key Down</kbd> while <kbd>Shift</kbd>+ <kbd>Tab</kbd> is mapped to <kbd>Key Up</kbd> and it wraps at the end of the search results.

How would you handle that and what is the order in which you would expect it to go through the page?

paulgibbs:
I'll talk to the accessibility expert on our team and get back to you with her suggestions.

rianrietveld:
Looking at:
```
<div class="md-flex__cell md-flex__cell--shrink">
        
	<label class="md-icon md-icon--search md-header-nav__button" for="search"></label>

	<div class="md-search" data-md-component="search">
		<label class="md-search__overlay" for="search"></label>
		<div class="md-search__inner">
			<form class="md-search__form" name="search">
				<input type="text" class="md-search__input" name="query" required="" placeholder="Search" accesskey="s" autocapitalize="off" autocorrect="off" autocomplete="off" spellcheck="false" data-md-component="query">
				<label class="md-icon md-search__icon" for="search"></label>
				<button type="reset" class="md-icon md-search__icon" data-md-component="reset">close</button>
			</form>
[...]

Issue #398, opened 2017-07-18
Search shortcut breaks forms


markeganfuller:
## Search shortcut breaks forms

We have a number of forms in our mkdocs site which are broken by the search shortcut (`s` / `f`).

### Expected behavior

Search only opens when not focused on a form element.

### Actual behavior

When typing in a form field (e.g. input / text area) an `s` or `f` will open the search.

### Steps to reproduce the bug

1. Add the following to a test page
```
<form>
<input></input>
<textarea></textarea>                                                                                                
</form>
```
2. Attempt to write `sausages` in the input / text area

### Package versions

* Python: `Python 2.7.13`
* MkDocs: `0.16.3`
* Material: `1.7.4`

### Potential fix

I think the code to fix this would be a change to: https://github.com/squidfunk/mkdocs-material/blob/master/src/assets/javascripts/application.js#L340

Something like this to check if the current focused element is part of a form:

```javascript
if (document.activeElement.form === undefined) {
  if (ev.keyCode === 70 || ev.keyCode === 83) {
    query.focus()
    ev.preventDefault()
  }
}
```

That seems to work in testing but doesn't pass the 'flow' check and I'm not quite familiar enough with js to quickly make this work.

squidfunk:
That's indeed a bug, thanks for reporting and the potential fix!

squidfunk:
Fixed in #407, will be released shortly with 1.7.5.

squidfunk:
Please provide a reproducible scenario (with configuration + Markdown, or even better, a repo), then we can fix it.

Kobzol:
Can I send you the website directly? Try this: https://docs.it4i.cz/modules-matrix/ focus the Module Filter input box and type "s". The repository for the website can be found here: https://code.it4i.cz/sccs/docs.it4i.cz.

squidfunk:
I've inspected your code - for your purposes: you can wrap the `input` in a `form` and the error should be gone. I may improve the detection of active form fields so it doesn't trigger for input fields that are not wrapped inside forms when I find the time at some point in the future.

Kobzol:
OK. That's an easy workaround, you're right. Thanks!

PR #402, opened 2017-07-21
Keyboard accessibility improvements for search form

paulgibbs:
Per https://github.com/squidfunk/mkdocs-material/issues/400#issuecomment-316968253 this change:

* Removes `accesskey`. This apparently is not recommend and accessibiltiy/browser implementation is too inconsistent.
* I've also removed the JS-powered "F" and "S" key binds that did basically the same thing. Same reason as above, but also because these were an internationalisation issue ("find" and "search" will probably not not begin with those letters in other languages).
* I've added the `dialog` role to the search model. There's more to do here (we need to add `aria-labelledby`), but that can be addressed in the future when the triple `label` elements are rationalised. I don't think adding the ARIA role hurts in the meanwhile.
* The tabbing JS was really good but I've made two improvements.
  * There was inconsistent modal behaviour if you hit tab and you had not made a search yet (i.e. no results).
  * Tab now cycles back to the input element, allowing keyboard users the ability to refine their search terms.

In testing, this makes the search tabbing feel much better to me.

I don't know modern JS very well, so I've tried my best to keep to your code standards, but do let me know if I should have done something different there so I can learn. Thanks!


paulgibbs:
This would also address #398 :)

squidfunk:
Thank you for this PR. It includes a lot of changes that need to be tested and validated cross-browser. As there are currently not end-to-end tests, this may take some time.

squidfunk:
I'm sorry but I will not completely merge this PR, because I personally find the <kbd>S</kbd> and <kbd>F</kbd> short cuts really practical. GitHub also has a <kbd>S</kbd> shortcut defined, so it's pretty common. I will incorporate the other changes you proposed in separate PRs (this PR is too big).

src/assets/javascripts/application.js
```diff
@@ -302,9 +302,10 @@ function initialize(config) { // eslint-disable-line func-style
 
         /* Retrieve all results */
         const links = Array.prototype.slice.call(
-          document.querySelectorAll("[data-md-component=search] [href]"))
-        if (!links.length)
-          return
+          document.querySelectorAll(
+            ".md-search__input, [data-md-component=search] [href]"
```
squidfunk: Also, why did you change this?

paulgibbs: The removal: makes tab behaviour consistent before you have made any search.
e.g. on master, load your site, hit tab several times. Then load your site again, make a search (so results show up), then hit tab several times. There is a behaviour difference and the removal was part of making the behaviour consistent.

The addition: lets you tab back to the search input box (otherwise how does a keyboard user go back to that box to refine their search?)

paulgibbs:
No worries about not merging the whole thing. Even if you can't cherry-pick specific commits, hopefully having the code will help you make improvements, anyway.

squidfunk:
Yes, thank you for investing your time, I'm about to incorporate your improvements with minor changes!

src/assets/javascripts/application.js
squidfunk: This is actually an awesome idea, feels much more natural.

squidfunk:
> Removes `accesskey`. This apparently is not recommend and accessibiltiy/browser implementation is too inconsistent.

Implemented in #406

> I've also removed the JS-powered "F" and "S" key binds that did basically the same thing. Same reason as above, but also because these were an internationalisation issue ("find" and "search" will probably not not begin with those letters in other languages).

Not implemented for reasons stated above

> I've added the `dialog` role to the search model. There's more to do here (we need to add `aria-labelledby`), but that can be addressed in the future when the triple `label` elements are rationalised. I don't think adding the ARIA role hurts in the meanwhile.

Implemented in #406 

> The tabbing JS was really good but I've made two improvements.
>  * There was inconsistent modal behaviour if you hit tab and you had not made a search yet (i.e. no results).
>  * Tab now cycles back to the input element, allowing keyboard users the ability to refine their search terms.

Implemented in #410

PR #450, opened 2017-09-06
Allows to restrict included toc items by their level

funkyfuture:
asserts that mkdoc's AnchorLink has a property named 'level', see https://github.com/mkdocs/mkdocs/pull/1272.

so, atm this is a proof-of-concept.


squidfunk:
Thanks for your PR. Three notes:

* You made changes in the `material` folder, which are overwritten by the build. See [this section](http://squidfunk.github.io/mkdocs-material/contributing/#submitting-a-pull-request-pr) on how to PR on this project.

* I don't know if this makes sense in general. I think it's better to use `<h*>` tags for specific headlines which you don't want to appear in the documentation, or CSS, e.g. `.md-nav--secondary .md-nav__list .md-nav__list .md-nav__list { display: none }`, to remove entire levels from the table of contents.

* You changed the dependency for `mkdocs` to 0.17, which does not exist, thus the build fails.

funkyfuture:
> You made changes in the material folder, which are overwritten by the build.

when i run `yarn install`, i get this error:

```
error css-mqpacker@6.0.1: The engine "node" is incompatible with this module. Expected version ">=5.0.0".
error Found incompatible module
```

do you have a hint what may cause that?

> I think it's better to use <h*> tags for specific headlines which you don't want to appear in the documentation

i'm not sure if there's a misunderstanding. this is not about excluding headings from a page's contents, but from the toc on the side. these are `a` elements in `li`s with the `md-nav__link` / `md-nav__item` class assigned. unless there'd be an option to collapse sub-items, i think the excluded items would bloat the html, in particular in cases where there are e.g. a lot of fourth order headings which ought not to be displayed.

> You changed the dependency for mkdocs to 0.17, which does not exist, thus the build fails.

this is anticipating that the feature will depend on a future release of `mkdocs`. should i point the dependency to my branch that includes the required change for now?

squidfunk:
Regarding the error, you are probably running a node version below 5.

I understood what you wanted to do, but I won't merge something that depends on a future release of MkDocs that isn't even released. Furthermore, I want to keep things backward compatible, as many, many people are probably running 0.16.x right now. Sorry, but for me this PR doesn't sound like a good idea, so I cannot merge it.

squidfunk:
Okay, but as this is a proof of concept which depends on a release of an upstream project that isn't even scheduled with a feature that isn't even merged, I would ask you to close it, because it is unclear how long it will float around here. You can re-open it later, when the feature is merged into MkDocs and then we can think about how we will tackle the problem.

funkyfuture:
okay. i amended some commits regarding the building you might want to cherry-pick.

squidfunk:
Thanks!

squidfunk:
I updated the contributing.md with a link to the build theme and symlinked the file in the docs to the top-level file - that was a good idea. The scripts are bound to change, will update them later. Closing this issue (for) now.

Issue #455, opened 2017-09-12
Bold Code And Normal Code Look The Same


Mehran-Baghi:
## Description

The result of these markdown lines are the same:
```
**`Bold Code`**
`Normal Code`
```

### Expected behavior

Fresh MkDocs Installation:
![mkdocs](https://user-images.githubusercontent.com/17176259/30332932-e2d7e052-97f0-11e7-983d-7a81c9659add.jpg)

### Actual behavior

Material for MkDocs:
![material](https://user-images.githubusercontent.com/17176259/30332962-f7bfb152-97f0-11e7-8de6-5c9a36dd0dc9.jpg)

### Steps to reproduce the bug

Use the markdown in the description.

### Package versions

* Python: `Python 2.7.13`
* MkDocs: `mkdocs, version 0.16.3`
* Material: `Version: 1.10.0`

### Project configuration
Not Relevant
``` yaml
site_name: My Docs
theme: 'material'
```

### System information

* OS: [Debian GNU/Linux]
* Browser: [Not Relevant]

facelessuser:
Project configuration, specifically extensions, may or may not be relevant.  I would post them anyways, at the very least the extension setup, and extra if you adding any custom CSS.

Most likely though, this is how styling is applied.  I assume a base style is applied to code blocks and then the provided Pygment's style augments the base.  I would provide the config regardless so it can be investigated further in case you are configuring things in an unexpected way.

I'm not sure I would consider wrapping the code block in bold and not getting bold a bug.  This may just due to an assumption that doesn't hold true with this theme.  Just because one theme handles things one way, does not mean all themes will handle things the same.

Mehran-Baghi:
Thank you @facelessuser for your time.
I have provided my configuration in issue report:
```yaml
site_name: My Docs
theme: 'material'
```
That is all of it.
Did you have any problems with reproducing?

> I'm not sure I would consider wrapping the code block in bold and not getting bold a bug. This may just be due to an assumption that doesn't hold true with this theme.

It depends on whether that assumption was a conscious choice or not. If the author consciously decided to make bold and normal code blocks look the same, I would be just disagree.
Otherwise I would suggest that `bold code blocks` be rendered as `bold code blocks`.

Mehran-Baghi:
[...]

PR #457, opened 2017-09-13
Fixing The Bold Code Block Issue

Mehran-Baghi:
046ef9f will fix the #455 .
It seems that the problem was enforcing the code font-weight to be 400 [here](https://github.com/squidfunk/mkdocs-material/blob/master/src/assets/stylesheets/base/_typeset.scss#L49).
Loading another google font weight is not necessary.
I have built the theme successfully.


Mehran-Baghi:
I'm not a web developer but I think we should get rid of `font-family` definitions in the `_typeset.scss` since we have already defined them in the `base.html`

Mehran-Baghi:
Also shouldn't we activate the [`font-feature-settings`](https://github.com/squidfunk/mkdocs-material/blob/master/src/assets/stylesheets/base/_typeset.scss#L37) by following them with either `1` or `on`? [Refrence](https://css-tricks.com/almanac/properties/f/font-feature-settings)

squidfunk:
The `font-family` definitions inside `_typeset.scss` are the defaults. If you set `extra.font = false`, the font is not defined inside `base.html`, so there needs to be a default definition. Also we have to check thoroughly if removing the font-weight completely does not cause any issues.

squidfunk:
Regarding the `font-feature-settings`: it's perfectly valid, [see](https://developer.mozilla.org/de/docs/Web/CSS/font-feature-settings)

squidfunk:
Fixed bold code in #458

PR #476, opened 2017-10-03
Feature/hidemenuitemasextraoption

reijoh:
I suggest this change to allow having markdown files that should be hidden from the menu. I needed this when I had articles that have linked sub articles and where I did not want the sub articles to show in the menu. This feature allow you to hide both menu items with children (folders with sub folders and all the markdown files within) and single page menu items. With this change, all you need to do is to update the mkdocs.yml with an extra option 'hidemenuitems' to hide menu items with specific title(s):

# Options
extra:
  hidemenuitems:
    - subarticles
    - media

### DIFF
--- a/src/partials/nav-item.html
+++ b/src/partials/nav-item.html
@@ -46,10 +46,10 @@
     <!-- Active checkbox expands items contained within nested section -->
     {% if nav_item.active %}
       <input class="md-toggle md-nav__toggle" data-md-toggle="{{ path }}"
-        type="checkbox" id="{{ path }}" checked />
+          type="checkbox" id="{{ path }}" checked />
     {% else %}
       <input class="md-toggle md-nav__toggle" data-md-toggle="{{ path }}"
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for your PR. However, this is not a very generic solution because some people may want to hide single articles or other levels. Material tried to be as generic as possible, so I cannot merge this, I'm sorry. I'll understand this to be a customization on your side.

reijoh:
Ok, I understand. Can I do something to make it more generic and if so what would be generic to you? Or is the feature of being able to hide menu items undesirable and not generic at all?

squidfunk:
There have been some requests for this feature, maybe try to use the search here to see what other people think. However, as far as I know there's also an open issue on the MkDocs tracker which would be the best vector to include this feature.

squidfunk:
Closing this because it is related to an open issue / feature request in an upstream project.

PR #488, opened 2017-10-12
Keep search open on enter without selection

lukasgeiter:
When hitting enter in the search and not having any of the results selected, the search window just closes. In my opinion this is definitely not optimal for the user experience. Even though hitting the enter key is not necessary here many users probably have it in their muscle memory to press enter after a search query.

This commit changes it so the search window in fact only closes if the user has selected a result. Otherwise it simply stays open.


squidfunk:
Great idea, thanks for your contribution!

PR #494, opened 2017-10-13
Nersc customization

brandongc:


brandongc:
wrong button in github UI, please ignore/delete

Issue #509, opened 2017-10-19
Call for Contributions: Add languages/translations to Material


squidfunk:
## Call for Contributions: translating Material for MkDocs

This is a call for contributions to add translations to Material for MkDocs. Default translations are a much requested feature. #507 finally adds the ability to set the language via `mkdocs.yml`:

``` yml
theme:
  language: 'de'
```

Our progress so far:

* [x] English `en` ([Source](https://github.com/squidfunk/mkdocs-material/blob/master/src/partials/language/en.html))
* [x] German `de` ([Source](https://github.com/squidfunk/mkdocs-material/blob/master/src/partials/language/de.html))
* [x] Danish `da` ([Source](https://github.com/squidfunk/mkdocs-material/blob/master/src/partials/language/da.html)) thanks to @jonas 
* [x] Polish `pl` ([Source](https://github.com/squidfunk/mkdocs-material/blob/master/src/partials/language/pl.html)) thanks to @makshh
* [x] Swedish `sv` ([Source](https://github.com/squidfunk/mkdocs-material/blob/master/src/partials/language/sv.html)) thanks to @martinbira 
* [x] Spanish `es` ([Source](https://github.com/squidfunk/mkdocs-material/blob/master/src/partials/language/es.html)) thanks to @aliciaphes 
* [x] Portguese `pt` ([Source](https://github.com/squidfunk/mkdocs-material/blob/master/src/partials/language/pt.html))  thanks to @argenta-web
* [x] French `fr` ([Source](https://github.com/squidfunk/mkdocs-material/blob/master/src/partials/language/fr.html)) thanks to @mvertes
* [x] Italian `it` ([Source](https://github.com/squidfunk/mkdocs-material/blob/master/src/partials/language/it.html)) thanks to @markainick
* [x] Norwegian `no` ([Source](https://github.com/squidfunk/mkdocs-material/blob/master/src/partials/language/no.html)) thanks to @oddcb
* [x] Chinese (Simplified) `zh` ([Source](https://github.com/squidfunk/mkdocs-material/blob/master/src/partials/language/zh.html)) thanks to @cxlwill
* [x] Chinese (Traditional) `zh-Hant` ([Source](https://github.com/squidfunk/mkdocs-material/blob/master/src/partials/language/zh-Hant.html)) thanks to @cxlwill
[...]

PR #518, opened 2017-10-20
Added swedish translation

martinbira:
References #509

### DIFF
--- a/src/partials/language/sv.html
+++ b/src/partials/language/sv.html
@@ -1,0 +1,42 @@
+<!--
+Copyright (c) 2016-2017 Martin Donath <martin.donath@squidfunk.com>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to
+deal in the Software without restriction, including without limitation the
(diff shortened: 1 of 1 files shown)


src/partials/language/sv.html
```diff
@@ -0,0 +1,42 @@
...
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+IN THE SOFTWARE.
+-->
+
+<!-- Translations: Swedish -->
+
+{% macro t(key) %}{{ {
+  "language": "sv",
```
makshh: Should't be "se"?

squidfunk: Yes, according to Wikipedia

squidfunk: Hmm, no wait. Lunr also uses `sv`, so let's leave it at that.

PR #530, opened 2017-10-23
add Simplified Chinese translation

cxlwill:


squidfunk:
Thanks! However, please make your changes in the `src` directory. The `material` directory is overwritten on every build.

squidfunk:
Is the language `zh` the correct identifier? I thought `cn`? Sorry, I'm not an expert in chinese ;-)

martinbira:
Language is `zh` general for China. You actually have a few different variants when dealing with both languages and locales. I think that zh_CN is the Simplified version, zn_TW is Traditional and zh_HK is Hong Kong.

cxlwill:
Yeah, I agree with @martinbira . Generally All Chinese is under `zh`. I'm not sure that sub-locale is allowed. If so, I can add  Traditional Chinese as well.

squidfunk:
Thanks for both translations, really awesome. However, the locale format `zh_CN` is not supported by HTML5, as this is the value that is inserted into the `html` tag. Can we switch to the valid/accepted codes? See https://www.w3schools.com/tags/ref_language_codes.asp

squidfunk:
... so we could offer the default (simplified?) under `zh` and traditional under `zh-Hant`.

cxlwill:
I agree with you.

squidfunk:
If you make the changes in the `language` attribute, I will merge your PR.

cxlwill:
Yeah, I set Simplified Chinese to `zh` and Traditional Chinese to `zh-Hant`, at filename and inner name.  I think everything is OK now.

jfcherng:
I have some words to say about this PR.

- The words chosen in that translation is for people who live in China. For example, Taiwanese doesn't use `搜索` and we use `搜尋` instead. But I know that zh-Han`t` is for Traditional, not for Taiwan. Would it possible to create variant as @martinbira said: `zh-TW`, `zh-HK` etc...?

- The translation `"clipboard.copied": "粘贴",` is shown when a visitor click the "copy-to-clipboard" button on a code block. It should be translated into `已复制(already copied)` rather than `粘贴(paste)`.

martinbira:
Probably not a bad idea if possible. When I worked with translations for the Sony Mobile web we hade zh-cn, zh-tw, and zh-hk as different versions of Chinese.

squidfunk:
`zh-TW` is not a valid code for the `lang` attribute of the `html` tag, see https://www.w3schools.com/tags/ref_language_codes.asp

cxlwill:
@martinbira Chinese language code for HTML keeps changing these years. It used to be zh-Cn kind of thing, which depends on users' locale. For lates zh-Hans kind, it consider language itself despite of using locale. So the connection looks like below:

- zh-CN → zh-CHS → zh-Hans → zh → Invariant
- zh-TW → zh-CHT → zh-Hant → zh → Invariant

cxlwill:
@jfcherng Actually I struggled about 搜索(search) and 搜尋(search) for quite a while. Do HKer use latter as well? I will change `already copied`, since it is purely my mistaken.

jfcherng:
@cxlwill I am not a HKer but I guess yes because both `yahoo.com.hk` and `google.com.hk` use `搜索`.

squidfunk:
Awesome work guys, thanks!

PR #545, opened 2017-10-30
Fix for note/seealso admonition title colors and icon if it is inside details

makshh:
This PR fixes the bug where `note/seealso` admonition which is inside `details` inherits admonition title styles (border, background and icon) from parent `admonition/summary`.

admonition-bug


squidfunk:
Thanks for the PR. I included it in a style-fix PR #549, there was one qualifier missing, see the PR. Should be fixed with the next release.

PR #554, opened 2017-11-01
Fixed meta title and description only rendering first character

squidfunk:


martinbira:
This may affect other meta information as well, see http://squidfunk.github.io/mkdocs-material/extensions/metadata/#__source

squidfunk:
Thanks for reporting!

PR #559, opened 2017-11-01
Fixed horizontal scrollbar between 1220px and 1234px

squidfunk:


src/assets/stylesheets/layout/_base.scss
```diff
@@ -37,6 +37,12 @@ $md-toggle__drawer--checked:
 html {
   height: 100%;
   font-size: 62.5%;
+  // Hack: some browsers on some operating systems don't account for scroll
+  // bars when firing media queries, so we need to do this for safety. This
+  // currently impacts the table of contents component between 1220 and 1234px
+  // and is to current knowledge not fixable.
+  overflow-x: hidden;
+  overflow-y: scroll;
```
makshh: Is this necessary? It will cause scrollbar even on pages with no content.

squidfunk: You're right.

PR #578, opened 2017-11-07
Generate Edit button only if page.edit_url

jszroberto:
Can you review this small change? I was planning to add custom urls for edition to mkdocs and I found out that your Edit Button wasn't generated if the `config.edit_uri` wasn't set.

### DIFF
--- a/material/base.html
+++ b/material/base.html
@@ -129,7 +129,7 @@
           <div class="md-content">
             <article class="md-content__inner md-typeset">
               {% block content %}
-                {% if config.edit_uri %}
+                {% if page.edit_url %}
                   <a href="{{ page.edit_url }}" title="{{ lang.t('edit.link.title') }}" class="md-icon md-content__icon">&#xE3C9;</a>
                 {% endif %}
                 {% if not "\x3ch1" in page.content %}
--- a/src/base.html
+++ b/src/base.html
@@ -242,7 +242,7 @@
               {% block content %}
 
                 <!-- Edit button, if URL was defined -->
-                {% if config.edit_uri %}
+                {% if page.edit_url %}
                   <a href="{{ page.edit_url }}"
                       title="{{ lang.t('edit.link.title') }}"
                       class="md-icon md-content__icon">&#xE3C9;<!-- edit --></a>


facelessuser:
@jszroberto, I imagine that is by design.  In the Mkdocs [documentation](http://www.mkdocs.org/user-guide/configuration/#edit_uri) it reads (emphasis mine):

> **When set**, and if your theme supports it, provides a link directly to the page in your source repository.

This implies to me when **not** set, it should not show the edit button.  Which makes sense as not everyone will want the edit button to show on their docs.

squidfunk:
@facelessuser absolutely true, I personally don't like it.

jszroberto:
@facelessuser I agree on that but setting config.repo_uri means that page.edit_url is also set and passed to the template. Consequently the result is the same.

[page.edit_url Documentation](http://www.mkdocs.org/user-guide/custom-themes/#pageedit_url)

If you check change I introduced, you will see that it doesn't make any difference for the scenario you exposed from an end user perspective.

facelessuser:
@jszroberto touché :slightly_smiling_face:.

squidfunk:
So can you explain what difference it would make? First: which version of MkDocs and Material are you targetting? `edit_uri` behavior changed with 0.17.1 - it must now be explicitly reset to `""` in order for the button to disappear, so I don't see any problem here.

jszroberto:
Latest for both. 

But the point is that I am going to PR mkdocs so that the support custom edit_uris for certain docs. I am targeting using mkdocs for multiple repositories and I want to be able to redirect to any of them in case someones wants to edit a page. 

Your theme basically doesn't allow to set "page.edit_uri" only for certain pages, because it needs the global `config.edit_uri`.

*  If `config.edit_uri` is nil, then no custom `page.edit_uri` is displayed
* If `config.edit_uri` is not nil, it generates the `page.edit_uri` with a wrong path because it points to `repo_url`, which doesn't necessarily contain a desired page.

funkyfuture:
fyi, i had a conversation w/ @waylan where he stated that the `edit_uri` stuff is essentially broken (one undesired outcome is addressed here) and the functionality shall be moved to a plugin with the 1.0 release.

squidfunk:
Makes sense. Merged. Thanks! Let's hope they fix in in 1.0.

PR #583, opened 2017-11-08
https link

coliff:


coliff:
related question... why does the link have a ` title="Material for MkDocs"`?

Seems unnecessary when the text in the link is Material for MkDocs. Why does it need to show Material for MkDocs as a tooltip when hovering over the link?

squidfunk:
Actually, I don't know. Probably I'm used to it because W3C validator complains all the time, but you're absolutely right  it's useless. Feel free to remove redundant titles and PR.

PR #585, opened 2017-11-10
Remove unnecessary title tag from links in footer

coliff:
Remove unnecessary title tag from links in footer


squidfunk:
Please make your changes inside the `src` and `material` folder or they will be overwritten by the next build.

PR #593, opened 2017-11-15
Better keyboard accesibility

makshh:
If you have any questions please feel free to ask.

I know it's chaos and it will take you some time to understand what's going on here, but I don't have too much time to make it better ;)

### DIFF
--- a/material/assets/javascripts/script.js
+++ b/material/assets/javascripts/script.js
@@ -55,7 +55,7 @@
 
   /**
    * --------------------------------------------------------------------------
-   * // Make TOC label (Table of contents) not focusable
+   * Make TOC label (Table of contents) not focusable
    * --------------------------------------------------------------------------
    */
 
@@ -118,8 +118,6 @@
 
   function mediaLarge(mql) {
     if (mql.matches) {
-      // Make site title not focusable on bigger screens
-      siteTitle.tabIndex = -1;
       // When screen is big and you click on the site title (Material for MkDocs in primary nav - it opens the drawer because of for attribute)
       siteTitle.setAttribute('for', '');
       Array.prototype.forEach.call(linksDrawer, function(el) {
@@ -128,6 +126,10 @@
       Array.prototype.forEach.call(focusableElements, function(el) {
         el.tabIndex = 0;
       });
+      // Make site title not focusable on bigger screens
+      siteTitle.tabIndex = -1;
+      // Make TOC label (Table of contents) not focusable
+      navTitles[navTitles.length - 1].tabIndex = -1;
       document.removeEventListener('focus', focusPrimary, true);
     } else {
       siteTitle.setAttribute('for', 'drawer');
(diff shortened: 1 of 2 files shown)


squidfunk:
Okay, great! This will take me some time, but it's good base work to finally solve #400.

makshh:
1. Maybe search reset button should be focusable using keyboard?
2. There is a lot of code, but this is because of the fact that you use labels for submenus instead of anchors. Maybe it would be easier to change HTML structure and some CSS instead of writing a lot of JavaScript.
3. It works a lot better with this fixes.

makshh:
Or maybe we should use something like [ally.js](https://allyjs.io). There are also great tutorials.

squidfunk:
Thanks for your work. Just let me know when you're finished. If ally.js makes sense and reduces the code we have to write, we could use it (if it's not 100kb payload). Would it be possible for you to elaborate?

makshh:
I'll look into it. It's about 75kB minified. It won't reduce the amount of code too much, but it can help with achieving the same effect in different browsers. I've had hard time to make it work the same for example in Chrome and Firefox, because they handle focus differently.

squidfunk:
75kb minified is way too much. The whole Material code currently is 100kb minified. I think we should first adopt your solution and invest time, if it's necessary in the future.

makshh:
Ok, I'll let you know when it's finished.

makshh:
Finished basic things, now only drawer 👍

makshh:
Okay, finished.

I have tried my best, I hope now you will be able to understand it with ease (improved code and comments).

1. Added skip to main content anchor.
2. Added focusedElementBeforeSearchOrDrawerOpened to focus on last focused element when closing search or drawer using keyboard.
3. When there is hash in URL - focus on headerlink related to hash and scroll to that section (with animation - opinionated) on page load, polyfill for scroll behaviour provided.
4. Toggle menu on M key (opinionated).
5. Allow to focus on reset button in search using keyboard.
6. `md-state="active"` on focus event for search input - thanks to that you don't have to TAB twice to focus on reset button sometimes.
7. Blur collapsible nav links on `mouseup` to avoid focus styling on labels.
8. Support for tabs (skip to content link is properly positioned).
9. You can't focus on search input using TAB key on smaller screens (focus only if search is truly opened) - it prevents from fetching JSON using TAB key even when you don't see search input.
10. Search results are not focusable when search is hidden.
11. When there are no search results - using TAB/SHIFT + TAB will close search and focus on previous/next element.
12. Close drawer on ESC.
13. Open search and drawer with Space/Enter/Return ocusing on icons in header on smaller screens.
14. Proper focus in drawer **(using only data attributes and CSS!).**
15. In drawer you can use horizontal arrows keys to navigate (left arrow = back to previous menu, right arrow = go to next menu when there is one on focused element).
16. Closing drawer when opening search with F/S keys.

How it works:

1. I'm adding tabindexes only to specific labels.
2. The rest is done in CSS (changing visibility to make things focusable/not focusable).

How to test it?

Just remove main `keydown` listener from your `application.js` file and include my JS and CSS files.

Tested on:

Mac: Chrome (latest), Firefox Quantum (latest)
Windows: Chrome (latest), Firefox Quantum (latest), IE11 (partially)

Problems:

1. When someone will be editing HTML markup it MAY not work as expected (I didn't think about it in details).
2. Firefox (Mac) - sometimes divs get focus when tabbing - not a big issue.
3. I don't see outline/border in IE11 (I don't know if you support that browser).

squidfunk:
Thank you very much for your work, it really has been a good guideline. Please see #659 for notes on what I have implemented and what is still an issue. I think the last PR should greatly improve accessibility, but yeah, still some stuff to do. I will definitely come back to this PR for reference.

PR #596, opened 2017-11-18
Improve accessibility

squidfunk:


makshh:
You don't have to add tabindexes to anchors. It can be done only in CSS.

squidfunk:
Closing in favor of #659

PR #615, opened 2017-11-26
Update mkdocs_theme.yml

amabdallah:
Please add a URL to where users could get/see ligature codepoints


squidfunk:
Added in #621.

PR #626, opened 2017-12-06
Added dutch translations

Elvander:


Elvander:
Oops, forgot that one

PR #659, opened 2018-01-11
Improve accessibility

### LINKED ISSUE #400: Search bar takes tab focus

squidfunk:
This PR fixes some of the issues with <kbd>TAB</kbd> navigation, improving accessibility.

The following issues have been addressed:

- [x] Added focus state for clipboard buttons
- [x] Fixed #400: Search bar steals tab focus
- [x] Fixed search not closing on <kbd>ENTER</kbd> when result is selected
- [x] Fixed search not closing when losing focus due to <kbd>TAB</kbd>
- [x] Fixed collapsed navigation links getting focus
- [x] Fixed outline being cut off on <kbd>TAB</kbd> focus of navigation links
- [x] Fixed bug with first search result navigation being ignored
- [x] Removed search result navigation via <kbd>TAB</kbd> (use <kbd>&#8593;</kbd> and  <kbd>&#8595;</kbd>)
- [x] Removed `outline` resets for links
- [x] Improved general tabbing behavior on desktop

The following issues have been discussed in #400 and have not (yet) been adressed:

- [ ] Missing submit button in the search - this will probably never be implemented, as it really doesn't go well with the way search works - it's a "search-as-you-type" dialogue that submits instantly
- [ ] Three labels for search (two buttons and the modal) are not very accessible, but are necessary, because all opening dialogues and modals are currently implemented with the Checkbox hack. This could in theory be dropped for search, as it doesn't work without JavaScript anyway (and is not shown), but it would deviate from the way the theme handles visibility and state

The following improvements suggested by @makshh in #593 have not (yet) been included:

- [ ] Explicitly set focused element depending on context when search is closed
- [ ] Support for <kbd>TAB</kbd> navigation inside drawer on tablet and mobile
- [ ] Support for opening search when focusing on search icon on tablet and mobile
- [ ] Support for opening drawer when focusing on menu icon on tablet and mobile
- [ ] Toggle nested menues in navigation with <kbd>SPACE</kbd> (this will be done)
- [ ] Toggle drawer with <kbd>M</kbd> and close it with <kbd>ESC</kbd>

This should greatly improve accessibility, even though it doesn't solve all problems. However, it's a start and I may continue work on the remaining issues at some point in the future.


makshh:
That's a great start! I don't have too much time now, but when I have I will try to test it and give you my thoughts if you don't mind. Cheers!

squidfunk:
Great idea!

I think it's more than time to say thank you @makshh for all the great work you have done on Material, starting with answering issues and questions up to the accessibility improvements you made. Awesome work! Let's keep it up.

squidfunk:
Skip to content was implemented in #666

PR #681, opened 2018-01-26
Add ability to disable Disqus via page meta

jfcherng:
I just don't want to let people leave public comments on certain pages. For example, an `About Me` page, which I usually leave other method to contact me such as email.

Wondering people may want to disable Disqus on certain pages like me.

### DIFF
--- a/material/base.html
+++ b/material/base.html
@@ -153,7 +153,7 @@
                 {% endblock %}
               {% endblock %}
               {% block disqus %}
-                {% if config.extra.disqus and not page.is_homepage %}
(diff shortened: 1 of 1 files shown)


squidfunk:
That's a great idea, but you made your changes in the `material `folder which will get overwritten by the next build. I will think about a way to include this. For the time, I will leave this PR open as a reminder.

squidfunk:
Fixed in Added ability to override Disqus integration using metadata, added ability to override the Disqus configuration. To disable Disqus on specific pages just set `disqus` in metadata to nothing:

```
disqus:
```

PR #684, opened 2018-01-28
Year fixed in the change log.

xorander00:
Not a big deal, but I was taking a look at the change log and was confused by the 2017 for a minute.  Figured I might was well create this PR for quick housekeeping.  Thanks for the work, it's appreciated! :)


squidfunk:
Oh, thanks, stupid copy&paste error 😅

PR #690, opened 2018-02-01
Duplicate slashes in source file URLs

MarinBernard:
Hi,

There is an issue with source URLs built by the template from `source` and `path` page metadata. The generated URLs include double slashes between the repository URL and the rest of the path. I don't know for Github, but Bitbucket is unable to handle such URLs correctly.

The source of the issue is that the `config.repo_url` variable is normalized by mkdocs to always include a trailing slash. As a consequence, the `join()` filter adds a trailing slash where it is not needed. 

The attached pull request excludes the `config.repo_url` variable from the `join()` operation, assuming repo URLs always include a trailing slash thanks to mkdocs normalization.

Thanks,

Marin.

### DIFF
--- a/material/base.html
+++ b/material/base.html
@@ -146,7 +146,9 @@
                     <h2 id="__source">{{ lang.t("meta.source") }}</h2>
                     {% set path = page.meta.path | default([""]) %}
                     {% set file = page.meta.source %}
-                    <a href="{{ [config.repo_url, path, file] | join('/') }}" title="{{ file }}" class="md-source-file">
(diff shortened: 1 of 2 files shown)


squidfunk:
My version of mkdocs, 0.17.2, does exhibit this behavior. However, I will provide a fix which will account for double slashes.

squidfunk:
Fixed in b3e293ece91e0ba4d3f81fc93aaca551d2358051. My version of MkDocs doesn't normalize the name, so I included code to handle both cases.

Issue #712, opened 2018-02-22
Preconnect to https://fonts.gstatic.com for improved performance


coliff:
I've started adding: `<link href="https://fonts.gstatic.com" rel="preconnect" crossorigin>
` to the head (just above the linked CSS files) on all the sites I manage which load Google Fonts. It makes a slight improvement to the page load speed.

There's a good article about it here:
https://www.cdnplanet.com/blog/faster-google-webfonts-preconnect/

Would you be interested in adding this to mkdocs-material ?

squidfunk:
Sure, feel free to PR!

squidfunk:
Fixed in 36ec35c2e47bf0bed8de75ca658cab2017c10d29, should be released soon.

squidfunk:
Released as part of 2.6.6.

PR #713, opened 2018-02-22
Add Google Font Preconnect

coliff:
Fixes: #712
Add <link href="https://fonts.gstatic.com" rel="preconnect" crossorigin> to the head makes a slight improvement to the page load speed.

There's a good article about it here:
https://cdnplanet.com/blog/faster-google-webfonts-preconnect


squidfunk:
You need to make the change in the `src` folder. However, it's only one line, I'll do it. Thanks again!

squidfunk:
Fixed in 36ec35c2e47bf0bed8de75ca658cab2017c10d29

PR #723, opened 2018-03-02
Refactoring: TypeScript/RxJS

squidfunk:
This PR will greatly improve modularity of the JavaScript included with Material, as it will make initialization of components modular and easily adjustable. Furthermore it will be possible to subscribe to state changes of components, so that component behavior can easily be extended.

For this reason, the code is rewritten in TypeScript using RxJS as a basis for handling of asynchronous events, as it is a perfect fit for the browser. It may take some time, but you are free to come back to this PR to get updates on the state.


squidfunk:
Closing for now due to a lack of time.

Issue #663, opened 2018-01-14
Structured Data, Facebook Open Graph and Twitter Cards support


yogeshbeniwal:
This is a feature/enhancement suggestion for this theme. 

It will be good to enrich page with below listed metadata:

Structured Data: https://developers.google.com/search/docs/guides/intro-structured-data. Preferably Organisation, Website, Article and Author
Facebook Open Graph: https://developers.facebook.com/docs/sharing/webmasters#markup
Twitter Cards: Twitter Summary Card with large Images https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/summary-card-with-large-image

squidfunk:
Thanks. Could you post in this thread which tags you exactly would like to integrate?

makshh:
This is what I'm using in my docs. I think every site is different and it will be hard to make it universal.

```html
<!-- Extrahead -->
{% block extrahead %}

  <meta property="og:url" content="{{ page.canonical_url }}">
  {% if page and page.meta and page.meta.title %}
    <meta property="og:title" content="{{ page.meta.title }}">
  {% elif page and page.title and not page.is_homepage %}
    <meta property="og:title" content="{{ page.title }} - {{ config.site_name }}">
  {% else %}
    <meta property="og:title" content="{{ config.site_name }}">
  {% endif %}
  <meta property="og:description" content="{{ config.site_description }}">
  <meta property="og:image" content="path/to/image.jpg">
  <meta property="og:image:alt" content="Image alt">
  <meta property="og:image:type" content="image/jpeg">
  <meta property="og:image:width" content="1200">
  <meta property="og:image:height" content="630">

  <meta name="twitter:card" content="summary_large_image">
  <meta name="twitter:site" content="@username">
  <meta name="twitter:creator" content="@username">
  {% if page and page.meta and page.meta.title %}
    <meta property="twitter:title" content="{{ page.meta.title }}">
  {% elif page and page.title and not page.is_homepage %}
    <meta property="twitter:title" content="{{ page.title }} - {{ config.site_name }}">
  {% else %}
    <meta property="twitter:title" content="{{ config.site_name }}">
  {% endif %}
  <meta name="twitter:description" content="{{ config.site_description }}">
  <meta name="twitter:image" content="path/to/image.jpg">
  <meta name="twitter:image:alt" content="Image alt">

{% endblock %}
```

I'm not experienced in `schema`.

squidfunk:
[...]

PR #750, opened 2018-03-31
Add more meta tags

PizzaFox:
This PR adds [Open Graph](http://ogp.me/) and [Twitter card](https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/abouts-cards.html) meta tags.

You may want to change the image tags. They currently use something called `detailedLogo` because the `logo` image I had was different than what I wanted for those tags.

### DIFF
--- a/material/base.html
+++ b/material/base.html
@@ -9,12 +9,18 @@
       <meta charset="utf-8">
       <meta name="viewport" content="width=device-width,initial-scale=1">
       <meta http-equiv="x-ua-compatible" content="ie=edge">
+      <meta property="og:type" content="website">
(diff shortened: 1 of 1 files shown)


squidfunk:
1. You made your changes in the `material` folder which is the dist, so they are not persisted. Please read the contribution guidelines before contributing.
2. Meta tags like OpenGraph and Twitter are highly case-dependent. For your specific case `<meta property="og:type" content="website">` may be suitable, but for others it is not. For this reason everything would have to be made configurable and this – after a thorough analysis – yields to many edge cases to be taken care of. See #663 for more directions.

PR #753, opened 2018-04-08
Fix MS Edge details issue

facelessuser:
Fixes issue where details are not rendered properly in MS Edge due to the fact that details tags are not styled as block elements.


squidfunk:
Oh, that addition looks to be absolutely necessary - wonder how it got forgotten. Too many browsers, too many cases. Thanks!

squidfunk:
Released as part of 2.7.2

PR #783, opened 2018-05-25
Use HTTPS for links to www.mkdocs.org

edmorley:
To prevent the unnecessary HTTP 301 (and reduce the redirects noise when running the W3C link checker tool against generated pages).

Also fixes the anchor fragment for `#using-the-theme-custom_dir`.


squidfunk:
Thanks!

PR #792, opened 2018-06-09
Fix ids not being unique enough in templates

facelessuser:
Fixed ids not being unique enough. Ids are now prefixed with `__` per our discussion.

I believe I've tested everything **except** Disqus support. I changed the id to `__disqus_thread`, but I did not see anyone referencing it, maybe it is referenced by the JS libraries that are meant to provide the support? If Disqus changes are not sufficient, I'd need a little direction in how to finish implementing it properly.

Image references work, table of content works, and I believe search is working fine. If you see something I missed, please let me know.


squidfunk:
Thanks for your PR! The `disqus_thread` must remain as-is because it is necessary for the Disqus SDK to work.

Issue #756, opened 2018-04-11
Update Font Awesome to 5.0 – breaking change


Sharom:
Please, update Font Awesome to version 5.0.

[Gitter](https://gitter.im) icon was added since 5.0 ;)

squidfunk:
This would break the Material integration because now there are multiple prefixes `.fa`, `.fab` etc. which Material doesn't know about, so configuration would change which in turn would mean an entirely new release. I will keep this in mind for 3.0.0 but it's probably not gonna happen anytime soon.

Sharom:
Ok, thank your for the answer.

squidfunk:
As I understood, the `.fa` prefix now is not universal but sometimes an icon needs the `.fa`, sometimes the `.fab` prefix. This would mean we need to change the way social links are configured in `mkdocs.yml`, as currently only the second part of the Font Awesome Glyph (i.e. `github` for `.fa .fa-github`) needs to be defined. This means the user has to adjust their `mkdocs.yml` to `.fab .fa-github` (for example), or am I mistaken?

squidfunk:
Hmm, so if we don't change the way it's integrated we limit usage to the brand icons. What if somebody wants to use another icon like the globe for a blog? That wouldn't be possible without extending the theme.

k3rn31p4nic:
Yeah, if they want to use any other icons besides the brand icons, that wouldn't be possible with this way.

Another way to do this is by adding another option to `mkdocs` configuration file, to check if it's a brand icon, like:
```yml
extra:
  social:
    - type: 'globe'
      link: 'https://example.com/blog'
      brand: false
    - type: 'twitter'
      link: 'https://twitter.com/...'
      brand: true
```
Then in the `social.html` file we can use the `.fab` prefix if `brand` is set to `true` in the `mkdocs` config, otherwise we can use either of `.fas`, `.far` or `.fal` prefix (all those have same set of icons, but with different font weight). This would allow users to use all the icons offered by Font Awesome.

proofit404:
Hi,

How could I help to make this happen?

Personally, I prefer to infer prefix from the `type` option in `mkdocs.yml`. So we will keep all the magic under the hood without changing any existed project configuration. Maybe it'll require some Python code.

Regards,
Artem.

squidfunk:
Thanks for your offering. We haven't really settled on an implementation yet. If you want to go ahead on this a good start would be to draft a proposal which addresses the following topics:

[...]

PR #797, opened 2018-06-13
Update CDN's URL for font-awesome v5.0.0

rhincodon-org:
I updated the code with reference to this site.
https://fontawesome.com/how-to-use/upgrading-from-4
If it's OK, please merge the code.


squidfunk:
We cannot upgrade without changing the configuration API, please see #756

kotakotakotakota:
Thank you for your review. I’m sorry to miss the discussions. 
I understand this code doesn’t satisfied your needs:(without changing the configuration API). 
The PR is backward compatibility, only default use case “fab” prefix. In case of using other prefix, we need to add “prefix: ‘far’” sentence. 
See you next.

PR #800, opened 2018-06-18
Fix highlight tables

facelessuser:
When codehilite is disabled, or if pymdownx.highlight is used instead of codehilite, the default class "highlight" is used instead of "codehilite".  This also applies to the number table format, and will use "highlighttable" instead of "codehilitetable".


squidfunk:
Awesome, thanks for fixing. Really hard to keep track of those many combinations of plugins.

squidfunk:
Released as part of 2.9.1

facelessuser:
> Awesome, thanks for fixing. Really hard to keep track of those many combinations of plugins.

Yeah. I'm basically dropping use of Codehilite in my documents (I never use indented code blocks the way Codehilite wants me to use them), and it is kind of weird to set my class to the name of a plugin that I'm not using 🙂. I've been meaning to patch this for a while, but finally got around to it.

PR #802, opened 2018-06-19
Opening edit link in new tab

NickBusey:
I serve my project's MkDocs inside an iframe and GitLab prevents itself from being iframe'd, so when you click edit you get a nice big `Blocked by Content Security Policy` page.

Plus, I feel like this is a better user workflow anyway. This way you still have the original tab open to refer back to if needed while making edits.

### DIFF
--- a/src/base.html
+++ b/src/base.html
@@ -272,6 +272,7 @@
                 {% if page.edit_url %}
                   <a href="{{ page.edit_url }}"
                       title="{{ lang.t('edit.link.title') }}"
+                      target="_blank"
                       class="md-icon md-content__icon">&#xE3C9;<!-- edit --></a>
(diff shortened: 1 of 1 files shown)


NickBusey:
Alright, fair enough. Guess I'll just have to maintain my fork then.

squidfunk:
You don't have to fork the theme, you can override the respective content block using [theme extension](https://squidfunk.github.io/mkdocs-material/customization/#extending-the-theme). Just override the `content` block with your own implementation (if it's only for the edit link):

https://github.com/squidfunk/mkdocs-material/blob/60d8b01468047fd5e3e716b318c21db05ebd3402/src/base.html#L269-L306

PR #817, opened 2018-07-11
Adding alternative title for logo in header.html

dduportal:
Hello dear maintainer!

This pull request introduces the following change: adding an alternative title for the logo's image in the header and in the navigation bar.

The "alt" is explaining the intent since this is a link to the homepage with an image as link's content.

Ref. https://www.w3schools.com/tags/att_img_alt.asp 

> Guidelines for the alt text:
> The text should describe the image if the image contains information
> The text should explain where the link goes if the image is inside an <a> element
> ...

=> The initial goal is to have a valid HTML when using mkdocs with material them, against w3c compliancy which raise an error.

### DIFF
--- a/material/partials/header.html
+++ b/material/partials/header.html
@@ -6,9 +6,9 @@
           {% if config.theme.logo.icon %}
             <i class="md-icon">{{ config.theme.logo.icon }}</i>
           {% elif config.theme.logo.startswith("http") %}
-            <img src="{{ config.theme.logo }}" width="24" height="24">
+            <img src="{{ config.theme.logo }}" alt="Link to homepage with theme's logo" width="24" height="24">
           {% else %}
-            <img src="{{ base_url }}/{{ config.theme.logo }}" width="24" height="24">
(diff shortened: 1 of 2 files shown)


squidfunk:
Thanks for your PR. I understand that this is a validation error, but it's actually intended. The `alt` text will be shown while the image is loading or if it fails to load, which will break the design because it's far longer than the image is wide. This can be observed by setting the icon to a glyph/ligature on a slow connection.

Furthermore, there are more W3C validation errors, see [the results](https://validator.w3.org/nu/?doc=https%3A%2F%2Fsquidfunk.github.io%2Fmkdocs-material%2F), as Material needs to introduce some hacks because of some strange browser quirks.

dduportal:
Hi @squidfunk , thanks for the explanation.

Since w3c is an effort to bring interoperability between web browsers, do you have any plan to make the theme compliant to ensure the future of this theme?

I'm going to close this PR since it does not fits your needs, but we are thinking changing things manually or even moving away, because the "alt" issue could impact users with specific input devices.

squidfunk:
Material is as W3C compliant as possible with an emphasis on browser/rendering compatibility. I understand that there are some warnings or errors, but software development is always a tradeoff.

PR #819, opened 2018-07-16
Fix dependency conflict on tornado with latest mkdocs

mihaipopescu:
I'm getting `error: tornado 4.5.3 is installed but tornado>=5.0 is required by {'mkdocs'}` there is a conflict on tornado version dependency between `mkdocs-material` and `mkdocs`.

One way to fix it, is to try to lift off temporary workaround which was used to fix a travis build.
Another way would be to limit the version to `mkdocs<=0.17.5`.

Let's see what travis thinks about this.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,6 +22,3 @@
 mkdocs>=0.17.1
 Pygments>=2.2
 pymdown-extensions>=4.11
-
-# Temporary fix for build errors on Travis
-tornado<5


mihaipopescu:
@squidfunk looks like that worked. How easy would be to make a new release with this in?

On the other hand, I have no idea why python thinks that it's a good idea to install an alpha version of a package.

mihaipopescu:
Just wanted to let you know that I managed to fix this in another way locally:

I just need to make sure I installed:
"mkdocs==0.16.3"

squidfunk:
Material depends on MkDocs >= 0.17, so that's not a solution. If you find a fix for the current version feel free to update this PR

mihaipopescu:
Well, the proposed fix is in the PR. It just removes that workaround and travis looks happy.

mihaipopescu:
But obviously for me I had to downgrade a lot of packages to make it work. If you merge this PR, I'll be able to revert my workaround.

squidfunk:
Ok, sorry, I haven't looked in the changes, so you just removed the Tornado dependency and it seems to build. I will check it when I find the time. I only read your comment on `install_requires`.

mihaipopescu:
If you look at the changes, you will see that it removes 3 lines. I don't change the dependencies in this package. I just wanted to let you know that I found a local workaround for my package. That's all :)

edmorley:
I was hoping to try the new mkdocs `1.0b1` beta, however running `mkdocs serve` currently fails when using mkdocs-material (TIL using mkdocs use of `pkg_resources` means runtime dependency checks take place):
```
pkg_resources.VersionConflict: (tornado 5.1 (.../.virtualenvs/mkdocs/lib/python3.6/site-packages), Requirement.parse('tornado<5'))
```

To help with the review of this PR, I did a little digging...

So it looks like the hard-coding of tornado version was previously required, since mkdocs <= `0.17.2` didn't put an upper bound on the tornado version:
https://github.com/mkdocs/mkdocs/blob/0.17.2/setup.py#L64

...which would have resulted in errors once Tornado 5 was released.

However mkdocs v0.17.3 added an upper bound to the tornado version, making the workaround in `mkdocs-material` unnecessary:
mkdocs/mkdocs#1428
https://www.mkdocs.org/about/release-notes/#version-0173-2018-03-07

As such, it seems that it should be safe for this PR to remove the tornado line from `requirements.txt`, so long as the mkdocs dependency is also increased from `mkdocs>=0.17.1` to `mkdocs>=0.17.3`.

squidfunk:
I removed the Tornado dependency and set mkdocs to >=0.17.3 as recommended by @edmorley in 9205a7a - Thanks for your contributions, guys!

PR #822, opened 2018-07-17
Added icon link to home

bkbncn:
In desktop mode, the icon at header has a link to home.

However, in mobile mode, the icon at nav doesn't.

This PR added icon link to home, in mobile mode.


squidfunk:
Thanks for your PR. The initial ideas was that the panel can be closed by tapping on the logo but actually your suggestions makes much more sense. However, I had to change some stuff to make it work as it should, as part of 7f3c72b

PR #838, opened 2018-08-05
Search result item go to correct page.

netoht:
When I click in any link from search result, I'm redirected to a page failed.


squidfunk:
Thanks for the bug report and fix! Will make a bugfix release immediately.

squidfunk:
Released as part of 3.0.1

PR #848, opened 2018-08-13
fix: disable F/S when inside a textarea

Ir1d:
fix: disable F/S when inside a textarea

----

Hi, I used gitment(instead of disqus) for commenting, and the comments were displayed in a textarea.
In the current settings, I could not type s/f in the comments due to the hotkey.
Hope this fixes.

link for reproduce: https://oi-wiki.cf/introduction/softwares/#__comments

### DIFF
--- a/src/assets/javascripts/application.js
+++ b/src/assets/javascripts/application.js
@@ -420,9 +420,10 @@ function initialize(config) { // eslint-disable-line func-style
         }
 
       /* Search is closed and we're not inside a form */
-      } else if (document.activeElement && !document.activeElement.form && (document.activeElement.tagName !== "TEXTAREA")) {
+      } else if (document.activeElement && !document.activeElement.form
+                 && (document.activeElement.tagName !== "TEXTAREA")) {
 
-        /* F/S: Open search if not in input field or a textarea*/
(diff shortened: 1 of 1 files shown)


squidfunk:
Is this really necessary? A text field should be inside a form and that should already be handled by the code. Please provide a reproducible example where this logic actually breaks.

Ir1d:
@squidfunk Im not an expert in HTML so not sure whether necessary. Besides, gitment requires login to make comments(because it creates comments on the corresponding issue), so no other example available.

@max-ci Thx for informing, I'll look into this.

Ir1d:
upd: I've managed to fix this by [customizing the gitment render](https://github.com/24OI/OI-wiki/blob/master/static/disqus.html#L13-L26) and the [gitment.js](https://github.com/24OI/24oi.github.io/blob/master/gitment.js#L3063).

Ir1d:
I found that it's ok to wrap a form with `onsubmit="return false;" for gitment/gitalk.

example: https://github.com/24OI/OI-wiki/blob/master/static/disqus.html#L17

so I guess this is not needed anymore. thx guys.

PR #890, opened 2018-10-01
Update base.html to intergrate facebook comment

ngtrian:
intergrate facebook comment

### DIFF
--- a/material/base.html
+++ b/material/base.html
@@ -165,6 +165,9 @@
                   {% endif %}
                 {% endblock %}
               {% endblock %}
+              {% block fb_comment %}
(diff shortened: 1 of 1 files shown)


squidfunk:
Why would we want a Facebook integration?

ngtrian:
because facebook popular. And disqus is block in my country

squidfunk:
Okay, I see. However, I would consider this being an edge case and actually don't want to support another commenting system. Material brings a lot of features and adding more means also more potential stuff that could break. For this reason I cannot accept this PR. Also, please don't open a separate PR for every file.

PR #900, opened 2018-10-17
add support for redirects via metadata extension

kcgthb:
This PR proposes an easy way to create redirects from existing documents to other places (other documents or any URL, really). Which can prove useful when navigation tree reorganisation is needed.

It does so by adding a check for `page.meta.redirect` in `main.html` and then generating a simple js location.href / meta-refresh snippet to redirect the current page to the URL specified in the `redirect:` tag.

So for instance, if a `old_doc.md` document contains this:
```
redirect: /new/doc
```
accessing https://server/old_doc  will redirect to https://server/new/doc

### DIFF
--- a/src/main.html
+++ b/src/main.html
@@ -22,12 +22,12 @@
 
 {% if page.meta.redirect %}
 <script>
-    document.location.replace({{ page.meta.redirect }})
+    // redirect anchors
+    var anchor = window.location.hash.substr(1);
+    location.href = '{{ page.meta.redirect }}' + (anchor?'#'+anchor:'');
 </script>
-<meta http-equiv="refresh" content="0; url={{ page.meta.redirect }}"/>
-<link rel="canonical" href="{{ page.meta.redirect }} " />
-{% else %}
-
-{% extends "base.html" %}
-
+<!-- no-js fallback -->
+<meta http-equiv="refresh" content="0; url={{ page.meta.redirect }}">
+<link rel="canonical" href="{{ page.meta.redirect }}">
 {% endif %}
+{% extends "base.html" %}
--- a/src/main.html
+++ b/src/main.html
@@ -29,5 +29,6 @@
 <!-- no-js fallback -->
 <meta http-equiv="refresh" content="0; url={{ page.meta.redirect }}">
 <link rel="canonical" href="{{ page.meta.redirect }}">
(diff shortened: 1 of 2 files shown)


squidfunk:
Nice idea! However, does this actually perform a 301 redirect?

kcgthb:
Mmh, that's a good question, but I don't think 301 redirects can be done client-side. 

What this PR does is a combination of:
* javascript redirect by setting `location.href`
* HTML meta-refresh (which I think Google considers as a 301)
* canonical URL to avoid duplicate indexing

So, although that's not a perfect 301 redirection, I think that's the best that can be done in a pure client-side setting. But if you can think of another way, that'd be great!

squidfunk:
The problem is: if it's a non-301 redirect, it's not semantically correct. Can you do some research if search engines (i.e. Google) actually consider this as a 301 redirect? It would be really convenient and nice feature, I totally agree, but if it comes with problem it's not worth encouraging non-best practices.

coliff:
I think this looks like a good feature to add. The Hugo static-site generator does something similar. Maybe for the pages which direct you could discourage search engines from indexing those pages by adding `<meta name="robots" content="noindex">` so it's like this:

```
<meta http-equiv="refresh" content="0; url={{ page.meta.redirect }}">
<link rel="canonical" href="{{ page.meta.redirect }}">
<meta name="robots" content="noindex">
```

kcgthb:
This is a good suggestion, thanks!

src/main.html
```diff
@@ -20,4 +20,15 @@
   IN THE SOFTWARE.
 -->
 
+{% if page.meta.redirect %}
+<script>
+    // redirect anchors
+    var anchor = window.location.hash.substr(1);
+    location.href = '{{ page.meta.redirect }}' + (anchor?'#'+anchor:'');
+</script>
+<!-- no-js fallback -->
```
squidfunk: Could we use the `no-js` fallback as the main solution?

squidfunk: Nevermind, just read the impact on browser history. Ignore this comment.

squidfunk:
I pulled your code into the `base.html` template - this seems to be a better place. Thanks for your contribution and a big sorry it took me so long to review and merge it!

kcgthb:
No worries and thank you very much!

PR #926, opened 2018-12-19
lowercase doctype in HTML template

coliff:
This gzip compresses slightly better (can save 2 bytes!)

REF:
https://encode.ru/threads/1889-gzthermal-pseudo-thermal-view-of-Gzip-Deflate-compression-efficiency

https://github.com/h5bp/html5-boilerplate/blob/master/dist/index.html
https://getbootstrap.com/docs/4.1/getting-started/introduction/#starter-template

### DIFF
--- a/material/base.html
+++ b/material/base.html
@@ -2,7 +2,7 @@
 {% set feature = config.theme.feature %}
 {% set palette = config.theme.palette %}
 {% set font = config.theme.font %}
-<!DOCTYPE html>
+<!doctype html>
 <html lang="{{ lang.t('language') }}" class="no-js">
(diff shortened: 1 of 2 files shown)


squidfunk:
2 byte. Let's make the world a better place, huh? 🗡 

However, semantically it's the same, so I'm happy to merge. Will be in the next release when I find some time.

squidfunk:
Oh, you only did your changes in the `material` folder. They will be overwritten on the next build. Please adjust the `base.html` in the `src` folder and build the theme.

coliff:
ok - thanks, done!

PR #952, opened 2019-01-06
Add site_logo block for site logo customization

int3l:
Allow custom block for the logo, which will allow manual adjustments for that area.
Example use case - fixing broken logo images like this:
Wrong svg size

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -159,6 +159,7 @@ The Material theme provides the following template blocks:
 | `scripts`    | Wraps the JavaScript application logic          |
 | `source`     | Wraps the linked source files                   |
 | `search_box` | Wraps the search form in the header bar         |
+| `site_logo`  | Wraps the logo element in the header bar        |
 | `site_meta`  | Wraps the meta tags in the document head        |
 | `site_name`  | Wraps the site name in the header bar           |
 | `site_nav`   | Wraps the site navigation and table of contents |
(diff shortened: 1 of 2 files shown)


squidfunk:
Blocks in partials actually don't allow for overriding, see the latest discussion on the Gitter channel and [the issue](https://github.com/pallets/jinja/issues/243) on added/removed support within Jinja.

We would need to inline all partials that include blocks into the `base.html` to allow for that, but this would require some serious restructuring and is questionable in terms of maintainability. For now, I would recommend overriding the templates, not the blocks.

int3l:
Oh well... Thanks for the info.
But then why we don't update the documentation to reflect that?
I made the PR based on the info from the docs :)

Also, should I close the PR, or you still think that it's good idea to have the site_logo block?
(I guess that it will be available, when/if you do the inlining)

squidfunk:
I checked the partials: those that contain blocks and would need to be inlined are:

- `header.html`
- `footer.html`
- `nav.html` (transitive)
- `source.html`

I will need to think about this a little, but I think it should be worth. I will leave this PR open until I decided on how to proceed with this.

squidfunk:
Thinking about it a little more: this would require a new major release because of the template structure changing in a backward-incompatible way with the potential for breaking a lot of projects. Partials that were there are now inlined, so they cannot be overridden on a per-file basis.

squidfunk:
Removed blocks from partials in a27372e.

squidfunk:
@sander76 see the comments above, overriding never worked for blocks in partials. However, you can target the site logo with CSS and set the width and height accordingly.

sander76:
Thanks. I'll do the css thingy then

PR #999, opened 2019-02-20
remove unneeded type="text/css"

coliff:
This isn't needed in HTML5 documents and removing keeps the page a bit lighter.

As recommended by Google HTML Styleguide: https://google.github.io/styleguide/htmlcssguide.html#type_Attributes

### DIFF
--- a/src/base.html
+++ b/src/base.html
@@ -111,12 +111,12 @@
     {% block styles %}
 
       <!-- Theme-related stylesheets -->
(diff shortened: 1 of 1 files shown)


coliff:
I've since realised that the generated HTML doesn't include `type="text/css"` on the stylesheets so this PR may not be of any use...

squidfunk:
I would regard it as a possible optimization for the minifier, which it catches, and would keep it for clarity in the code base.

PR #1042, opened 2019-03-21
Add Python package metadata in package.json

raimon49:
Added Classifiers metadata as [shown in the left column of mkdocs package](https://pypi.org/project/mkdocs/).

I felt that you value the convention of declaring and using metadata in `package.json`. So I did the same.

The specified classifiers were selected from the [list_classifiers](https://pypi.org/pypi?%3Aaction=list_classifiers). You can add other classifications as well.

This theme is a great idea and design. Thanks for your maintenance work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,14 @@
     "documentation",
     "theme"
   ],
+  "classifiers": [
+      "Development Status :: 5 - Production/Stable",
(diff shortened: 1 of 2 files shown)


squidfunk:
Awesome, thanks for contributing!

Issue #1051, opened 2019-03-31
Can't build from Dockerfile


Anton-Latukha:
## Description

Current Dockerfile seems not working.

### Expected behavior

Possibility to build from Dockerfile the image locally.

### Actual behavior

Exits with exit code 2 on `python setup.py install`

### Steps to reproduce the bug

1. `git clone https://github.com/squidfunk/mkdocs-material.git`
2. cd mkdocs-material
3. `docker build -t mkdocs-material .`

Anton-Latukha:
Resolved #1052

PR #1052, opened 2019-03-31
Dockerfile: add README.md because setup.py requires

Anton-Latukha:
Simple enough, `python setup.py install 2>/dev/null` was silenced.
On debugging seen that it seeks for `README.md`.

`setup.py` now uses `README.md`:
https://github.com/squidfunk/mkdocs-material/blob/62ca75e27d7d2f0aa7071eef2f684c0a64fd2862/setup.py#L35-L38

Closes: https://github.com/squidfunk/mkdocs-material/issues/1051

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,7 @@ COPY MANIFEST.in MANIFEST.in
 COPY package.json package.json
 COPY requirements.txt requirements.txt
 COPY setup.py setup.py
+COPY README.md README.md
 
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for your PR! That's indeed an error that would have occurred on the next automatic Docker build.

squidfunk:
4abd18f also unsilences `setup.py`.

PR #1081, opened 2019-04-19
Changing the DOCTYPE case to support Mac build

sebastienlevert:
We've been experiencing some weird behavior using Azure DevOps and a Mac Agent where if the doctype was lower case, the generated doctype was : 

`<!doctype doctype html>` instead of `<!DOCTYPE html>`.

This fixes it.

### DIFF
--- a/src/base.html
+++ b/src/base.html
@@ -27,7 +27,7 @@
 {% set palette = config.theme.palette %}
 {% set font = config.theme.font %}
 
-<!doctype html>
+<!DOCTYPE html>
 <html lang="{{ lang.t('language') }}" class="no-js">
   <head>
 


squidfunk:
Could you please elaborate on why this is a problem? We changed the doctype to lowercase in #926 and it shouldn't be a problem, as [the spec](https://html.spec.whatwg.org/multipage/syntax.html#the-doctype) says:

> A DOCTYPE must consist of the following components, in this order:
> - A string that is an ASCII __case-insensitive__ match for the string "<!DOCTYPE". One or more ASCII whitespace.
> - A string that is an ASCII __case-insensitive__ match for the string "html".
> - Optionally, a DOCTYPE legacy string.
> - Zero or more ASCII whitespace.
> - A U+003E GREATER-THAN SIGN character (>).

sebastienlevert:
Actually, after review, it might be an issue isolated to Azure DevOps (our build engine). We are seeing incorrect output of the doctypes in the files being built :

`<!DOCTYPE doctype html>` instead of the expected `<!doctype html>`. In the end, using this PR might not even fix it, so I'll close it. Still really weird and we currently have to add a `<!DOCTYPE html>` to our main.html to "override" the one being generated. 

One of the reason we are suspecting is Pything and as we don't have the control over the Python version (stucked at 3.7.2) right now, we can't confirm it's a mkdocs, mkdocs-material or python issue. As soon as we have access to 3.7.3 we'll test and see if it was maybe just an issue of some weird HTML parsing happening in the back end.

We can close it for now. Thanks!

PR #1100, opened 2019-05-12
Escape quotes in TOC title attributes

comp500:
When there are quotes in a heading, invalid HTML is produced. The item title therefore needs to be escaped in the template.

### DIFF
--- a/src/partials/toc-item.html
+++ b/src/partials/toc-item.html
@@ -22,7 +22,7 @@
 
 <!-- Table of contents item -->
 <li class="md-nav__item">
-  <a href="{{ toc_item.url }}" title="{{ toc_item.title }}"
+  <a href="{{ toc_item.url }}" title="{{ toc_item.title | e }}"
       class="md-nav__link">
     {{ toc_item.title }}
   </a>


squidfunk:
I tested your PR but there are some problems with it. I create a headline with the following characters:

```
## Installation & < > " '
```

Without your change, it is rendered as such:

``` html

<a href="#installation" title="Installation &amp; &lt; &gt; " '" class="md-nav__link">
  Installation &amp; &lt; &gt; " '
</a>
```
So the `&`, `<` and `>` characters are escaped, but the quotes are not. If I apply your fix, the source will get double-escaped:

``` html
<a href="#installation" title="Installation &amp;amp; &amp;lt; &amp;gt; &#34; &#39;" class="md-nav__link">
  Installation &amp; &lt; &gt; " '
</a>
```

Frankly, this sounds like an issue with MkDocs or Jinja to me. Quotes should be escaped, or they may break HTML. Skimming [through the Jinja docs](http://jinja.pocoo.org/docs/2.10/templates/#html-escaping), I found no other function for escaping except `e` and `safe`, both of which don't seem to escape quotes.

comp500:
You are correct, this seems to be an issue with MkDocs (and potentially Python-Markdown). To generate the TOC, MkDocs uses a Python-Markdown extension which assumes that quotes are valid in TOC headings. MkDocs parses the generated TOC and passes it to Jinja, which does no further escaping (I believe automatic escaping is disabled in Jinja). I'll make an issue on the MkDocs repo shortly.

comp500:
Would a better solution be to just remove the title attribute? The title variable is not intended to be used within attributes, so this technically isn't a problem for Python-Markdown or MkDocs.

If there isn't an absolute need for the title variable to be used within attributes, code to escape the variable (e.g. a custom template filter) would unnecessarily increase the maintenance load for MkDocs.

squidfunk:
Thanks for your efforts on discussing this issue over at MkDocs. As the title shows the exact same content as the inner HTML, we should just remove it. Could you check the templates for other uses that might be removed and open a new PR? Happy to merge.

comp500:
Will do, I've also found some other areas that aren't correctly escaped as well, like the site title.

PR #1103, opened 2019-05-15
font-display: swap

xen:
Support for `font-display: swap;` directive to speed up page rendering while fonts are loading. More info https://font-display.glitch.me/


squidfunk:
You made changes in the `material` folder which gets overwritten when the theme is compiled. I integrated your change in 94d6f2c, which will be part of the next release.

xen:
Sorry, it was lame. Thank you anyway.

squidfunk:
Released as part of 4.3.0.

PR #1104, opened 2019-05-17
Display page.meta.title in nav bar (if given)

duncanmmacleod:
This PR patches `partials/header.html` to display the `page.meta.title` given in page metadata (using the `meta` extension), if given, otherwise uses the normal `page.title`.

### DIFF
--- a/material/partials/header.html
+++ b/material/partials/header.html
@@ -22,7 +22,11 @@
               {{ config.site_name }}
             </span>
             <span class="md-header-nav__topic">
-              {{ page.title }}
+              {% if page and page.meta and page.meta.title %}
+                {{ page.meta.title }}
+              {% else %}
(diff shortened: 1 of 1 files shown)


squidfunk:
You made changes in the material folder which gets overwritten when the theme is compiled. I integrated your change in 76eebf5, which will be part of the next release.

duncanmmacleod:
Ok, thanks for considering this, apologies for the confusion.

squidfunk:
I just pushed a change that minifies all files in the `material` folder and add a "do not edit" banner, so I hope there's less confusion in the future ;-) This also leads to a saving of 4kb (0.7kb gzipped) per page request, as there's no unnecessary whitespace anymore.

squidfunk:
Released as part of 4.3.0.

PR #1107, opened 2019-05-21
Fixed spelling error in danish translation

jaras:
It should be indtast not indtask.


squidfunk:
Released as part of 4.3.1.

PR #1117, opened 2019-05-28
Fix incorrect escaping of site_name, remove some title attributes

comp500:
This ensures that config.site_name is escaped correctly in all places, while also removing the title attributes discussed in #1100.

### DIFF
--- a/src/base.html
+++ b/src/base.html
@@ -306,7 +306,7 @@
                   as the main headline.
                 -->
                 {%- if not "\x3ch1" in page.content -%}
-                  <h1>{{ page.title | default(e(config.site_name), true)}}</h1>
+                  <h1>{{ page.title | default(config.site_name | e, true)}}</h1>
                 {%- endif -%}
 
                 <!-- Content -->
--- a/src/base.html
+++ b/src/base.html
@@ -42,16 +42,16 @@
       <!-- Page description -->
       {%- if page and page.meta and page.meta.description -%}
         <meta name="description"
-            content="{{ page.meta.description }}" />
+            content="{{ page.meta.description | e }}" />
       {%- elif config.site_description -%}
-        <meta name="description" content="{{ config.site_description }}" />
+        <meta name="description" content="{{ config.site_description | e }}" />
       {%- endif -%}
 
       <!-- Redirect -->
       {%- if page and page.meta and page.meta.redirect -%}
         <script>
           var anchor = window.location.hash.substr(1)
-          location.href = '{{ page.meta.redirect }}' +
+          location.href = {{ page.meta.redirect | tojson | safe }} +
             (anchor ? '#' + anchor : '')
         </script>
 
@@ -67,9 +67,9 @@
 
       <!-- Page author -->
       {%- if page and page.meta and page.meta.author -%}
-        <meta name="author" content="{{ page.meta.author | first }}" />
+        <meta name="author" content="{{ page.meta.author | e }}" />
       {%- elif config.site_author -%}
-        <meta name="author" content="{{ config.site_author }}" />
+        <meta name="author" content="{{ config.site_author | e }}" />
       {%- endif -%}
 
       <!-- Localization -->
--- a/src/base.html
+++ b/src/base.html
@@ -99,11 +99,11 @@
     <!-- Site title -->
     {%- block htmltitle -%}
       {%- if page and page.meta and page.meta.title -%}
-        <title>{{ page.meta.title }}</title>
+        <title>{{ page.meta.title | e }}</title>
       {%- elif page and page.title and not page.is_homepage -%}
-        <title>{{ page.title }} - {{ config.site_name }}</title>
+        <title>{{ page.title }} - {{ config.site_name | e }}</title>
       {%- else -%}
-        <title>{{ config.site_name }}</title>
+        <title>{{ config.site_name | e }}</title>
       {%- endif -%}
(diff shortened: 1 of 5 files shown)


src/partials/header.html
```diff
@@ -30,7 +30,7 @@
       <!-- Link to home -->
       <div class="md-flex__cell md-flex__cell--shrink">
         <a href="{{ config.site_url | default(nav.homepage.url, true) | url }}"
-            title="{{ config.site_name }}"
+            title="{{ config.site_name | e }}"
```
squidfunk: Doesn't this lead to the same problem, i.e. if the site name contains special characters like & and " that they get (double) escaped? Have you tested this?

comp500: I've tested this, and site_name isn't escaped at all, it's just copied from the config file directly. This appears to be the same for all config variables.
I also found that the copyright config variable isn't escaped, but I left it as you may want to enter HTML there - it isn't escaped in any of the default themes, so I assume it is supposed to be HTML.

squidfunk: Ok, good. Is `config.site_name` escaped in the any of the default themes? Are any variables escaped in default themes? Are there maybe other variables to consider escaping? Just want to "catch 'em all".

comp500: I had another look through all of mkdocs-material, and found a couple more areas where escaping is needed. I'll go through the default themes and see if they have any issues as well.

src/base.html
```diff
@@ -42,16 +42,16 @@
...
-        <meta name="description" content="{{ config.site_description }}" />
+        <meta name="description" content="{{ config.site_description | e }}" />
       {%- endif -%}
 
       <!-- Redirect -->
       {%- if page and page.meta and page.meta.redirect -%}
         <script>
           var anchor = window.location.hash.substr(1)
-          location.href = '{{ page.meta.redirect }}' +
+          location.href = {{ page.meta.redirect | tojson | safe }} +
```
squidfunk: What exactly is happening here? I've been testing your changes but couldn't come up with a case to test drive this.

src/base.html
```diff
@@ -306,7 +306,7 @@
                   as the main headline.
                 -->
                 {%- if not "\x3ch1" in page.content -%}
-                  <h1>{{ page.title | default(config.site_name, true)}}</h1>
+                  <h1>{{ page.title | default(e(config.site_name), true)}}</h1>
```
squidfunk: Doesn't work for me, I tried and think it should be `{{ page.title | default(config.site_name | e, true) }}`

squidfunk:
Regarding the removal of the `title` attribute we should be fine. First, they mirror exactly what is contained in the `a` tag. Second, screen readers don't even seem to read them, see:

- https://silktide.com/blog/2013/i-thought-title-text-improved-accessibility-i-was-wrong
- https://www.mediacurrent.com/blog/dont-rely-title-attribute-accessibility-seo/

src/base.html
comp500: I was going through the mkdocs website and saw the `tojson` filter: https://www.mkdocs.org/user-guide/custom-themes/#tojson and decided to use it. It probably isn't much of an issue, as you wouldn't want to use single quotes in a URL anyway, but it might fix some edge cases.

src/base.html
comp500: You are correct, I forgot to test that.

squidfunk:
I'm not sure how to proceed here. I did some research and the default themes don't escape the variables at all, see for example:

https://github.com/mkdocs/mkdocs/blob/ddf84aefd5f43bca53228c49d503707457175018/mkdocs/themes/mkdocs/base.html#L4-L17

Of course, escaping is necessary in attributes because of quotes and in text nodes because of `&` characters. However, I looked into your PR again and saw that you escaped __some__ references of `config.site_name`, mostly within attributes but also some within text nodes, but there are still unescaped references. Is there a reason for this?

Furthermore, I would want a opinion of the MkDocs team and a clear strategy before proceeding with this PR. We need to tackle those issues systematically. Do you want to take the lead on this and open an issue on the MkDocs tracker for discussion?

comp500:
Apologies for the delay. I think I didn't escape the references that were in `<title>` tags, because these tags cannot have subtags so they don't need escaping for most special characters (e.g. `<title><a></title>` works). I can see how this could be an issue if `</title>` is in the title text, so I'm going to push another commit to escape those.

I'll open an issue on the MkDocs tracker shortly as you have suggested, to discuss these issues with the MkDocs team.

squidfunk:
I followed the discussion on the MkDocs issue tracker and came to the following conclusion:

I think escaping of configuration attributes should be left to the user. If we start escaping template variables, we must escape all variables to be consistent. However, if we do, it won't be possible to add HTML to configuration options like some people do with the Metadata extension, e.g. for [custom heros](https://hexagonkt.com/) (which I didn't know was possible before I've seen it). I think this is pretty cool and don't want to take this away from the user.

Users often ask for more configuration options, but better than providing options is giving them degrees of freedom like for example adding your own hero HTML. This PR would reduce this kind of freedom.

Nevertheless, we should definitely add a block on escaping to the _Getting Started_ guide. Happy to merge PRs.

squidfunk:
Closing for the reasons stated. I want to thank you again for taking the time to investigate this issue and facilitating the discussion with the guys over at MkDocs, but I think it doesn't make any sense progressing further down this way.

PR #1119, opened 2019-05-29
Update zh.html

zaxayn:
I am Chinese, which is more in line with Chinese habits. If you have Chinese friends, you can check with him：)


squidfunk:
Thanks for your PR! I checked with Google Translate and it returns "previous page" and "Next page" as opposed to "back" and "go ahead". LGTM. Will be part of the next release.

PR #1120, opened 2019-05-30
Add GNU Make information

k4kfh:
If you're trying to build mkdocs-material on Windows, it seems GNU make is required. It's not hard to install it, but it's worth linking in the docs.


squidfunk:
Thanks!

Issue #1133, opened 2019-06-13
Consider using font-display: fallback for Google Fonts


coliff:
It's great that Google Fonts supports the `display` parameter. I found that setting `display=swap` always displays the fallback font immediately and switches - even if the page and Google font loads **_very_** fast which can be a bit annoying.

I suggest to instead use `display=fallback`. This will only show the fallback font after about 100ms which means the user doesn't see the font switching, but will never need to wait more than 100ms until they can start reading the content.

I've done quite a bit of testing on this on several different sites and prefer but this - but it's just my personal opinion though - so feel free to close this, but give it a try and see what you think. If you set  `display=fallback` you'll still get a pass for it on Google Lighthouse which is good.

REF: https://font-display.glitch.me/

squidfunk:
@xen is the one who PR'd `font-display: swap` in https://github.com/squidfunk/mkdocs-material/pull/1103 - any thoughts?

xen:
I think it is pretty reasonable arguments. My first intention was to make rendering better on long pages. But I didn't dig so deep. If this trick works on the mobile then it probably better to accept.

squidfunk:
Thanks! @coliff could you PR this? Happy to merge.

coliff:
Hi - I made a PR by changing the file in `src`.. I can't make a build of the site so sorry for not following your contribution guidelines. :-\

PR #1134, opened 2019-06-14
improves google font display experience

coliff:
This fixes: #1133
This change prevents the font changing if the Google font loads within 100ms.


squidfunk:
That was quick, thanks. Need to rebuild the theme though. Will do that when I issue the next release. Maybe I find some time on the weekend.

PR #1217, opened 2019-08-22
Fix issue where html elements in H1's cause footer layout to be broken

timpharo:
## Expected Behaviour
When adding a html tags to H1's in markdown files, such as Font awesome icons, the footers next/previous links should look as expected

## Current Behaviour
Adding html to H1's in markdown files displays the html correctly in the link itself, but as the H1 is injected into the title attribute of the link, which is invalid as this should not contain html, the layout breaks

## Possible Solution
Stripping all html when reading the property in as a Jinja template.  This PR addresses the issue in this way.

## Steps to Reproduce
1. Add `fontawesome_markdown` extension and `awesome-pages` plugin
2. Add Font Awseome icon to any .md file
3. Navigate to an adjacent page to the one you have edited 
4. The footer layout will be broken because the html is injected into the links title attribute


squidfunk:
LGTM!

squidfunk:
Released as part of 4.4.1.

timpharo:
Thanks for the quick turn around, now I can remove the partial from our code.

Issue #1214, opened 2019-08-21
Feature Request: Allow 'black' to be selected as primary color for theme


mightymackem:
## Description

I would like to switch my theme to use black as the primary color. Black is included in the [Material Design color palette](https://material.io/design/color/#tools-for-picking-colors) but can't be selected as an option under `palette: primary` in the `mkdocs.yml' file. 

Please consider updating the theme so that I can select "black" as the main color for my theme.

Thanks for your consideration!

squidfunk:
Released as part of 4.4.1.

mightymackem:
Just wanted to say how appreciative I am that you added this enhancement so quickly, thank you very much!

PR #1218, opened 2019-08-22
Address Issue #1214: Allow primary colour black

timpharo:


squidfunk:
Unfortunately, this will also render all links in black color, which makes them indistinguishable from body copy. This needs to be addressed in a similar manner as the `white` color which is a special case inside the `application-palette` file.

squidfunk:
Took the time and fixed it in 6b7535b

mightymackem:
If you select `black` as the primary color and want to use `light blue` as the color for links, is that possible? If so, what needs to change in the `application-palette.scss` to make this happen?

squidfunk:
We can only set one link color per primary color, so no. I chose indigo as the link color as 1) it’s the default of this theme and 2) its similar to the default link color of all browsers. You could however set the link color yourself with one line of CSS. See the customization guide for guidance.

Issue #1207, opened 2019-08-16
Suggestion: Remove unneeded title attribute from sidebar links


coliff:
## Description

Currently the links in the sidebar have `title` attribute so hovering over a link shows a tooltip with the same text as the link. It's a bit distracting and unnecessary I think. Could these be removed?

Example: 

![image](https://user-images.githubusercontent.com/1212885/63175728-9124d300-c07f-11e9-93d0-f14deaecbd81.png)

### Expected behavior

would prefer `title` to not be on text links in sidebar

### System information

* OS: Windows 10 
* Browser: Microsoft Edge (Chromium)

squidfunk:
I put them there because the W3C validator complained that it was missing, but I guess from todays perspective this would be considered an antipattern. I think we already did some research somewhere on this topic, but can't find it.

coliff:
Yep, I remember that you previously merged a PR I made for removing unneeded Title from links in the footer. #585.

I've quite certain that title tags are not needed on text links here. The only possible reason that it could be needed that I can think of would be if the text was shortened or truncated so if you hover over the title tag would show the full title.

squidfunk:
Ah yes, I remember. If you like, you can sift through the source and remove all unnecessary titles in a PR. Happy to merge.

coliff:
cool. would happy to do that... probably Monday morning but maybe over the weekend. I had trouble making a build before though -is it okay to just make changes in `src` ?

squidfunk:
Would be interesting where you had trouble but yeah, basically it’s fine because I will build it anyway before the next release.

PR #1224, opened 2019-08-23
Remove unneeded title attribute from text links in nav and sidemenu

coliff:
Fixes: #1207

This PR removes the unneeded title attribute from text links in nav and side-menu. There are a few other instances of title attributes being used, but I left them for the time being. The ones in the footer aren't so annoying and hovering over the 'next page' `>` icon and seeing page title could be helpful to some.


squidfunk:
> The ones in the footer aren't so annoying and hovering over the 'next page' > icon and seeing page title could be helpful to some.

On mobile, previous is renderer without the page name, so it's definitely worth keeping.

squidfunk:
Thanks!

PR #1283, opened 2019-10-02
Missing repo_url results in "/", not a falsy value

DiscoStarslayer:
Ran into a bit of a config edge case with mkdocs-material.

Looking at the readme, it seems like we should be able to omit `repo_url` from the `mkdocs.yml` to hide the link to a source repository. This does work, but only partially, the `div` that contains this link is not removed, and still takes up space in the navigation bar. See below image for an example, there is a lot of white space to the right of the search bar:

before

After a bit of investigating, I found that simply omitting the `repo_url` from the `mkdocs.yml` config results in `repo_url` being `"/"`, not `None` or `""`.

This seems strange, and I'll admit my Python-fu isn't the strongest, so maybe there is a way to explicitly set `repo_url` to a falsy value instead of `"/"`.

This PR simply modifies the templates to check for `"/"` instead of a falsy value to hide this section from the page.

This does fix the issue, see below image:
after

I am definitely open to suggestions to my approach! Fixing the config to actually set `repo_url` to a falsy value feels like the proper approach for this, but I can't figure out where to do that. Maybe in mkdocs upstream?

### DIFF
--- a/material/partials/header.html
+++ b/material/partials/header.html
@@ -40,7 +40,7 @@
           {% include "partials/search.html" %}
         {% endif %}
       </div>
-      {% if config.repo_url %}
+      {% if config.repo_url != "/" %}
         <div class="md-flex__cell md-flex__cell--shrink">
           <div class="md-header-nav__source">
             {% include "partials/source.html" %}
--- a/material/partials/nav.html
+++ b/material/partials/nav.html
@@ -12,7 +12,7 @@
     </a>
     {{ config.site_name }}
   </label>
-  {% if config.repo_url %}
+  {% if config.repo_url != "/" %}
     <div class="md-nav__source">
       {% include "partials/source.html" %}
     </div>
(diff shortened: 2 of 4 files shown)


squidfunk:
Could you share your configuration? I'm not able to reproduce the problem you're trying to solve with a minimal configuration:

``` yml
site_name: Test
theme:
  name: material
```

Bildschirmfoto 2019-10-01 um 17 31 24

DiscoStarslayer:
@squidfunk the full configuration is here: https://github.com/xqemu/xqemu.com/blob/master/mkdocs.yml

You not being able to re-create it was suspicious though, I can confirm the theme works as expected with a blank config.

Fuzzing our config a bit, it looks like having the property `edit_url` sets the `repo_url` property to `"/"`.

Looking at mkdocs upstream, looks like this is the line of code causing us the issue: https://github.com/mkdocs/mkdocs/blob/efdb5cf7f0d80e7e0d315719e9605ef6e9c23f04/mkdocs/config/config_options.py#L312

Apologies for bugging you, but thank you for testing, it helped a lot to point me in the right direction!

Issue #1266, opened 2019-09-23
Color schemes: choose scheme based on user preference


MatushkaRossiya:
## Description

Add support for [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) CSS feature.

### Use Cases

MkDocs pages with Material theme will automatically switch between light and dark look based on user's OS preference (Windows 10, macOS 10.14+, Android 10+, iOS 13+, etc), matching system UI, web browser and other apps.

squidfunk:
Thanks for your suggestion.

Material (currently) doesn’t have a dark mode, so this would be the first thing that needs to be implemented. This could be done I a way that it automatically supports the level 5  ‘prefers-color-scheme` and even `light-level` media queries.

However, this would need to be done orthogonally to the current possibility of color customization, so both things can be combined.

I’m happy to discuss concepts and ideas here and help to craft a PR, as I’m not able to do it myself at the moment.

Ir1d:
btw: In [mdui](https://www.mdui.org/docs/#mdui-dialog), when selecting dark-mode, color of the nav-bar is overwritten to black. (it seems that the color customization can be overwriten when set to dark mode. but it is somehow beyond my ability to implement a front-end option to change themes :|

squidfunk:
Partially solved by #1305

henrywhitaker3:
I have added a [second stylesheet](https://github.com/henrywhitaker3/mkdocs-material-dark-theme/blob/master/docs/assets/css/dark-theme-system.css) to my dark theme [repo](https://github.com/henrywhitaker3/mkdocs-material-dark-them) which is just the main dark theme CSS wrapped in the `@media(prefers-color-shceme: dark)` media query. I've only been able to test it on Android 10, but it works on that.

squidfunk:
@henrywhitaker3 probably not necessary to provide a separate stylesheet, as it could be done via `<link ... media="(prefers-color-scheme: dark)" />` which could be configured via a flag in `mkdocs.yml`

henrywhitaker3:
Oh yeah for sure, just didn't/don't have the time to work out how to do that haha

Stanzilla:
I implemented it like this:

```js
!function() {
	const darkMode = document.getElementById('dark-mode-toggle');

	if (darkMode) {
		const isDarkSchemePreferred = () => window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches;

[...]

Issue #73, opened 2016-12-19
RFE: customize colors in mkdocs.yml


kcgthb:
Hi Martin, 

First of all, thanks a lot for your awesome work on this theme!

Do you think it would be possible to allow customizing the color palette in `mkdocs.yml` directly? I'd like to use colors that are not in the [Material Design color palette](http://www.materialui.co/colors), and currently, my only option is to modify `src/assets/stylesheets/_palette.scss` to add custom colors, and rebuild the whole theme. Which seems a bit heavy for just changing colors.
Plus, it makes tracking changes and fixes in the theme more difficult, as it requires a rebuild to add those custom colors each time the theme is updated.

Having an option to specify hex codes for primary and accent colors in `mkdocs.yml` would be a much easier way to customize colors outside of the material palette. 

What do you think?

squidfunk:
It's not on the roadmap. We would have to put all rules that use the colors into the main template (or a partial), so it would always be rendered inline. This is an ugly mixture of markup and styles which is already necessary for custom webfonts, but far more rules are necessary for the color and it would worsen maintainability.

The Material theme supports 19 colors. Show me a another mkdocs theme that does that ;-) Also compiling is fairly easy and you can adjust everything by just changing two variables.

kcgthb:
Thanks for the explanation. Compiling is indeed not complicated and the theme supports a wide variety of colors already, but sometimes, when integrating the theme in an ecosystem of websites that all need to adhere to the same graphical rules, you need very specific colors that are not in the theme.

But that's perfectly fine, being able to customize the theme is already very convenient.

Thanks!

squidfunk:
I will provide an example which classes to override with extra CSS, rather than compiling the project. Basically all classes are listed in `src/assets/stylesheets/application.palette.css`, so you just need to copy those and put them in a separate CSS file which needs to be included.

lgeiger:
> I will provide an example which classes to override with extra CSS, rather than compiling the project

Do you have a link to this example? I think it would be very helpful for users who want to customize the appearance.

squidfunk:
There are already a lot of examples for a multitude of different use cases in the issues.

Issue #1292, opened 2019-10-07
Table of Contents is blurry


tuner:
__I checked that...__

- [x] ... the documentation does not mention anything about my problem
- [x] ... the problem doesn't occur with the default MkDocs template
- [x] ... the problem is not in any of my customizations (CSS, JS, template)
- [x] ... there are no open or closed issues that are related to my problem

## Description

Table of Contents (the right sidebar) is slightly blurry on Chrome and Safari on Mac. Firefox is ok. I did not test on Windows.

![blurry](https://user-images.githubusercontent.com/399972/66311956-5a00ce80-e918-11e9-95df-28bb980f1fe6.png)

It's likely related to the transform rule in css. Disabling it makes the text crisp again (with TOC at wrong place, though)

```
@media only screen and (min-width: 60em)
.md-sidebar--secondary {
  transform: translate(-100%);
}
```

### Steps to reproduce the bug

1. Navigate to: https://squidfunk.github.io/mkdocs-material/
2. And resize the windows so that it is at least 1600px wide.

squidfunk:
Thanks for reporting. You're right, it is related to the `transform` rule which induces a new layer on the GPU. The text is blurry, because `width` and `height` are not divisible by 2 (see [this link](https://keithclark.co.uk/articles/gpu-text-rendering-in-webkit/)) which cannot be guaranteed with pure CSS.

It might however be solvable by using `calc()` with `margin`, effectively making the `transform` unnecessary. Things to do:

1. Try to replicate the TOC behavior with `calc()`
2. If this works, research browser support on `calc()`

Support should be good by now, so if it works we can make an informed decision and switch to a `calc()`-based position strategy. Maybe there is an even simpler approach.

tuner:
Yup, I have been using calc for years.

I did some initial experiments with Chrome's dom inspector and looked at the sass sources.

In _sidebars.scss:

```scss
width: px2rem(242px);

@include break-from-device(tablet landscape) {
      display: block;
      margin-left: 100%;
      transform: translate(-100%, 0);
      ...
}

@include break-from-device(screen) {
      margin-left: px2rem(1220px);
      ...
}
```

Couldn't you just have something like:

```scss
$width: px2rem(242px);
width: $width;

@include break-from-device(tablet landscape) {
      display: block;
      margin-left: calc(100% - $width);
      ...
}

@include break-from-device(screen) {
      margin-left: px2rem(1220px) - $width;
      ...
}
```

[...]

PR #1293, opened 2019-10-08
Feature/custom color values

diba1013:
This PR should address some issues for custom color values as primary and accent color. Secondly, this also provides a step towards dynamic themes (i.e. dark and light theme) as discussed in #1266.

#### Disclaimer

I know that this is customizable with the Google Material Design Palette, or by compiling your own theme while changing the values in `_config.scss`. However, I think, this is a great effort and needs constant updates to stay up to date with your work (contrary to your [comment](https://github.com/squidfunk/mkdocs-material/issues/73#issuecomment-268190426)).

Secondly, please feel free to improve, remove or discard my changes, as you see fit. I will adjust any comments from your side. If so, I will happily fall back to the existing solutions.

### State of the PR (see commits)

This PR allows the customization for the theme by not just allowing a fixed number of theme colors, but also an arbitrary hex string as color code. This is done with two steps:

1. Use variables to change address the color codes such that they can be changed with one line and at runtime.
    1. Introduce mixins for easy variable generation
    2. Define the base theme which is overwritten by more specialized (i.e color palette themes)
2. Use JavaScript to read, parse the color code and update the variable. _The real magic happens here_.

I also provided a draft to update to the documentation.

#### Things to consider

Since the color code is updated with JavaScript&mdash;which takes a few milliseconds to load&mdash;there is a transition from the default theme to the color code. This could be addressed by following proposals:

1. Use a separate variable, so that one can choose a close color palette and the change is not that drastic.
2. Disable the color transition on initial load, which might be faster initially but potentially more disturbing.

_You might need to format the code with your formatter._

#### Compatibility

This PR uses CSS variables and `var()`, which is [supported](https://developer.mozilla.org/en-US/docs/Web/CSS/var) by almost every browser (except IE).

_However, if you strive for full support (even really old versions), please feel free to dismiss this PR._

#### Future Work

1. I did not provide the updated theme as I am not sure if these are platform dependent.
2. The JavaScript can/has to be improved, as I did not want to spent too much time digging through your code. Thus this is implemented rather rudimentary.
    * Add change listeners to update the values automatically. This would update the theme every time the `data-md-color-primary` and `data-md-color-accent` changes.
    * Derive the rest of the colors from the base color to update the theme. This is still supported through the palette colors, but not possible in CSS as `mix` and `transparentize` or similar Sass-Functions [do not work](https://stackoverflow.com/q/55823860) with variables at the moment.
3. It might be worth to have a input field within the documentation, similar to your palette buttons to change the theme dynamically.

_If needed, I will create an issue to track further work._

#### Build

This PR was built and tested on both Windows and WSL (Debian) with [recheck-web](https://github.com/retest/recheck-web) before and after the build with no changes&mdash;testing major browsers and all specified [resolutions](https://github.com/squidfunk/mkdocs-material/blob/master/src/assets/stylesheets/_config.scss#L38).

_Disclaimer: I am one of the developers of recheck-web._

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -165,6 +165,17 @@ Design color palette. Valid values are: `red`, `pink`, `purple`, `deep purple`,
 `yellow`, `amber`, `orange`, `deep orange`, `brown`, `grey`, `blue grey` and
 `white`. The last four colors can only be used as a primary color.
 
+You may also define custom hex values instead of the names to easily adapt the
+theme to your style. Since this is dynamic and cannot be statically embedded
+into CSS, these values are updated once the page is loaded.
+
+``` yaml
+theme:
+  palette:
+    primary: '#123456'
+    accent: '#789ABC'
+```
+
 If the color is set via this configuration, an additional CSS file that
 defines the color palette is automatically included. If you want to keep things
 lean, clone the repository and recompile the theme with your custom colors set.
--- a/src/assets/javascripts/application.js
+++ b/src/assets/javascripts/application.js
@@ -84,6 +84,20 @@ function initialize(config) { // eslint-disable-line func-style
       return !!navigator.userAgent.match(/(iPad|iPhone|iPod)/g)
     })
 
+    // TODO bind this to variable?
+    const colorPrimary = document.body.dataset.mdColorPrimary
+    if (colorPrimary.startsWith("#")) {
+      document.body.style.setProperty("--primary-color", colorPrimary)
+      // TODO how to calculate --primary-color-dark
+    }
+
+    // TODO bind this to variable?
+    const colorAccent = document.body.dataset.mdColorAccent
+    if (colorAccent.startsWith("#")) {
+      document.body.style.setProperty("--accent-color", colorAccent)
+      // TODO how to calculate --accent-color-hover
+    }
+
     /* Wrap all data tables for better overflow scrolling */
     const tables = document.querySelectorAll("table:not([class])")              // TODO: this is JSX, we should rename the file
     Array.prototype.forEach.call(tables, table => {
--- a/src/assets/stylesheets/application-palette.scss
+++ b/src/assets/stylesheets/application-palette.scss
@@ -42,6 +42,14 @@
 // ----------------------------------------------------------------------------
 
 // Color tile for presentation in theme documentation
+button[data-md-color-primary] {
+  background-color: var(--primary-color);
+}
+
+button[data-md-color-accent] {
+  background-color: var(--accent-color);
+}
+
 button[data-md-color-primary],
 button[data-md-color-accent] {
   width: px2rem(130px);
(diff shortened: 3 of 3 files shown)


squidfunk:
This is a great idea. I'm currently refactoring the three year old JavaScript code base and the CSS could be updated, too. A lot has changed in three years. Together with #1292 where we scrap `transform` for `margin` + `calc()` we could rethink the (S)CSS architecture a little. Browser support is, indeed, pretty decent.

However, I don't like the JavaScript solution. Overriding should be done via `mkdocs.yml` and/or CSS. There should also be a fallback for browsers that don't support CSS variables. Also, defining variables in a mixin really smelly like an anti-pattern. So, nice idea, but needs some work. I would leave this PR open and reconsider the ideas in it when I've finished revamping the JavaScript architecture (and migrating everything to TypeScript + RxJS). Material for MkDocs v5 should/could be based on CSS variables.

diba1013:
> However, I don't like the JavaScript solution

I, too, think that styling should be done in CSS, as it is pretty much just transferring a html attribute into CSS. At the current state of CSS&mdash;at least to my knowledge&mdash;there is nothing like variable capturing. Thus for dynamic content it is sadly unavoidable to use JavaScript or to dispense the more or less dynamic, but rather configurable approach. 

An alternative approach&mdash;if possible&mdash;would be to generate the stylesheets (or even just the variables) whenever MkDocs builds the website.

> Also, defining variables in a mixin really smells like an anti-pattern

I'm fairly new into web technologies. So whatever works out of the 1000 possibilities is fine by me.

> So, nice idea, but needs some work

Please feel free to alter and adapt any part as you see fit.

> We're talking rather weeks than days

Sure, take your time. In the meantime, I will fallback to the current approaches and customize your theme via CSS or just leave it as is, since the colors are pretty close to our corporate identity. Just need to keep an eye out for the new releases.

squidfunk:
> I, too, think that styling should be done in CSS (hence the name), as it is pretty much just transferring a html attribute into CSS. At the current state of CSS&mdash;at least to my knowledge&mdash;there is nothing like variable capturing. Thus for dynamic content it is sadly unavoidable to use JavaScript or to dispense the more or less dynamic, but rather configurable approach. 

The color could be set in `mkdocs.yml` and then transferred via the template to an inline CSS style where it is set, e.g.:

Bildschirmfoto 2019-10-08 um 19 21 12

Custom colors could be set from `mkdocs.yml` omitting the need to provide extra CSS. We would, however, need to think about how to provide the color palette. This could be done by a template macro (similar to how i18n is implemented) which is generated from the original Material color palette. The color is then picked from the macro and inlined into the template. If the color isn't found in the template, it is inlined as a literal. This enables the usage of colors from the Material palette (e.g. `red-900`) and custom colors (e.g. `#FF0000`) at the same time from `mkdocs.yml`.

I will prepare something for the next major release. Material 5 will be a big major release, so there will be a beta/RC stage which we'll use for testing.

diba1013:
That sounds like a good solution. Do you mind giving me a heads up once you release the beta, so that I can test it with our documentation?

Should we move this discussion into an issue and close this PR? I will keep my changes and the branch so that you may refer to it.

squidfunk:
Yes, feel free to create an issue with the ideas discussed.

diba1013:
I will close this, as this is now represented as an issue #1297. Feel free to check out the [original branch](https://github.com/diba1013/mkdocs-material/compare/master...diba1013:feature/custom-color-values).

PR #1298, opened 2019-10-09
reduce number of github requests

samuelcolvin:
Should significantly reduce the number of requests made to github to get forks and stars, by:
* increasing the number of items per page to 100 (the maximum allowed)
* ordering by update timestamp (descending since that's the default) - this is based on the assumption that projects with mkdocs sites are likely to be updated regularly and are therefore likely to be found sooner with this approach.

I found this since for https://pydantic-docs.helpmanual.io/ i noticed:
image

This is worthwhile since:
* it should reduce the delay until the stars/forks are shown
* it should use less bandwidth and battery to render the page
* nice to reduce the load on github's servers


samuelcolvin:
by the way, thank you for this theme. It's beautiful.

squidfunk:
Great idea!

PR #1330, opened 2019-11-06
Fix EmojiOne URL to prevent a HTTP redirect

coliff:
EmojiOne is now JoyPixels
https://blog.joypixels.com/emojione-is-now-joypixels/

### DIFF
--- a/docs/extensions/pymdown.md
+++ b/docs/extensions/pymdown.md
@@ -194,7 +194,7 @@ emojis. Happy scrolling :tada:
 
   [13]: https://facelessuser.github.io/pymdown-extensions/extensions/emoji/
   [14]: https://emoji.codes/
-  [15]: http://emojione.com
+  [15]: https://www.joypixels.com/
   [16]: https://creativecommons.org/licenses/by/4.0/legalcode
   [17]: http://emojione.com/licensing/
 


facelessuser:
Yeah, they've changed, and their license for their emoji images is now even less friendly than it was with version 3 release. Pymdown Extensions still uses their short name index as the short name index license is under MIT, but it still references the old CDN for emoji images with their version 3 emoji. Version 2 probably had their most user friendly license though.

I most likely will switch over to the Twemoji emoji index in my docs which uses the friendly name index from JoyPixels, but uses Twitter's emoji images. You get SVG through Twemoji as well which is pretty cool and they are pretty up to date. I liked the look of EmojiOne images a bit more, but Twemoji images aren't bad though.

squidfunk:
@facelessuser why is the license less friendly now? What has changed? Should we update the docs to recommend Twemoji?

facelessuser:
@squidfunk, you can check out their license stuff. I'm no lawyer or anything:

- https://www.joypixels.com/licenses
- https://www.joypixels.com/licenses/free

Maybe its fine, but it got real specific.

They are focusing harder on monetizing the work, maybe it is safe to still use, but I feel like you have to be more careful.

I wasn't sure, so while I still use the latest short name index, which can be freely used, I still link to the old v3 CDN.

Twemoji is just more friendly: https://github.com/twitter/twemoji#license.

Feel free to take a look and give me your opinion. I'd love to get a secondary opinion on JoyPixels current free tier.

facelessuser:
Small correction, I point to the 4.5 CDN which was the last under EmojiOne.

squidfunk:
> - Free license includes low-res PNG files and emoji fonts.
> - Businesses can use our free emoji for testing purposes only.

Not cool. Many companies (businesses) use Material, though I don't know how many use Emojis. Nevertheless, I would second your idea of switching to an alternative. Happy to merge contributions to the docs with further directions.

facelessuser:
Yup, that was my feeling as well. It is just too strict for my taste. I feel Twemoji fills the gap much better.

facelessuser:
I may make Twemoji pymdown-extensions the default moving forward. I'm not sure if any style change is required on Material side for targeting emoji classes, but I'll look into it.

squidfunk:
Perfect. I'm merging this PR now. Let's tackle the switch to Twemoji in another PR.

facelessuser:
FYI, I can't even find the license anymore for emojione 3 and 4. I wanted to compare the license again to 2.2.7, the last truly free version of EmojiOne. Unfortunately, it is impossible to find those licenses prior to the 5.0 release.

I've decided to walk back EmojiOne support to 2.2.7, which honestly has a ton of emoji support, and the only version I'm 100% sure is safe to use. EmojiOne will be locked to 2.2.7 indefinitely moving forward.

My personal projects will be using Twemoji moving forward which has the latest Unicode 12.0, but I wont force it as the default yet.

I will release a new version of PyMdown Extensions which will be using this safer version of EmojiOne, and then make a PR here showing people how to use Twemoji for the latest Unicode 12 emoji.

squidfunk:
@facelessuser perfect, thanks for your help!

PR #1334, opened 2019-11-07
Fixed: greek translation - switched next and previous order

BanForFun:
In the Greek language file, the translations for 'previous' and 'next' where in the wrong order.


squidfunk:
LGTM. Thanks!

PR #1338, opened 2019-11-16
Use Twemoji by default and document how to use Twemoji

facelessuser:
- Document how to use Twemoji and explain why it may be a better choice than the current default EmojiOne (now JoyPixels).
- Fix links links that are no longer relevant or unhelpful adding new , more helpful links.
- Have MkDocs Material use Twemoji in its documentation by default.

### DIFF
--- a/docs/extensions/pymdown.md
+++ b/docs/extensions/pymdown.md
@@ -183,69 +183,89 @@ all Admonition qualifiers can be used, e.g. `note`, `question`, `warning` etc.:
 ### Emoji
 
 [Emoji][13] adds the ability to insert a :shit:-load of emojis that we use in
-our daily lives. See the [EmojiOne demo][14] for a list of all available
-emojis. Happy scrolling :tada:
+our daily lives.
+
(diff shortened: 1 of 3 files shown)


facelessuser:
I did not rebuild the assets.

squidfunk:
Awesome! Thanks for your help on this front. I'll rebuild the theme and draft a new release shortly.

squidfunk:
Do we need to update `requirements.txt`?

https://github.com/squidfunk/mkdocs-material/blob/401dfd9e2097e03b02e5a07f4326c6b29e7619d2/requirements.txt#L25

facelessuser:
Ah, you could force it to use 6.2 which locks EmojiOne to 2.2.7.

facelessuser:
I was going to update the requirements in my next pull which will avoid 6.3 and the next Python Markdown do to code changes.

facelessuser:
I should have it ready shortly.

squidfunk:
Ah okay, then I'll wait for this one before releasing 4.5.

PR #1339, opened 2019-11-16
Update requirements

facelessuser:
- Restrict Markdown to less that 3.2 to prevent CSS breakage as 3.2
  will introduce changes to how code blocks are formatted.
- Use at least version 6.2 of Pymdown Extensions as 6.2 limits EmojiOne
  usage to v2.2.7 which has the most agreeable license for commercial
  and personal use. Additionally, limit the version to to less than 6.3
  as 6.3 will follow Python Markdown 3.2 in code block changes for
  SuperFences.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,4 +22,5 @@
 mkdocs>=1
 mkdocs-minify-plugin>=0.2
 Pygments>=2.2
-pymdown-extensions>=4.11
+markdown<3.2
+pymdown-extensions>=6.2,<6.3


facelessuser:
At some point in the future, once Python Markdown and PyMdown Extensions release their next major version, MkDocs Material should update the code styles to support the latest code format. Once that occurs, MkDocs should also update the minimum requirements for PyMdown Extensions.

I am uncertain if PyMdown Extensions will require Markdown 3.2 in the next major release or still allow lower versions, so if the later occurs, we'll need to update Python Markdown's minimum requirement as well. If we go the former route, you can remove the Python Markdown requirement altogether.

squidfunk:
Thanks again!

> At some point in the future, once Python Markdown and PyMdown Extensions release their next major version, MkDocs Material should update the code styles to support the latest code format. Once that occurs, MkDocs should also update the minimum requirements for PyMdown Extensions.

Could I ask you to open another PR once those conditions are met?

facelessuser:
Yeah, I was more providing the comment as documentation, but that was my intention.

squidfunk:
Great! 4.5 with your changes will be released in a few minutes.

PR #1356, opened 2019-11-29
update min MkDocs version in Getting Started

coliff:
MkDocs Material has required MkDocs v1.x since version 3 so I think it'd be right to update the minimum version here now.
MkDocs v1 was released more than a year ago and older versions are unsupported.
I also bumped the versions of Python and PIP just for good measure as those (although they are just example output) are really old versions too.

(minor nitpicky thing is that the first release of MkDocs v1 release on https://github.com/mkdocs/mkdocs/releases is actually 1.0, but I put it above as 1.0.0 for consistency... feel free to change :-D

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -10,19 +10,19 @@ good to go with the following commands:
 
 ``` sh
 python --version
-# Python 2.7.13
+# Python 3.8.0
 pip --version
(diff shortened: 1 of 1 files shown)


squidfunk:
Pip 9 -> 19? WTF just happened? Do they have Chrome's release policy now?

PR #1365, opened 2019-12-04
Compress teaser image.

XhmikosR:
Before: 1305331 bytes
After:  299544 bytes

This was a lossy compression but no super-important visual differences AFAICT (tested on 4K with 150% and 250% DPI). Could probably be improved with zopflipng later but read below :)

IMHO, this is not the proper fix, but at least it saves ~ 1MB. The proper fix would be to use responsive images and then compress those. But since:

1. I'm not aware if this image is only used in the homepage
2. I'm not sure what is the biggest breakpoint here
2. there's no way to use responsive images in Markdown so we'd need to use raw HTML

I didn't proceed with the responsive images approach.


squidfunk:
LGTM, thx!

XhmikosR:
Not bad :) https://gtmetrix.com/compare/mm9htmlQ/1htJKxM3

squidfunk:
Nice, thanks!

PR #1366, opened 2019-12-04
Update base.html

XhmikosR:
Escape `|` in the Google Fonts URL.


squidfunk:
Could you please add some more information why you made this change, what problem it should solve and what the potential impact is?

XhmikosR:
The character isn't allowed there. Just run the HTML validator yourself.

squidfunk:
> The character isn't allowed there. Just run the HTML validator yourself.

I really value contributions and a good explanation saves time and overhead. This is all I'm asking for.

PR #1375, opened 2019-12-10
Change generated file manifest name

### LINKED ISSUE #1374: manifest.json handling
Hey there, I don't quite understand providing a functioning manifest.json and serviceworker should work with mkdocs-material. It is recommended to put both sw.js and manifest.json in the root of the site, that's not possible when using a custom build of material theme because the material theme overwrites it with it's own manifest.json which is a list of files.

Using the paths from your examples here https://github.com/squidfunk/mkdocs-material/issues/810#issuecomment-438442460

My suggestion would be altering the webpack config to add the `fileName: "file_manifest.json"` argument to it so the file names no longer clash

Stanzilla:
In order to not clash with the web standard manifest.json, rename our file manifest used for cache busting to file_manifest.json and adjust theme accordingly.
Fixes #1374


squidfunk:
Thanks, however, there are too many unnecessary files in this commit. Please do only change what's necessary, i.e. don't perform a complete build. Furthermore, a better way would be to move it out of the `material` folder, either to the root or to `src`. It's actually wrong in the `material` folder, it shouldn't be there at all, as the file names have already been replaced during the build.

Stanzilla:
Better?

squidfunk:
Are you sure?

```
Error: Cannot find module '/home/runner/work/mkdocs-material/mkdocs-material/file_manifest.json
```

Doesn't look like a .gitignore error to me.

squidfunk:
Fixed in 633bcac.

PR #1381, opened 2019-12-12
Fix URL on revision-date.md

coliff:
Currently the first link doesn’t work because they are both referencing [1]


squidfunk:
Thanks!

PR #1383, opened 2019-12-12
provide an alt tag for the logo img.

dajohi:


squidfunk:
Thanks for your PR. In both cases the logo is wrapped inside a link with `title="{{ config.site_name }}"`, so the screen reader will most likely read out the link. IMO it makes no sense to set an `alt` tag here.

Stanzilla:
I guess it's more a out of principle thing since Google Lighthouse complains about it.

squidfunk:
Superseded by #1391

PR #1390, opened 2019-12-18
Separate plugin docs from extensions

martinbira:
Introducing a new section `Plugins` for plugin documentation such as revision date and minify HTML. Update Getting started to have Plugins in a similar format as Extensions.


squidfunk:
Looks great! Thanks for your work!

PR #1391, opened 2019-12-18
a few accessibility fixes from lighthouse report

Stanzilla:
as per gitter

### DIFF
--- a/src/assets/fonts/material-icons.css
+++ b/src/assets/fonts/material-icons.css
@@ -16,6 +16,7 @@
   font-family: "Material Icons";
   font-style: normal;
   font-weight: 400;
+  font-display: swap;
   src: local("Material Icons"),
        local("MaterialIcons-Regular"),
        url("specimen/MaterialIcons-Regular.woff2") format("woff2"),
--- a/src/assets/fonts/material-icons.css
+++ b/src/assets/fonts/material-icons.css
@@ -13,11 +13,9 @@
  */
 
 @font-face {
-  font-display: swap;
   font-family: "Material Icons";
   font-style: normal;
   font-weight: 400;
-  font-display: swap;
   src: local("Material Icons"),
        local("MaterialIcons-Regular"),
        url("specimen/MaterialIcons-Regular.woff2") format("woff2"),
--- a/src/assets/fonts/material-icons.css
+++ b/src/assets/fonts/material-icons.css
@@ -13,6 +13,7 @@
  */
 
 @font-face {
+  font-display: swap;
   font-family: "Material Icons";
   font-style: normal;
   font-weight: 400;
--- a/src/assets/stylesheets/base/_icons.scss
+++ b/src/assets/stylesheets/base/_icons.scss
@@ -28,8 +28,8 @@
 
 // Icon placeholders
 %md-icon {
-  font-family: "Material Icons";
   font-display: swap;
+  font-family: "Material Icons";
   font-style: normal;
   font-variant: normal;
   font-weight: normal;
--- a/src/assets/stylesheets/base/_icons.scss
+++ b/src/assets/stylesheets/base/_icons.scss
@@ -28,7 +28,6 @@
 
 // Icon placeholders
 %md-icon {
-  font-display: swap;
   font-family: "Material Icons";
   font-style: normal;
   font-variant: normal;
--- a/src/assets/stylesheets/base/_icons.scss
+++ b/src/assets/stylesheets/base/_icons.scss
@@ -29,13 +29,13 @@
 // Icon placeholders
 %md-icon {
   font-family: "Material Icons";
+  font-display: swap;
   font-style: normal;
   font-variant: normal;
   font-weight: normal;
   line-height: 1;
   text-transform: none;
   white-space: nowrap;
-  speak: none;
   word-wrap: normal;
   direction: ltr;
 
--- a/src/base.html
+++ b/src/base.html
@@ -145,7 +145,7 @@
             href="https://fonts.googleapis.com/css?family={{
               font.text | replace(' ', '+') + ':300,400,400i,700%7C' +
               font.code | replace(' ', '+')
-            }}&display=swap" />
+            }}&display=fallback" />
         <style>
           body, input {
             font-family: "{{ font.text }}", "Helvetica Neue",
--- a/src/base.html
+++ b/src/base.html
@@ -165,7 +165,7 @@
 
     <!-- Progressive Web App Manifest -->
     {% if config.extra.manifest %}
-      <link rel="manifest" href="{{ config.extra.manifest | url }}">
+      <link rel="manifest" type="application/manifest+json" href="{{ config.extra.manifest | url }}" crossorigin="use-credentials">
     {% endif %}
 
     <!-- Custom stylesheets -->
--- a/src/base.html
+++ b/src/base.html
@@ -148,12 +148,10 @@
             }}&display=swap" />
         <style>
           body, input {
-            font-display: swap;
             font-family: "{{ font.text }}", "Helvetica Neue",
               Helvetica, Arial, sans-serif;
           }
           pre, code, kbd {
-            font-display: swap;
             font-family: "{{ font.code }}", "Courier New",
               Courier, monospace;
           }
--- a/src/base.html
+++ b/src/base.html
@@ -145,7 +145,7 @@
             href="https://fonts.googleapis.com/css?family={{
               font.text | replace(' ', '+') + ':300,400,400i,700%7C' +
               font.code | replace(' ', '+')
-            }}&display=fallback" />
+            }}&display=swap" />
         <style>
           body, input {
             font-family: "{{ font.text }}", "Helvetica Neue",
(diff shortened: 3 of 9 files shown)


src/assets/fonts/material-icons.css
```diff
@@ -16,6 +16,7 @@
   font-family: "Material Icons";
   font-style: normal;
   font-weight: 400;
+  font-display: swap;
```
coliff: I think font-display: swap might not be the right option for icon fonts (the fallback font would be of no use to users)

Stanzilla: Oh, that is actually a good argument, and something Google does not consider in their lighthouse checks.

src/partials/social.html
```diff
@@ -26,7 +26,7 @@
     <link rel="stylesheet" type="text/css"
         href="{{ 'assets/fonts/font-awesome.css' | url }}" />
     {% for social in config.extra.social %}
-      <a href="{{ social.link }}" class="md-footer-social__link
+      <a href="{{ social.link }}" rel="noopener" title="{{ social.type }}" class="md-footer-social__link
```
coliff: If a link here doesn't open in a new tab/page then it doesn't need rel="noopener"

src/assets/fonts/material-icons.css
squidfunk: As a general note: we shouldn't optimize for 100% lighthouse score. Increasing the score by providing better accessibility or faster load time is great, but some things make no sense: https://www.matuzo.at/blog/building-the-most-inaccessible-site-possible-with-a-perfect-lighthouse-score/

src/partials/social.html
squidfunk: I also remember it being a mitigation for `target=_blank`

Stanzilla: Oh indeed, I have external links set to open in a new window in my fork, so I guess I can drop that here.

src/base.html
```diff
@@ -165,7 +167,7 @@
 
     <!-- Progressive Web App Manifest -->
     {% if config.extra.manifest %}
-      <link rel="manifest" href="{{ config.extra.manifest | url }}">
+      <link rel="manifest" type="application/manifest+json" href="{{ config.extra.manifest | url }}" crossorigin="use-credentials">
```
squidfunk: Could you elaborate on this change?

src/assets/fonts/material-icons.css
Stanzilla: Ref: 
- https://github.com/FortAwesome/Font-Awesome/issues/14387
- https://www.zachleat.com/web/font-display-icon-fonts/
- https://stackoverflow.com/questions/49461308/correct-font-display-value-for-icon-fonts

src/base.html
Stanzilla: yes so sometimes (for us because of corp auth) the mimetype of the file was not properly set and crossorigin makes sure it works with CORS (https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin)

src/partials/social.html
Stanzilla: @squidfunk I remember you not being the biggest fan of `target="_blank"` but I think it makes sense for social links because of tracking etc, would you be open to adding that to just the social partial then?

src/base.html
squidfunk: I would assume that a manifest file would be hosted on the same domain, so is there really need for CORS?

Stanzilla: For us there was because of a redirect, not sure if it's generally useful or having it "doesn't hurt" either. I can just remove it if you want.

src/assets/fonts/material-icons.css
Stanzilla: I filed https://github.com/GoogleChrome/lighthouse/issues/10127

src/assets/stylesheets/base/_icons.scss
```diff
@@ -35,7 +35,6 @@
   line-height: 1;
   text-transform: none;
   white-space: nowrap;
-  speak: none;
```
Stanzilla: to be clear, I removed this because it's not actually a valid CSS rule. I did not replace it with a value of the (new?) `speak-as` because there is no equivalent to `none` in there. Ref: https://www.w3.org/TR/css-counter-styles-3/#descdef-counter-style-speak-as

src/base.html
squidfunk: I'm just unsure whether this might be a potential security leak. From my knowledge `use-credentials` means: send cookies to the foreign host.

src/assets/stylesheets/base/_icons.scss
squidfunk: You're right, its not valid and we should remove it. I included it originally because it was part of the original Material Design icon classes. `speak-as` is a descriptor, not a property, and thus only valid within `@counter-style` at rules.

src/base.html
```diff
@@ -145,13 +145,15 @@
...
-            }}&display=fallback" />
+            }}&display=swap" />
         <style>
           body, input {
+            font-display: swap;
             font-family: "{{ font.text }}", "Helvetica Neue",
               Helvetica, Arial, sans-serif;
           }
           pre, code, kbd {
+            font-display: swap;
```
coliff: I don't think this needs to be defined in the style block here... the `font-display:` is just for the @font-face rule

src/base.html
Stanzilla: That is correct, yeah. Not sure if leak though? I just checked and the above linked spec even mentions the requirement for manifest even if it's on the same domain.

```
The use-credentials value must be used when fetching a manifest that requires credentials, even if the file is from the same origin.
```

src/partials/social.html
squidfunk: The reason why we removed it is [here](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/). I guess we could add it back to the social links. I don't know. Does somebody else has an opinion on this?

src/base.html
squidfunk: Correct, it's [a descriptor](https://www.w3.org/TR/css-fonts-4/#font-display-desc). We should remove it from here.

src/partials/social.html
Stanzilla: That concern is fixed by also having noopener set as well though, no?

coliff: I personally don't use the social links in MkDocs-Material, but if I did I'd prefer them to open in a new tab. So adding `target="_blank" rel="noopener"` to just the social links gets a 👍  from me.

src/base.html
```diff
@@ -145,7 +145,7 @@
             href="https://fonts.googleapis.com/css?family={{
               font.text | replace(' ', '+') + ':300,400,400i,700%7C' +
               font.code | replace(' ', '+')
-            }}&display=fallback" />
+            }}&display=swap" />
```
squidfunk: https://developers.google.com/web/updates/2016/02/font-display#swap
https://developers.google.com/web/updates/2016/02/font-display#fallback

Stanzilla: Okay since this body text, we'll want to keep fallback here

src/partials/social.html
squidfunk: Ok, then let's go ahead and add `target=_blank` back. I just wanted another opinion, as I'm comfortable with the current solution, but that may be due to the intrinsic blindness I have when I see a project using this theme.

src/assets/fonts/material-icons.css
Stanzilla: Already got an answer, https://github.com/GoogleChrome/lighthouse/issues/10127#issuecomment-567093362

So it is not actually looking for `swap` but for any flag, the maintainer also commented on the FontAwesome issue about this, informing them of that.

So two choices: 
- We set it to block for our icons fonts
- We don't change it and hope that FA and MaterialIcons adopt the change. I would create an issue on the MI repo as well.

src/partials/social.html
Stanzilla: Alright, added! What about the footer links? Remove noopener?

squidfunk: Well, to be consistent, I would also add `target=_blank` + `noopener` back to the footer links.

src/base.html
squidfunk: As I understand it's relevant for when HTTP Basic Auth is required, so I'm ok with this.

src/partials/header.html
```diff
@@ -31,11 +31,12 @@
       <div class="md-flex__cell md-flex__cell--shrink">
         <a href="{{ config.site_url | default(nav.homepage.url, true) | url }}"
             title="{{ config.site_name }}"
+            aria-label="{{ config.site_name }}"
```
squidfunk: Adding an ARIA label + the alt tag on the image - how is this handled by screen readers?

Stanzilla: The Windows Narrator just says "sitename" once.

Stanzilla: And lighthouse complained about there missing a label to the a because the img inside of it was not enough, for whatever reason.

squidfunk: Okay great, so this is ready to merge?

Stanzilla: Yes from my side!

PR #1405, opened 2019-12-25
Add ServiceWorker

Stanzilla:
As per Gitter, initial proposal to get a discussion started on what we would want out of a serviceworker for a static website. Note: Missing the hook from manifest.json because the theme does not come with one by default.


src/manifest.json
```diff
@@ -0,0 +1,18 @@
+{
+  "short_name": "mkdocs-material",
+  "name": "mkdocs-material",
+  "start_url": "/?utm_source=web_app_manifest",
+  "icons": [
+    {
+      "src": "assets/images/favicon.png",
```
Stanzilla: would love to use mkdocs.yml parameters here but not sure how.

material/partials/integrations/serviceworker.html
```diff
@@ -0,0 +1,10 @@
+{#-
+  This file was automatically generated - do not edit
+-#}
+<script>
+  if ("serviceWorker" in navigator) {
+    window.addEventListener("load", () => {
```
RDIL: This would need to go through babel, since its es6

Stanzilla: Oh, maybe? Note sure what the plans are for IE11 support? https://caniuse.com/#feat=arrow-functions

squidfunk: Jep, this needs to be a proper `function`, not an arrow function.

Stanzilla: Done!

Stanzilla:
Also having tons of problems with it on my fork, if your site lives in a directory and not on the root of the domain, the theme and serviceworkers really disagree about paths a lot.

RDIL:
Maybe it shouldn't be added then - I am a bit biased to talk however as I despise service workers

Stanzilla:
I'm mostly of the opinion that I have no idea what I'm doing and that it can be fixed. What's your grief with them though?

Stanzilla:
Maybe use https://github.com/arthurbergmz/webpack-pwa-manifest?

Stanzilla:
Probably moving over to https://developers.google.com/web/tools/workbox/guides/get-started that is so much better than anything else so far.

src/manifest.json
squidfunk: I think that's not possible, since Jinja only accepts `.html` and not `.json` files.

src/serviceworker.js
```diff
@@ -0,0 +1,140 @@
...
+          // This catch will handle exceptions thrown from the fetch operation.
+          // Note that a HTTP error response will NOT trigger an exception.
+          // It will return a normal response object that has the error code set
+          console.error("Fetching failed:", error)
+
+          throw error
+        })
+    })
+  )
+})
```
squidfunk: This looks super-generic. Is there maybe an out-of-the-box solution which we can configure and use?

Stanzilla:
I would love to switch to Google Workbox (https://developers.google.com/web/tools/workbox) and their webpack module https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin

squidfunk:
Revisiting my last comment: any new insights? I still think this might be an excellent use case for a plugin.

Stanzilla:
I still think we should use Workbox since it offers the most generic solution

squidfunk:
As I read from the docs the main use cases are for caching files. I guess we could just specify everything under the docs and assets directory and probably Google Fonts and should have Offline Support given that no external resources are used. Haven’t used or implemented an SW yet.

Stanzilla:
Yup, pretty much what I did. My config looks like this:

```js
/* eslint-disable require-jsdoc */
/* eslint-disable no-unused-vars */
/* eslint-disable array-callback-return */
/* eslint-disable no-underscore-dangle */
/* eslint-disable no-undef */

workbox.precaching.cleanupOutdatedCaches()
workbox.precaching.precacheAndRoute(self.__precacheManifest || [])
workbox.core.clientsClaim()
workbox.googleAnalytics.initialize()

workbox.routing.registerRoute(
  /\.(?:js|json)$/,
  new workbox.strategies.StaleWhileRevalidate({
    cacheName: "js-cache",
    fetchOptions: {
      credentials: "include"
    }
  })
)

function cacheKeyWillBeUsed({ request }) {
  const url = new URL(request.url)
  url.pathname = url.pathname.replace(/\/index\.html$/, "/")
  url.pathname = url.pathname.replace(/\.html$/, "/")
  return url.href
}

const networkFirstStrategy = new workbox.strategies.NetworkFirst({
  cacheName: "docs-cache",
  fetchOptions: {
    credentials: "include"
  },
  plugins: [{ cacheKeyWillBeUsed }]
})

const navigationRoute = new workbox.routing.NavigationRoute(networkFirstStrategy)
workbox.routing.registerRoute(navigationRoute)
workbox.routing.setCatchHandler(({ event }) => {
  switch (event.request.destination) {
    case "document":
      return caches.match(workbox.precaching.getCacheKeyForURL("offline.html"))

    default:
      // If we don't have a fallback, just return an error response.
      return Response.error()
  }
})

workbox.routing.registerRoute(
  /\.css$/,
  // Use cache but update in the background.
  new workbox.strategies.StaleWhileRevalidate({
    cacheName: "css-cache",
    fetchOptions: {
      credentials: "include"
    }
  })
)

workbox.routing.registerRoute(
  /\.(?:png|jpg|jpeg|svg|gif|ico|mp4)$/,
  // Use the cache if it's available.
  new workbox.strategies.CacheFirst({
    cacheName: "image-cache",
    fetchOptions: {
      credentials: "include"
    },
    plugins: [
      new workbox.expiration.Plugin({
        // Cache only 50 images.
        maxEntries: 50,
        // Cache for a maximum of a day.
        maxAgeSeconds: 24 * 60 * 60
      })
    ]
  })
)

// Cache the Google Fonts stylesheets with a stale-while-revalidate strategy.
workbox.routing.registerRoute(/.*(?:googleapis)\.com.*$/,
  new workbox.strategies.StaleWhileRevalidate({
    cacheName: "googleapis-cache"
  })
)

// Cache the underlying font files with a cache-first strategy for 1 year.
workbox.routing.registerRoute(
  /^https:\/\/fonts\.gstatic\.com/,
  new workbox.strategies.CacheFirst({
    cacheName: "google-fonts-webfonts",
    plugins: [
      new workbox.cacheableResponse.Plugin({
        statuses: [0, 200]
      }),
      new workbox.expiration.Plugin({
        maxAgeSeconds: 60 * 60 * 24 * 365,
        maxEntries: 30
      })
    ]
  })
)

addEventListener("message", event => {
  if (event.data && event.data.type === "SKIP_WAITING") {
    skipWaiting()
  }
})
```

RDIL:
Personally, I am against service workers because I am constantly updating material and my docs, so I want them to always be up to date.

squidfunk:
Yeah I’m also quite unsure whether a general implementation caters to most use cases. Quite undecided. Caching assets should work as long as they have content hashes attached which is probably not the case for doc-related assets.

Stanzilla:
That's the cool thing about it, you can just cache what you want. images that never change and assets like fonts are a good candidate.

squidfunk:
In the end it’s up to the user to decide what to cache which sounds like an excellent fit for theme customization. We cannot anticipate what assets are candidates for caching, except for all the theme related stuff but that’s only half of the story.

squidfunk:
Is it okay if we close this for now? The PR was a really good start for getting a discussion going but I don’t see a clear direction and with v5 on the horizon I would say at this time we’d leave it up as an opportunity for customization. However, thanks for your time and work on this topic!

Stanzilla:
Yea

Issue #1306, opened 2019-10-14
Material 5.0.0 Status


squidfunk:
> __TL;DR__ This issue is meant for updates and discussions on the progress of the rework of the theme. Please feel free to checkout and try the [`refactor/rxjs-typescript`](https://github.com/squidfunk/mkdocs-material/tree/refactor/rxjs-typescript) branch and comment on the current state of development.

## Description

The last major releases of Material for MkDocs were mainly issued for compatibility reasons. While version 2 and 3 were related to backward incompatible changes of MkDocs (0.17.1 and 1.0 respectively), version 4 was released due to backward incompatible changes related to users with Chinese system languages (#911). The next iteration, version 5, is a substantial rewrite of the underlying JavaScript functionality to a TypeScript and RxJS-based architecture. This will yield the following major benefits:

1. __All features that are provided by the theme (e.g. the sidebars, tabs, search, etc.) will be observable by 3rd party JavaScript__. This allows for very easy extension and customization. This will address issues like #1102 where users want to extend the functionality of the theme without re-compiling or forking it. Also, disabling functionality should be easier.

2. __Improving search performance__. Search will be moved to a web worker and exhibit better caching and re-building behavior. This will mitigate problems were the UI freezes due to large search indexes which pre-building on the server-side only partly solves. Furthermore, search will be re-architected to fit into the plugin concept introduced by MkDocs 1.x.

3. __Color customization will be re-implemented using CSS variables__. Changing the color to your custom brand colors will be even simpler and the theme will become smaller (as the palette CSS file will not be necessary anymore). [Support for CSS variables](https://caniuse.com/#feat=css-variables) is pretty great by now, but there are some browsers that don't support it, respectively IE 11, Opera mini, QQ browser and Baidu browser. These browsers will receive a (probably neutral) fallback color. If you must support those browsers, you can still extend the theme and customize the build will custom CSS colors as already described in the customization guide.

There might also be some slight face lifting, but the main aspects of Material 5 is the introduction of a more modern architecture. The rewrite will also make the theme testable, so unit tests can and will be written to ensure functionality.

I will document my progress here and there will be a beta phase similar to before the release of 1.0.0 (#46). Some lists with things that need to be done will follow.

## Evaluate

- [x] ~Consider [flexsearch](https://github.com/nextapps-de/flexsearch) as an alternative to lunr.js~

#### General

- [x] Move everything except JavaScript compilation out of Webpack into Makefile
[...]

PR #1409, opened 2019-12-25
CSS Variables

Stanzilla:
This is probably how I would start the conversion, do note that I added a dark-mode root element already but it's unused so far.


squidfunk:
Thanks for your work! However, we definitely need to move dark mode and theme switching out of this PR, as I've prioritized them post-v5 (see #1306). The main reason is that I haven't decided where to put the switch and what the actual semantics should be, e.g. light mode, dark mode, automatic mode (based on preference), etc.

Stanzilla:
Will do

Stanzilla:
Is the current state what you had in mind or want something else?

squidfunk:
I'm still on vacation, haven't had the time to check it out. I think I'll come back with some feedback in the middle of January.

src/assets/stylesheets/app-palette.scss
```diff
@@ -128,7 +128,7 @@ button[data-md-color-accent] {
 
       // Repository containing source
       .md-nav__source {
-        background-color: mix($color, $md-color-black, 75%);
+        background-color: $color; // TODO: https://codyhouse.co/blog/post/how-to-combine-sass-color-functions-and-css-variables mix($color, var(--md-color-text), 75%);
```
squidfunk: We should use `hsl` as a color space, because then we can easily mix colors and deduce shades.

Stanzilla: Yup, that's what the link is about, just not sure which parts we should implement as custom functions yet.

squidfunk: I read the article and must say, that it sounds very complicated for what we're trying to achieve and it doesn't solve all of our problems, namely that users should be able to override colors with their own CSS variables. IMHO the UX should be:

1. Set one of the default Material colors via `mkdocs.yml` - this should be possible as of now
2. Override colors via CSS variables on the `:root` pseudo class or any element via extra CSS

We cannot use any SASS functions or mixins implementing this approach, because users should be able to override colors without compiling the whole theme again. This would mean that users would be obliged to define the `H`, `S` and `L` parts of the color separately, so all shades can be deduced. I guess that's too complicated and will lead to a lot of issues, so we should maybe (at the current moment) stick to `RGB`. The property definitions using the `mix` function should be normalized as separate CSS variables.

src/assets/stylesheets/_config.scss
```diff
@@ -75,6 +79,54 @@ $md-color-white--lighter:             hsla(0, 0%, 100%, 0.3) !default;
...
 $md-color-white--transparent:         hsla(0, 0%, 100%, 0) !default;
 
+// ----------------------------------------------------------------------------
+// CSS Variables: base colors
+// ----------------------------------------------------------------------------
+
+:root {
+  // Primary and accent colors
+  --md-color-primary:                         #{$md-color-primary};
+  --md-color-primary-values:                  #{hex2rgb($md-color-primary)};
```
squidfunk: This looks like a smell to me. Where do we need those values? For different opacities? We should define those color-opacity combinations as variables.

Stanzilla: Here https://github.com/squidfunk/mkdocs-material/pull/1409/files#diff-5c58c014d3bb6c0f3026ce46fab89998R126 to replace opacify()

squidfunk: I'll check out your changes when I'm back and will take care of merging. Thanks again for your support!

squidfunk: I will proceed on this PR when I finished all the JavaScript migration. Sorry for the delay!

squidfunk:
This PR is GOLD! Thanks for doing all the leg work!

squidfunk:
I refactored the variable naming a little so we can better distinguish two cases: normal back- and foreground (black/white) and primary/accent color + text, as those may need to be styled separately. Think white text color on blue background in the header even though in dark mode.

https://github.com/squidfunk/mkdocs-material/blob/309c2b7c03004df3dee2f4d240a38f4928afd4ec/src/assets/stylesheets/base/_colors.scss#L27-L56

Stanzilla:
Yeah, good idea. Also we probably still want some variables that are actually black/white in both modes

squidfunk:
Yep, we can integrate that. However, I also excluded a dark mode for now as this definitely needs some more work. Let's tackle dark mode in 5.x!

squidfunk:
I love it!

Ohne Titel

Stanzilla:
Ohh that is really cool

facelessuser:
😍 This is really cool. I'm excited to play with this.

PR #1435, opened 2020-01-20
Fix improperly calculated Sidebar height

DavBfr:
I think this line got duplicated during a merge or a rebase and should not be there.
It hides a part of the TOC on wired conditions.

Screen Shot 2020-01-20 at 14 42 25

### DIFF
--- a/src/assets/javascripts/components/Material/Sidebar/Position.js
+++ b/src/assets/javascripts/components/Material/Sidebar/Position.js
@@ -106,7 +106,6 @@ export default class Position {
 
     /* Calculate new offset and height */
     const height = visible - bounds.top
-                 - Math.max(0, this.offset_ - offset)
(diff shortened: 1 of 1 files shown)


squidfunk:
This line hasn't been touched for the last 3 years, see the output of [blame](https://github.com/squidfunk/mkdocs-material/blame/611c8404ce24d7749817a693c41ee1a7a118c619/src/assets/javascripts/components/Material/Sidebar/Position.js#L109).  I would guess that this is not the root cause of the problem, as I think it would have been reported by now.

You can open an issue and provide the necessary details/context as stated in the issue template in order for us to reproduce the problem.

squidfunk:
Closing for the reasons stated. A reproducible test case would be a great starting point in finding and fixing this error. Feel free to open an issue if the error persists!

PR #1437, opened 2020-01-23
Update _codehilite.scss

Stanzilla:
Property is ignored due to the display. With 'display: block', vertical-align should not be used.


squidfunk:
Thanks! That’s correct. `vertical-align` is only relevant for inline elements. Pitty me 😊

PR #1458, opened 2020-02-14
Include Plugin modules in docker image

### LINKED ISSUE #1455: Docker container doesn't include mkdocs-git-revision-date-localized-plugin module
__I checked that...__

## Description

The docker container doesn't include the mkdocs-git-revision-date-localized-plugin module, so using it to build a site that uses that plugin will fail.  I believe adding this to requirements.txt would fix the issue, and be consistent with how the minify plugin, though it would mean the plugin would get installed for everyone installing mkdocs-material (though, again, that's already the case with the minify plugin)

### Expected behavior

running build with squidfunk/mkdocs-material builds the site

### Actual behavior

```
PS C:\Users\bajurny\Documents\GitHub\mkdocs-demo> docker run --rm -v ${PWD}:/docs squidfunk/mkdocs-material build
ERROR   -  Config value: 'plugins'. Error: The "git-revision-date-localized" plugin is not installed 

Aborted with 1 Configuration Errors!
```

1. Include `git-revision-date-localized` in plugins in mkdocs.yml
```
plugins:
  - search
  - git-revision-date-localized
```
2. Execute build using docker container
`docker run --rm -v ${PWD}:/docs squidfunk/mkdocs-material build`

### Package versions

```
PS C:\Users\bajurny\Documents\GitHub\mkdocs-demo> docker pull squidfunk/mkdocs-material
Using default tag: latest
latest: Pulling from squidfunk/mkdocs-material
Digest: sha256:532ef540a52aff90be2a3c861ff39bc0b35651cdaadda8baafa73a76b8ae0779
Status: Image is up to date for squidfunk/mkdocs-material:latest
docker.io/squidfunk/mkdocs-material:latest
PS C:\Users\bajurny\Documents\GitHub\mkdocs-demo> docker image ls
REPOSITORY                                 TAG                            IMAGE ID            CREATED             SIZE
squidfunk/mkdocs-material                  latest                         bd6005f44296        2 days ago          117MB
```
Running the shell against the running container and executing `pip list`:
```
PS C:\Users\bajurny\Documents\GitHub\mkdocs-demo> docker run --rm -it --entrypoint /bin/sh squidfunk/mkdocs-material
/docs # pip list
Package              Version
-------------------- -------
Click                7.0
Jinja2               3.0.0a1
jsmin                2.2.2
Markdown             3.2
mkdocs               1.0.4
mkdocs-material      4.6.2
pip                  19.1.1
Pygments             2.5.2
PyYAML               5.3
setuptools           41.0.1
six                  1.14.0
wheel                0.33.3
You should consider upgrading via the 'pip install --upgrade pip' command.
/docs #
```

### Project configuration

``` yaml
plugins:
  - search
  - git-revision-date-localized
```

FISHMANPET:
This fixes #1455 per discussion there.

First, I've updated the docker image to the latest python alpine image at the time, 3.8.1-alpine3.11.  Not sure what kind of testing would be required to ensure this is working correctly, but with it I was able to build a site and run the built-in server via the container.  I was also able to run the same "test" command used in your CI publish task.  Not sure if there's any deeper testing you'd want to do there.

I added a pip install command to the existing run block for the two plugin modules that the theme supported.  I could have added an additional RUN step instead of adding onto the existing one, I decided to add on since you'd already grouped it into one mega command instead of separating them.

I've also removed mkdocs-minify-plugin from requirements.txt, because it's not a requirement for the base module.  The module itself is using requirements.txt, so this will change the dependencies of the installed module as well.  You mentioned in the issue to add the install commands at the same place as the existing pip install, except there's no pip install in the current Dockerfile, the requirements are installed as part of running `setup.py install`.

Finally, I added a .dockerignore file, [based on this comment on a docker build issue](https://github.com/docker/toolbox/issues/613#issuecomment-319632983).  Without this, the docker engine reads every file in the directory to pass it to the build daemon, and there are thousands of node files that ultimately aren't used by docker, so this speeds up the build of the image significantly, at least on windows.

The image grew a bit, the current image is 117MB on my machine, whereas this one is 184 after building.  The python base image grew from 79.1MB to 109MB as a result of the version change.  The rest of the growth is from the `mkdocs_git_revision_date_localized_plugin` and it's dependencies, mostly `babel` which is 28.2MB on disk.

Happy to make any revisions to this that you'd like.


squidfunk:
I tested your changes and everything seems to work. Just released 4.6.3. which also includes a new Docker image.

PR #1459, opened 2020-02-14
Add support pymdownx.tabbed via pymdown-extensions 7.0b1

facelessuser:
Adds support for the new general purpose tab extension called Tabbed. This requires pymdown-extensions 7.0b1 which adds this support.

Style needs to built, but I was doing this on Windows and it created a mess of line ending things. I can maybe rebuild the styles on mac.


facelessuser:
Rebased and rebuilt theme. Should be good to go.

src/assets/stylesheets/extensions/pymdown/_tabbed.scss
```diff
@@ -59,8 +59,26 @@
...
+            margin-bottom: px2rem(12px);
+          }
+
+          > {
+            div.highlight,
+            div.codehilite,
+            pre,
+            table.highlighttable,
+            table.codehilitetable {
+              &:last-child:first-child {
```
squidfunk: I think we can use `:only-child` here, as supports seems [to be good](https://developer.mozilla.org/de/docs/Web/CSS/:only-child).

Also, can’t we drop `div.higlight` and `div.codehilite` selectors due to Markdown 3.2?

squidfunk:
Furthermore, this should be frictionless to upgrade to unless people changed the superfences styles with custom CSS, or am I missing something?

src/assets/stylesheets/extensions/pymdown/_tabbed.scss
facelessuser: :man_facepalming: yup. Don't know why I fixated on doing it like this.

facelessuser:
The only thing in 7.0 is Tabbed. But  if it seems solid, then we are good. As far as Material is concerned, it should go seemless to release them together. Only if someone has some custom CSS will they be impacted. And since I went ahead and released a beta, you could release tomorrow and Material would just pull the beta. Since it is prerelease, only things explicitly requesting it will get it.

Anyone on an old material when I push the official 7 may get impacted, which is why it is a major release even if it only includes 1 thing. Because I'm going to break old SuperFences styles as we work towards retiring the old code tab mechanics.

As always, I'm open to alternatives, but that is my current plan.

src/assets/stylesheets/extensions/pymdown/_tabbed.scss
squidfunk: ... a conversation of two people, both of which have implemented a selector engine 😂

squidfunk:
Ah right, I missed the syntax changes. It’s 7 in the morning over here! 😴

We should add them to the migration guide then. My plan is to collect feedback with the RC and release a step-by-step guide as there will be an increasing amount of necessary changes with an increased customization level.

src/assets/stylesheets/extensions/pymdown/_tabbed.scss
facelessuser: I know, that is the most embarrassing part...I implemented `:only-child`...

As far as `div.highlight` and `div.codehilite` is concerned, this is one case where we cannot drop it. For styling code blocks, all we care about is the `<pre><code>` part, but in this case, we want to know if the code block is an only child under the tab, and Pygments still wraps the `<pre><code>` in a `<div>`. It's just `<div><pre><code>` now instead of `<div><pre>`. And since we need to know that the only child `<div>` is a code `<div>`, we need to check the class because we can't just go `div:has(pre)`.

squidfunk: You’re right! That’s also why the child selector is absolutely necessary here.

squidfunk: ... but we could drop the tag parts (`table`, `div`) of the selectors. Probably even more future-proof.

facelessuser:
No worries, it's late here. I'm probably done until the morning. I went ahead and pushed the `:only-child` change.

One small off topic note: with Material 5, maybe it will be fixed in the RC, but by default I get the announcement bar test in my theme when I try to use Material 5. I have to provide my own `main.html` silence it.

squidfunk:
Yep, that needs to go somewhere else. It’s just handy for development but I’ll remove it before the RC!

src/assets/stylesheets/extensions/pymdown/_tabbed.scss
facelessuser: Agreed, table and div are unnecessary.

squidfunk:
I'll take care of the selector optimizations (removing `div` and `table`).

squidfunk:
Fixed in 0d9fbcce

squidfunk:
@facelessuser the default announcement bar is gone

PR #1467, opened 2020-02-19
Inline numbers

facelessuser:
This adds support for support for pymdownx's inline highlight tables for SuperFences. While the styling has been added, I have not forced any of Materials tables to use it. To enable it, you would simply include:

```yaml
  - pymdownx.superfences:
  - pymdownx.highlight:
      linenums_style: pymdownx-inline
```

Then, when you use lines with fences, you should get the new style:

````
```python linenums="1"
import something
```
````

I figure you may want to tweak the style or play around with it, so I consider this just the initial CSS.


facelessuser:
This is pulled on the wrong branch.

PR #1468, opened 2020-02-19
Inline numbers

facelessuser:
This adds support for support for pymdownx's inline highlight tables for SuperFences. While the styling has been added, I have not forced any of Materials tables to use it. To enable it, you would simply include:

```yaml
  - pymdownx.superfences:
  - pymdownx.highlight:
      linenums_style: pymdownx-inline
```

Then, when you use lines with fences, you should get the new style:

````
```python linenums="1"
import something
```
````

I figure you may want to tweak the style or play around with it, so I consider this just the initial CSS.


squidfunk:
Many thanks! I'll revisit the styles before the RC.

facelessuser:
Just as a note. In Firefox mobile, the use of sticky on the line numbers can cause some jitter on the update of sticky positions. I just think Firefox mobile doesn't handle that many sticky elements well. All other browsers I've tried update fine. At the very least, if this is an issue, you can just disable sticky on the numbers and let it scroll as normal inline elements. They stay out of the search either way.

squidfunk:
Thanks for reporting. Wouldn't consider this being a big problem.

facelessuser:
Cool. I've used it with sticky and without in the past, and both are fine. I just forgot about this Firefox quirk until today. But now if you see lagged sticky updates on a page, you'll at least know what the issue is and how to fix it if it is a problem. I still think without sticky that it is my preferred line number format.

squidfunk:
This is actually one of the reasons I don't like Firefox. Their rendering engine is worse than Webkit, especially when it comes to compositing stacking contexts. I hope they catch up some time, because, well, Chrome, Google.

jeremyong:
> In Firefox mobile, the use of sticky on the line numbers can cause some jitter on the update of sticky positions.

Doesn't firefox mobile use webkit internally? Or are you referring to Android (not sure if they use their own renderer or not)?

facelessuser:
As far as I know, Firefox still use their own engine.

Stanzilla:
It only uses WebKit on iOS because you can't use anything else there.

jeremyong:
Right, not using WebKit on IOS also implies you can't have a functional JIT compiling Javascript runtime which would be a non-starter.

PR #1476, opened 2020-02-26
const instead of var for redirect script

Stanzilla:
Use const instead of var, the other style changes actually match your stylelint config.

### DIFF
--- a/src/base.html
+++ b/src/base.html
@@ -49,9 +49,8 @@
       <!-- Redirect -->
       {% if page and page.meta and page.meta.redirect %}
         <script>
-          var anchor = window.location.hash.substr(1)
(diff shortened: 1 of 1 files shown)


squidfunk:
It's true that we want to support only modern browsers but I would rather just leave it as a `var` definition. There's no downside, as it's in the global scope anyway and the redirect happens instantly.

Stanzilla:
I really only did the PR because my linter complained :D And I'm still using your stylelint file from master, so feel free to just close if not wanted

squidfunk:
We should add all distribution files to `.stylelintignore` 😉

squidfunk:
Closing for the reasons stated.

Stanzilla:
I'm using ESLint, yeah :)

PR #1480, opened 2020-03-03
Serbian latin translation

majkinetor:


squidfunk:
Thanks! Is this a valid HTML5 language code?

majkinetor:
Thats what I wanted to ask but decided not to :) It looks like its not:

https://gist.github.com/JamieMason/3748498#file-html-languages-txt-L192

It looks like `sr` should be latin (or `sr-Latn-CS`) and cyrlic that is currently `sr` should be `sr-Cyrl-CS`

What should I do ?

majkinetor:
Oh, there is already `sh` here. I'll just close this. Sorry :)

squidfunk:
No worries, sometimes hard to keep track what we already have 😉

PR #1487, opened 2020-03-06
Add preconnect + dns-prefetch for Google Analytics

Stanzilla:
According to Lighthouse this will save 260ms


squidfunk:
260ms? Unlikely, but why not.

PR #1488, opened 2020-03-06
Add title attribute to social links

Stanzilla:


squidfunk:
There's actually no more `social.type` option and I think we shouldn't add one just for the sake of a title. I understand that optimizing for Lighthouse is a thing, but it would mean that the author would have to provide an explicit title for every link. What about adding `aria-hidden`?

Stanzilla:
Hrm does that add a hover tooltip?

Stanzilla:
I honestly don't quite like that change to icon formats, would it be a lot of work to allow the old scenario to work again?

squidfunk:
> I honestly don't quite like that change to icon formats, would it be a lot of work to allow the old scenario to work again?

What exactly don't you like about it?

Stanzilla:
That I lose access to the name variable, I would love to have URL, icon and name available as variables.

squidfunk:
Suppose we introduce a `name` variable - what is the default if somebody doesn't provide it?

Stanzilla:
Can we no longer deduct the name from the icon file name?

squidfunk:
Yes, we could introduce a `name` member for `config.extra.social.*` and use that as a title, but then we have to think about what we use as a name when the author doesn't provide one.

Stanzilla:
the svg file name?

squidfunk:
bfea3315 automatically extracts the domain part from the URL – IMHO this is a better solution than to ask the author to provide a description for every link – does this solve your problem?

Ohne Titel

Stanzilla:
Certainly helps, will get back to you if I need more, thanks!

Stanzilla:
@squidfunk This is not good enough for Lighthouse (https://web.dev/link-name/?utm_source=lighthouse&utm_medium=devtools) want a new issue?

squidfunk:
How is this not good enough? I thought we fixed this?

Stanzilla:
image

squidfunk:
I see, not fixable. We would have to introduce a new option to the social icons to set a name for each link. That’s impractical, do I would disregard Lighthouse recommendations in this case.

Stanzilla:
Why not make the option optional?

squidfunk:
We could but improving blindly for Lighthouse is no path I want to follow.

Stanzilla:
Well people being able to add a little more descriptive texts to their social links would be nice, ignoring Lighthouse for now.

squidfunk:
Yes, but then there will be other requests for even more possible customizations At other places. We have to cut at some point. You forked the theme, so it should be no problem to add it in your fork if desired.

squidfunk:
... you‘re (currently) the only one who wants this feature

PR #1497, opened 2020-03-09
Don't expand tabs in mobile

facelessuser:
Tab interface used to expand to because they only contained code, but
now they can contain anything. When expanding tabs to fill full width
in mobile without code content, it can look very odd.


squidfunk:
Could you provide before/after screenshots?

facelessuser:
Okay, this is assuming we don't stick with what I view as problematic tinting of the non code background. In the before image it looks okay only because we are tinting the tab content (even though it isn't code). But if you don't tint it, there are no borders, and no general indication that the tab has ended. So to be fair, it is problematic because I reset the tint of normal tab backgrounds in my own custom style. But if we plan to no longer tint the background of non-code content, we'll want the borders back and basically treat them no different than we do with admonitions and details.

So maybe this should incorporate changes to fix tabs in general?

Regardless of whether the tint is no longer applied to non-code content or not, the change is harmless and looks fine even without the rest of the tab changes. If we don't want to merge it yet, maybe it can be left open as reminder to not forget this when/if we do remove tint from non-code tabbed content?

Before

before

After:

after

squidfunk:
Let's revisit this topic when I find some time to think about how we could get rid of the tint. The box-shadow discussion on Gitter today was a great starting point, but I'm still not 100% convinced that we have a good solution in place.

facelessuser:
Not a problem. If I have anymore tab ideas, I'll post them here.

facelessuser:
Here's another idea for tab format.  It borrows from the tabbed convention you already have in place.

chrome_CCWxeJHcKe

Now we don't have to use the same primary fg color. You could use the color that table already uses, unless that would get confusing having them both the same. Then you'd have to maybe make it look different somehow. I guess you could bring back the underscore:

chrome_mWAV7sFTpS

Anyways, just some ideas.

squidfunk:
So my proposal looks like this:

Bildschirmfoto 2020-03-10 um 15 03 51

I removed the border and shadow - both can be added back with a single of CSS each, but I find this to better match alternative content which is not a callout like an admonition, details or table. The content is not smaller (like for callouts) and there's no shadow.

Futhermore I unified the spacing of tables, tabs and code blocks so that they align in all combinations:

Bildschirmfoto 2020-03-10 um 14 53 29

squidfunk:
Oh, and tabs don't expand anymore on mobile.

facelessuser:
Interestingly I had a similar proposal I was going to make today with the exception of adding a light border on the top of the tab content that matches other material themes. But I can always do that in my personal repo if I feel strongly about it.

squidfunk:
Okay, so let's start with this proposal. We can iterate on it if necessary after v5 is out.

facelessuser:
Sounds good. Thanks for humoring me in this. I realize a lot of this stuff is very subjective to the person.

squidfunk:
Definitely. I just like to start with the lowest common denominator and adding extra CSS to change appearance is a no-brainer. Closing for now, thanks for your input and research!

squidfunk:
New tabs styled added with b6657bcc btw. Furthermore I refactored code blocks. Since we now always have `pre > code` there was a lot of possibility for unification and I could shave off another kb. Also, line numbers are now the same for inline and tables - I've adapted the styles you submitted for the inline styles.

squidfunk:
+ I think you're right - a thin line works quite well 😊

Bildschirmfoto 2020-03-10 um 16 54 48

Added in e3127489

facelessuser:
Looks perfect :ok_hand:.

PR #1500, opened 2020-03-10
Code fixes

facelessuser:
- Line highlight pad should match the parent
- Float the gutter left to force code and numbers not to wrap
- Provide a background for the gutter


facelessuser:
Wrong branch.

PR #1501, opened 2020-03-10
Fix inline line highlight padding, wrapping, and gutter background

### LINKED ISSUE #1499: Code block styling issues with pymdownx.inline and inline
__I checked that...__

## Description

Issue with 5RC1 inline line highlighting.  Support for inline highlighting is a little broken. Particularly my interest was in support for `pymdownx-inline`, but I've generally investigated and made normal inline also look the same (whether we want to actually do anything with that is up to you).

1. Background is transparent, so when scrolling content underneath, the code shows through.
2. Margins and such are not right with recent changes.
3. Line highlight wraps to next line.
4. Highlight padding is different between table and inine styles, but only table inline padding is used causing inilne code with highlights to have scrollbars.

This was discussed in Gitter.

Fixes here: https://github.com/facelessuser/pymdown-extensions/blob/master/docs/src/scss/_code.scss

Note:

1. Highlight fix is here: https://github.com/facelessuser/pymdown-extensions/blob/master/docs/src/scss/_code.scss#L29 and here: https://github.com/facelessuser/pymdown-extensions/blob/master/docs/src/scss/_code.scss#L78. You could also just make them the same padding and have universal padding.
2. Normal inline consistency changes are here: https://github.com/facelessuser/pymdown-extensions/blob/master/docs/src/scss/_code.scss#L35
2. Pymdownx-inline fixes are here: https://github.com/facelessuser/pymdown-extensions/blob/master/docs/src/scss/_code.scss#L56

"Special" css can be ignored unless you wish to support pygments special class attached to line numbers.

### Expected behavior

Inline, especially with line highlighting, doesn't quite look right.

### Actual behavior

See description

1. ...
2. ...
3. ...

### Package versions

* Python: Any
* MkDocs: Latest
* Material: 5.0.0rc1

### Project configuration

I use `pymdownx-inline` with line numbers though I believe there are issues with normal `inline` as well.

### System information

* OS: Any
* Browser: Any

facelessuser:
- Line highlight pad should match the parent
- Float the gutter left to force code and numbers not to wrap
- Provide a background for the gutter

Fixes #1499

### DIFF
--- a/src/assets/stylesheets/base/_colors.scss
+++ b/src/assets/stylesheets/base/_colors.scss
@@ -51,6 +51,6 @@
   --md-accent-bg-color--light:         var(--md-default-bg-color--light);
 
   // Code block color shades
-  --md-code-bg-color:                  hsla(0, 0%, 92.5%, 0.5);
+  --md-code-bg-color:                  hsla(0, 0%, 96%, 1);
   --md-code-fg-color:                  hsla(200, 18%, 26%, 1);
 }
--- a/src/assets/stylesheets/base/_typeset.scss
+++ b/src/assets/stylesheets/base/_typeset.scss
@@ -224,7 +224,7 @@ kbd {
     > code {
       display: block;
       margin: 0;
-      padding: px2rem(10.5px) px2em(16px, 13.6px);
+      padding: px2rem(10.5px) px2rem(16px);
       overflow: auto;
       word-break: normal;
       box-shadow: none;
--- a/src/assets/stylesheets/base/_typeset.scss
+++ b/src/assets/stylesheets/base/_typeset.scss
@@ -224,7 +224,7 @@ kbd {
     > code {
       display: block;
       margin: 0;
-      padding: px2rem(10.5px) px2rem(16px);
+      padding: px2rem(10.5px) px2em(16px, 13.6px);
       overflow: auto;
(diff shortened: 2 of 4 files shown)


facelessuser:
How do you want to handle the coloring of the gutter background? I basically eye balled it, but it would be nice to derive it

squidfunk:
> How do you want to handle the coloring of the gutter background? I basically eye balled it, but it would be nice to derive it

That is a good question, as the code background has alpha transparency which will not have the intended effect when used on the line numbers. I think we should switch to an opaque background. I removed code background in admonition titles in a recent commit, so I think the transparency is of no use anymore, or am I missing something?

facelessuser:
Results:

aZUgB9HVus

squidfunk:
Thanks for providing the visual aid. Unfortunately the gutter doesn't match the background color, but that should be resolved when switching to opaque colors and CSS variables.

facelessuser:
Root level, stretched code blocks use a different padding, so we have to stretch the highlighting differently for this that case as well. Last commit accounts for this case.

facelessuser:
Sorry for the churn, but I think I got them all:

7kV0SLCo7E

squidfunk:
LGTM, thanks for taking the time to investigate and fix this!

PR #1516, opened 2020-03-18
Adds aria-label to all nav items to make them accessible.

jaimeiniesta:
This ensures navigation landmarks are unique. Related:

https://dequeuniversity.com/rules/axe/3.5/landmark-unique?application=rocketvalidator
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role
https://www.w3.org/TR/WCAG20-TECHS/ARIA14.html

### DIFF
--- a/src/partials/language/en.html
+++ b/src/partials/language/en.html
@@ -27,8 +27,10 @@
   "clipboard.copy": "Copy to clipboard",
   "clipboard.copied": "Copied to clipboard",
   "edit.link.title": "Edit this page",
+  "footer.title": "Pie",
   "footer.previous": "Previous",
   "footer.next": "Next",
+  "header.title": "Header",
   "meta.comments": "Comments",
   "meta.source": "Source",
   "search.config.lang": "en",
@@ -39,8 +41,11 @@
   "search.result.none": "No matching documents",
   "search.result.one": "1 matching document",
   "search.result.other": "# matching documents",
+  "search.reset": "Reset",
   "skip.link.title": "Skip to content",
   "source.link.title": "Go to repository",
   "source.revision.date": "Last update",
-  "toc.title": "Table of contents"
+  "tabs.title": "Tabs",
+  "toc.title": "Table of contents",
+  "nav.title": "Navigation"
 }[key] }}{% endmacro %}
--- a/src/partials/language/es.html
+++ b/src/partials/language/es.html
@@ -26,8 +26,10 @@
   "clipboard.copy": "Copiar al portapapeles",
   "clipboard.copied": "Copiado al portapapeles",
   "edit.link.title": "Editar esta página",
+  "footer.title": "Pie",
   "footer.previous": "Anterior",
   "footer.next": "Siguiente",
+  "header.title": "Cabecera",
   "meta.comments": "Comentarios",
   "meta.source": "Fuente",
   "search.config.lang": "es",
@@ -36,8 +38,11 @@
   "search.result.none": "No se encontraron documentos",
   "search.result.one": "1 documento encontrado",
   "search.result.other": "# documentos encontrados",
+  "search.reset": "Restablecer",
   "skip.link.title": "Saltar a contenido",
   "source.link.title": "Ir al repositorio",
   "source.revision.date": "Última actualización",
-  "toc.title": "Tabla de contenidos"
+  "tabs.title": "Pestañas",
+  "toc.title": "Tabla de contenidos",
+  "nav.title": "Navegación"
 }[key] }}{% endmacro %}
--- a/src/partials/search.html
+++ b/src/partials/search.html
@@ -31,7 +31,6 @@
         type="text"
         class="md-search__input"
         name="query"
-        aria-label="search"
         placeholder="{{ lang.t('search.placeholder') }}"
         autocapitalize="off"
         autocorrect="off"
@@ -48,7 +47,7 @@
       <button
         type="reset"
         class="md-search__icon md-icon"
-        aria-label="reset"
+        aria-label="{{ lang.t('search.reset') }}"
         data-md-component="search-reset"
         tabindex="-1"
       >
--- a/src/partials/toc.html
+++ b/src/partials/toc.html
@@ -23,7 +23,7 @@
 {% import "partials/language.html" as lang with context %}
 
 <!-- Table of contents -->
-<nav class="md-nav md-nav--secondary">
+<nav class="md-nav md-nav--secondary" aria-label="{{ lang.t('toc.title') }}">
   {% set toc = page.toc %}
 
   <!--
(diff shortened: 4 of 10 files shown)


src/partials/toc.html
```diff
@@ -23,7 +23,7 @@
 {% import "partials/language.html" as lang with context %}
 
 <!-- Table of contents -->
-<nav class="md-nav md-nav--secondary">
+<nav class="md-nav md-nav--secondary" aria-label="Table of contents">
```
squidfunk: Please use the translation `"toc.title"` here. See the base template for how to use translations.

squidfunk:
Hmm, only adding English titles is not that great actually. How are the `aria-label` titles used? Do screen readers read them aloud?

jaimeiniesta:
> Hmm, only adding English titles is not that great actually. How are the `aria-label` titles used? Do screen readers read them aloud?

I think that's the idea yes, these labels can be read aloud or provided as a structured list to aid with navigation maps. So if there were a single `nav` in the page there is no problem, but if there are many, each of them needs to be uniquely labelled so they can be properly described. Instead of "navigation, navigation, navigation", the user will hear "navigation: main, navigation: footer, etc.".

So yes these should be translated.

squidfunk:
okay then, could you add the translations to `src/partials/language/en.html` then? This is the fallback for when no other translations are given.

We can call the keys:

- `header.title`
- `footer.title`
- `nav.title`
- `tabs.title`

jaimeiniesta:
Done! I've translated all the `aria-label`s - the new ones and an existing one for the search placeholder.

I've also included Spanish translations for the new keys.

src/partials/language/en.html
```diff
@@ -27,8 +27,10 @@
   "clipboard.copy": "Copy to clipboard",
   "clipboard.copied": "Copied to clipboard",
   "edit.link.title": "Edit this page",
+  "footer.title": "Footer",
```
squidfunk: Minor issue: please ensure that keys are sorted alphabetically.

src/partials/language/es.html
```diff
@@ -26,8 +26,10 @@
   "clipboard.copy": "Copiar al portapapeles",
   "clipboard.copied": "Copiado al portapapeles",
   "edit.link.title": "Editar esta página",
+  "footer.title": "Pie",
```
squidfunk: Same issue (alphabetical ordering)

src/partials/search.html
```diff
@@ -27,7 +27,7 @@
   <label class="md-search__overlay" for="__search"></label>
   <div class="md-search__inner" role="search">
     <form class="md-search__form" name="search">
-      <input type="text" class="md-search__input" aria-label="search" name="query"
+      <input type="text" class="md-search__input" aria-label="{{ lang.t('search.placeholder') }}" name="query"
```
squidfunk: Why do we need an `aria-label` when there's already a placeholder set?

squidfunk: Ah, sorry, haven't seen that it was already there. Somebody added it at some point. Can we remove this?

squidfunk:
I just did a search and there're more `aria-label` entries, like for example `reset` for the search reset button. We should also provide translations for that. So to sum up, I think we should do the following things:

- PR to refactoring branch
- Order labels alphabetically
- Check if we need `aria-label` if there's a placeholder set
- Add translation for the search reset button

src/partials/search.html
jaimeiniesta: I don't think we need it, I was just using the translation here (this `aria-label` was there before I came). 😄 

I'll remove that one.

jaimeiniesta:
> * PR to refactoring branch

Sorry what branch do you mean?

squidfunk:
This one:
https://github.com/squidfunk/mkdocs-material/tree/refactor/rxjs-typescript

This is where v5 currently lives.

jaimeiniesta:
Alright, I need to fix those conflicts, I'll have it ready today.

src/partials/language/en.html
jaimeiniesta: Alright, I'll fix the rest of the keys also as they were not in order.

squidfunk: Ah wait. Okay, alphabetical order makes no sense, as `language` and `directionality` should be at the top. Nevermind, then.

jaimeiniesta:
Sorry, I cannot get `npm run watch` to run on the new branch, what's the command now to run the dev server?

jaimeiniesta:
I'm following https://deploy-preview-1486--mkdocs-material-preview.netlify.com/customization/#development-mode to run my development server but I can't get it to work - when I run `npm start` I see webpack compiling the assets but `mkdocs serve` doesn't seem to be started, as localhost:8000 is not accessible.

On the `master` branch, running `npm run watch` runs fine.

squidfunk:
You must start `mkdocs serve` in a separate terminal session. `npm start` will only start the Webpack watchdog.

Stanzilla:
> EDIT: btw, MkDocs is not automatically started anymore because the dev server of MkDocs is..... IMHO rather bad. When there's a fresh Webpack build, it will queue up as many refreshes as there are new files, which is 5k since we added the icons to the repositories and then refresh the window every 5 seconds. Absolutely impractical. For this reason I've transitioned to starting the server after the initial build manually.

Oh so that is what happens there. Yeah it's super annoying.

jaimeiniesta:
Thanks for the explanation. I can update the readme with these instructions on this PR as well.

squidfunk:
Sure, we should just add a paragraph to the customization guide saying that `mkdocs serve` must be started in a second session.

Stanzilla:
Can we not && chain it?

jaimeiniesta:
> Can we not && chain it?

Don't think so, as the first command doesn't end until you stop it, so the second can't start.

squidfunk:
Exactly. I'm still working for a better approach to completely ditch the MkDocs development server, but didn't find something that works reasonably well yet.

jaimeiniesta:
OK I've fixed and force-pushed my branch, including the changes requested. Can you take another look please?

jaimeiniesta:
The netlify preview doesn't seem to be taking the latest changes in the branch, maybe because of the force-push?

squidfunk:
Looks good. Merging straight away. The only thing that I feel can be improved is the translation for "reset". I think we should call it "Clear". Is "Limpiar" the correct translation in Spanish?

squidfunk:
> The netlify preview doesn't seem to be taking the latest changes in the branch, maybe because of the force-push?

This is because you did not build the theme. However, nevermind, I'll handle it.

jaimeiniesta:
> Looks good. Merging straight away. The only thing that I feel can be improved is the translation for "reset". I think we should call it "Clear". Is "Limpiar" the correct translation in Spanish?

Yes, Clear and Limpiar are OK.

Thanks! 🎉

squidfunk:
Funny side note: as my Spanish is very basic, I associate "limpiar" with the lady at the market asking me whether I want her to remove the guts of the squids I'm buying 😅

squidfunk:
The deploy preview was updated

squidfunk:
Oh and thanks for your help! Glad we could further improve accessibility

PR #1517, opened 2020-03-18
Bring back aria-label to search form

jaimeiniesta:
As a continuation of https://github.com/squidfunk/mkdocs-material/pull/1516 this PR:

* Bring back aria-label to search form as form elements must have labels, a placeholder is not enough.
* Updates Development server guide.

Related:

https://dequeuniversity.com/rules/axe/3.5/label?application=rocketvalidator

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -201,15 +201,19 @@ npm install
 
 ### Development mode
 
-The development server can be started with:
+Start the Webpack watchdog with:
 
 ``` sh
 npm start
 ```
 
-This will also start the MkDocs development server which will monitor changes
-on assets, templates and documentation. Point your browser to
-[localhost:8000][190] and you should see this documentation in front of you.
+Then, in a second session, start the MkDocs server with:
+
+```sh
+mkdocs serve
+```
+
+Point your browser to [localhost:8000][9] and you should see this documentation in front of you.
 
 !!! warning "Automatically generated files"
 
--- a/src/partials/search.html
+++ b/src/partials/search.html
@@ -31,6 +31,7 @@
         type="text"
         class="md-search__input"
         name="query"
+        aria-label="{{ lang.t('search.placeholder') }}"
         placeholder="{{ lang.t('search.placeholder') }}"
         autocapitalize="off"
         autocorrect="off"


squidfunk:
Huh, I thought [we just discussed that](https://github.com/squidfunk/mkdocs-material/pull/1516#discussion_r394334302)?

jaimeiniesta:
> Huh, I thought [we just discussed that](https://github.com/squidfunk/mkdocs-material/pull/1516#discussion_r394334302)?

Yes sorry, but my head was on the "landmarks must be unique" issue then. So removing this one didn't make it worse, but a separate issue appeared that I had not thought of -- labels are indeed needed for form inputs, a placeholder is not enough.

jaimeiniesta:
So, we're expected to run `npm run build` and commit the changes in the `material` folder when we make a PR?

I find that weird, should netlify run this on their instances instead?

squidfunk:
> Yes sorry, but my head was on the "landmarks must be unique" issue then. So removing this one didn't make it worse, but a separate issue appeared that I had not thought of -- labels are indeed needed for form inputs, a placeholder is not enough.

Okay, no problem.

squidfunk:
> So, we're expected to run npm run build and commit the changes in the material folder when we make a PR?
> 
> I find that weird, should netlify run this on their instances instead?

It's the best I could come up with. The thing is that the theme should always be installable from `git` which means that the distribution files which are all in the `material` folder must be checked in. We could build it in CI, but then CI must create another commit to add the distribution files which we must detect not to trigger a cycle. For this reason, manually building and checking it in is what we did for the last four years. We could build it during the netlify depoyment, but we wont detect whether the files that are checked in are actually those that are shown which is not a good idea.

jaimeiniesta:
OK, thanks for the explanation, makes sense.

I've included the build in this PR.

squidfunk:
Could you rebase your PR so I can merge it? Probably the easiest is to pull the latest changes from upstream and rebuild it.

jaimeiniesta:
> Could you rebase your PR so I can merge it? Probably the easiest is to pull the latest changes from upstream and rebuild it.

Sure, this time I've not included the build to minimize conflicts.

squidfunk:
Added build files in 5f2f497f, should be shortly available in deploy preview.

PR #1538, opened 2020-03-23
Fix: minify inline JS

RDIL:
The minify plugin can't handle inline JS so minify it here.

Signed-off-by: Reece Dunham <me@rdil.rocks>


squidfunk:
This part is actually still a work in progress, so it doesn't make sense to do it at this moment.

RDIL:
Okay.

PR #1543, opened 2020-03-25
l10n: zh-TW localization update

visig9:


src/partials/language/zh-TW.html
```diff
@@ -31,13 +32,16 @@
   "meta.comments": "留言",
   "meta.source": "來源",
   "search.language": "ja",
+  "search.pipeline.stopwords": false,
```
squidfunk: Could you explain why you added those?

src/partials/language/zh-TW.html
```diff
@@ -23,6 +23,7 @@
 <!-- Translations: Chinese (Taiwanese) -->
 {% macro t(key) %}{{ {
   "language": "zh-Hant",
+  "direction": "ltr",
```
squidfunk: This can be removed, `ltr` is set by default.

src/partials/language/zh-TW.html
visig9: Nice catch! It my fault. I re-check the code base and this line should be deleted in chinese context, and also `trimmer` line.

Long story short, I found `zh-TW.html` missing the `source.revision.date` and cause `git-revision-date-localized` plugin not work correctly (no default string in `5.0.0rc3`). So I think should not leave any fields missing in this file and try to copy some line from other file. And when I test, the search result still not too bad. :P

Anyway this should be clear out. Thanks so much. I will update this PR ASAP.

squidfunk:
LGTM!

PR #1570, opened 2020-04-07
Add Manticore Games to users

Stanzilla:


squidfunk:
As discussed on Gitter, this section is reserved for very big companies using Material for MkDocs. Unfortunately, we cannot add every company.

PR #1571, opened 2020-04-07
Fix example of plugins.search.separator

peaceiris:
Fixed the example of `plugins.search.separator` on the v5 release note.


squidfunk:
Ah, that somehow got lost in the merge. Thanks!

PR #1599, opened 2020-04-14
Add LifeOmic

hemp:
Would like to show our adoption off.  No worries if not!


squidfunk:
Hey, that is super nice, but this space is actually reserved for super large companies 😉

PR #1601, opened 2020-04-14
Async load fonts from Google

Stanzilla:
To not block rendering, we can load in Google Fonts in asynchronous way by using [Google Web Font Loader](https://developers.google.com/fonts/docs/webfont_loader)

Ref: https://www.denisbouquet.com/google-fonts-render-blocking/

### DIFF
--- a/src/base.html
+++ b/src/base.html
@@ -119,19 +119,15 @@
       <!-- Load fonts from Google -->
       {% if font != false %}
         <link href="https://fonts.gstatic.com" rel="preconnect" crossorigin />
-        <script type="text/javascript">
-          WebFontConfig = {
-            google: { families: [ '{{ font.text }}:300,400,400i,700|{{ font.code }}' ] }
-          };
-          (function() {
-            var wf = document.createElement('script');
-            wf.src = 'https://ajax.googleapis.com/ajax/libs/webfont/1/webfont.js';
-            wf.type = 'text/javascript';
-            wf.async = 'true';
-            var s = document.getElementsByTagName('script')[0];
-            s.parentNode.insertBefore(wf, s);
-          })();
-        </script>
+        <link
+          rel="preload"
+          as="style"
+          onload="this.rel = 'stylesheet'"
+          href="https://fonts.googleapis.com/css?family={{
+            font.text | replace(' ', '+') + ':300,400,400i,700%7C' +
+            font.code | replace(' ', '+')
+          }}&display=fallback"
+        />
         <style>
           body, input {
             font-family: "{{ font.text }}",
--- a/src/base.html
+++ b/src/base.html
@@ -126,7 +126,7 @@
           href="https://fonts.googleapis.com/css?family={{
             font.text | replace(' ', '+') + ':300,400,400i,700%7C' +
             font.code | replace(' ', '+')
-          }}&display=fallback"
+          }}&display=swap"
         />
         <style>
           body, input {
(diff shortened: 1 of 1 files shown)


squidfunk:
AFAIK, it doesn't block rendering due to `font-display swap`.

Stanzilla:
Not quite, sadly. Also see https://johnfraney.ca/posts/2019/12/09/improve-page-speed-google-fonts/

Stanzilla:
We could also use something like this

```js
 <link rel="preload" as="style" crossorigin
    href="https://fonts.googleapis.com/css?display=swap&family=Gentium+Basic:400,400i,700|Limelight"
    onload="
    performance.measure('link-onload', 'start-stylesheet');
    performance.mark('apply-stylesheet');
    this.onload = () => {
      performance.measure('stylesheet-onload', 'apply-stylesheet');
    };
    this.rel = 'stylesheet'
  ">
```

Stanzilla:
changed it to that, also note that we are using fallback and not swap here

squidfunk:
I see absolutely no difference when using Slow 3G.

Stanzilla:
For me it fixes the render blocking resources issue in Lighthouse completely but font swap is visible on the first page load

squidfunk:
Could you provide a side by side comparison, i.e. some data? Also, what about cross browser support?

Stanzilla:
After

image

Before

image

squidfunk:
That’s potential savings, which is not accurate. Lighthouse is really no good primary source of truth. What I meant was real data.

Stanzilla:
Well if potential savings are gone after, that would mean they are saved now?

squidfunk:
We need real, measured data and data on cross browser support for the preload approach. This would be a very substantial change.

XhmikosR:
For what is worth I don't see this issue here.

Stanzilla:
@XhmikosR the lighthouse complaint or the font popping?

XhmikosR:
The lighthouse complaint.

* https://googlechrome.github.io/lighthouse/viewer/?psiurl=https%3A%2F%2Fsquidfunk.github.io%2Fmkdocs-material%2F&strategy=mobile&category=performance&category=accessibility&category=best-practices
* https://lighthouse-dot-webdotdevsite.appspot.com//lh/html?url=https%3A%2F%2Fsquidfunk.github.io%2Fmkdocs-material%2F

Stanzilla:
both show them as potential saving but as a passed audit, like it did with my site

XhmikosR:
Then I'd say just ignore it. font-display is the way to go here.

squidfunk:
Please, [don't rely on Lighthouse](https://www.matuzo.at/blog/building-the-most-inaccessible-site-possible-with-a-perfect-lighthouse-score/)! It makes no sense to consider this PR for merging unless there's evidence that this significantly improves the UX. Also, cross-browser support has to be researched.

Stanzilla:
Alright so, should we then maybe reconsider using `swap` instead of `fallback` and ditch the other changes?

squidfunk:
See #1133 for a discussion where we changed `swap` for `fallback`.

Stanzilla:
That's a nice back and forth

squidfunk:
That's the point. I have no interest in changing it back and forth 😅

squidfunk:
I've decided to close this PR. It went off rails and I see no upside to changing `fallback` back to `swap`. We don't have any data, except Lighthouse, which is unreliable. I would call it premature optimization because we don't have any data which shows a significant improvement. This is an opinionated project. It's an HTML template after all. If somebody feels like changing `fallback` to `swap`, you're always free to fork.

Stanzilla:
fine with me! :)

PR #1606, opened 2020-04-16
Fix for GitLab stars

NickBusey:
The current regex grabs the trailing / on the repo name like `squidfunk/mkdocs-material/` but GitLab doesn't expect that and will 404 with the trailing /.

This new regex does not grab the trailing slash, and should fix GitLab stars not rendering.

### DIFF
--- a/src/assets/javascripts/patches/source/index.ts
+++ b/src/assets/javascripts/patches/source/index.ts
@@ -74,7 +74,7 @@ function fetchSourceFacts(
 
     /* GitLab repository */
     case "gitlab":
-      const [, base, project] = url.match(/^.+?([^\/]*gitlab[^\/]+)\/(.+)/i)
+      const [, base, project] = url.match(/^.+?([^\/]*gitlab[^\/]+)\/([A-Za-z0-9\-\_]+\/[A-Za-z0-9\-\_]+)/i)
       return fetchSourceFactsFromGitLab(base, project)
 
     /* Everything else */


squidfunk:
Thanks for the PR! From what I've seen, GitLab projects can be nested on deeper levels, or not? This Regex expects two levels.

NickBusey:
I'm not aware of that. For organizations, it is still two levels.

Really a better fix would be to just have the gitlab project id be included in the config and use that directly, much more reliable, but does add an extra setup step.

src/assets/javascripts/patches/source/index.ts
```diff
@@ -74,7 +74,7 @@ function fetchSourceFacts(
 
     /* GitLab repository */
     case "gitlab":
-      const [, base, project] = url.match(/^.+?([^\/]*gitlab[^\/]+)\/(.+)/i)
+      const [, base, project] = url.match(/^.+?([^\/]*gitlab[^\/]+)\/([A-Za-z0-9\-\_]+\/[A-Za-z0-9\-\_]+)/i)
```
XhmikosR, line 77: Any reason you don't just use `\w`?

XhmikosR, line 77:
```suggestion
      const [, base, project] = url.match(/^.+?([^/]*gitlab[^/]+)\/([\w-]+\/[\w-]+)/i)
```

squidfunk:
Fixed in 387970c6.

PR #1612, opened 2020-04-17
Fix npm lint scripts

XhmikosR:
* Double quote globs since single quotes don't work on all shells
* Drop `find` and just use a glob pattern; `find` is a nix utility while this works everywhere

Tested this on my Windows dev VM and now it works fine.


squidfunk:
Thanks!

PR #1613, opened 2020-04-17
Tweak CI config

XhmikosR:
* add the `CI` environment variable; removes npm postinstall messages
* move Node.js and Python versions to environment variables
* add the Node.js and Python versions in cache filenames

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,11 @@ on:
   - push
   - pull_request
 
+env:
+  CI: true
+  NODE_VERSION: 10.x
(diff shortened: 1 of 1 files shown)


squidfunk:
LGTM.

XhmikosR:
One thing I notice @squidfunk is that the deploy action runs for [forks](https://github.com/XhmikosR/mkdocs-material/commits/gh-pages) too. I will make a PR to change this so that it only runs for your repo.

squidfunk:
Consider that some forks may want this and/or rely on this.

XhmikosR:
OK, let's leave it as is then.

PR #1617, opened 2020-04-17
Losslessly compress PNG files

XhmikosR:
Before: 538 KB (551.329 bytes)
After:  487 KB (499.630 bytes)


XhmikosR:
@squidfunk not bad :)

Screenshot_2020-04-17 Compare Reports squidfunk github io, squidfunk github io GTmetrix

BTW I was wondering if we could add dns-prefetch for `api.github.com`? Also, personally, I'd only use `dns-prefetch` for `https://www.google-analytics.com/`. `preconnect` slows down the initial HTML load time so should be used sparsely.

Another thing would be to move loading of search.json when the user clicks on the input, but it might not be so easy (should probably have a loaded indicator and so on).

squidfunk:
> BTW I was wondering if we could add dns-prefetch for api.github.com? Also, personally, I'd only use dns-prefetch for https://www.google-analytics.com/. preconnect slows down the initial HTML load time so should be used sparsely.

Not everyone uses GitHub, so we would need to implement some template logic to detect whether this is a git repository or not, which is something I'd not want to maintain for a very small upside, as the star count is cached anyway for subsequent calls. You could, however, add the prefetch instruction yourself using theme extension.

> Another thing would be to move loading of search.json when the user clicks on the input, but it might not be so easy (should probably have a loaded indicator and so on).

That's actually something I thought about too, which could be provided via an option on template level (like the search transform). There's already a ticket tracking the loading indicator which is related. Might be added in the future, but I cannot put an estimate on it.

XhmikosR:
>Not everyone uses GitHub, so we would need to implement some template logic to detect whether this is a GitHub repository or not (or GitLab or whatever), which is something I'd not want to maintain for a very small upside, as the star count is cached anyway for subsequent calls. You could, however, add the prefetch instruction yourself using theme extension.

Makes sense. I will add it in our template and see if it improves things.

PR #1619, opened 2020-04-17
Fix splitChunks on Windows

XhmikosR:
https://webpack.js.org/plugins/split-chunks-plugin/#optimizationsplitchunks

>When files paths are processed by webpack, they always contain / on Unix systems and \ on Windows. That's why using [\\/] in {cacheGroup}.test fields is necessary to represent a path separator. / or \ in {cacheGroup}.test will cause issues when used cross-platform.

And here's the patch from my Windows dev VM: https://github.com/squidfunk/mkdocs-material/compare/master...XhmikosR:build

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -333,7 +333,7 @@ export default (_env: never, args: Configuration): Configuration[] => {
         splitChunks: {
           cacheGroups: {
             vendor: {
-              test: /\/node_modules\//,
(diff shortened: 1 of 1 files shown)


squidfunk:
Doesn't the backslash need to be escaped as well?

XhmikosR:
I don't think so: https://regex101.com/r/Tkq9Se/1

It's what they also have in their official docs

PR #1585, opened 2020-04-11
Feature/how to GitHub

Andre601:
Implements the Tutorial about how to use GitHub Pages with MkDocs and MkDocs-Material.

I created a tutorials folder for the other planned tutorials (#1573) and integrated the github-pages,md file as follows to the nav:
```yaml
nav:
  # Other paths up to Plugins
  - Tutorials:
    - GitHub Pages: tutorials/github-pages.md
  # Releases and other paths.
```

Issue #1498, opened 2020-03-10
🚀 Material for MkDocs 5.0.0 RC 4


squidfunk:
__Help test Material 5 RC 4! [Deploy preview](https://deploy-preview-1486--mkdocs-material-preview.netlify.com/)__

Please post any problems you encounter during migration in this issue.

## Installation

Using `pip`:

```
pip install "mkdocs-material>=5.0.0rc4"
```

Using `docker`:

```
docker pull squidfunk/mkdocs-material:5.0.0rc4
```

## Features

- [x] Reactive architecture – try `__material.dialog$.next("Hi!")` in the console!
- [x] Instant loading – make Material behave like a Single Page Application!
- [x] Improved CSS customization with [CSS variables](https://github.com/squidfunk/mkdocs-material/blob/081d540127c41703da3d73aa69eb521615c1cb89/src/assets/stylesheets/base/_colors.scss#L27-L55) – define your CI colors!
- [x] Improved CSS resilience, e.g. proper sidebar locking for customized headers
- [x] Improved icon integration and configuration – now including over 5k icons!
- [x] Added possibility to use any icon for logo, repository and social links
- [x] Search UI does not freeze anymore (moved to web worker)
- [x] Search index built only once when using instant loading
- [x] Improved extensible keyboard handling
- [x] Support for [prebuilt search indexes](https://www.mkdocs.org/user-guide/configuration/#prebuild_index)
- [x] Support for displaying stars and forks for GitLab repositories
- [x] Support for scroll snapping of sidebars and search results
- [x] Reduced HTML and CSS footprint due to deprecation of Internet Explorer support
- [x] Slight facelifting of some UI elements (Admonitions, tables, ...)

## Fixed in RC 2

- [x] #1505: Search does not close – c0abaf5
- [x] #1503: Exception when setting only logo icon – fcbd47c
- [x] #1502: Tabbed content alignment – 4d386df, 447d409
- [x] #1499, #1501: Code block styling issues with `pymdownx.inline` and `inline`

## Fixed in RC 3

- [x] #1515: Return link of footnote not show after convert to pdf enhancement
- [x] Fixed Admonitions for print media
- [x] Fixed Details for Safari (iOS and macOS) 
- [x] Fixed hover states for nested items in mobile navigation
- [x] Improved rendering performance
- [x] Improved social icons and copyright notice alignment
- [x] Improved accessibility

## Fixed in RC 4

[...]

PR #1620, opened 2020-04-17
docs/add icon set

diba1013:
# Description

This describes the steps to perform to add the your own `.svg` icons that can be referenced within `mkdocs.yml` as `theme.icon`.

A rendered version can be found [here](https://diba1013.github.io/mkdocs-material-playground/customization/icons/) (small changes done within this PR, but gives an overall overview) 

## State of this PR

I have built the documentation and verified the rendered result. As per discussion on #1585, I specified `docs/tutorials` as base branch (which is slightly outdated, thus this PR contains a half broken link to `../extensions/pymdown.md#icons`).

### Questions

1. How do you want this to be setup? For now I have gone for a style: "All necessary steps are already described in the main documentation, so this will only guide you through without explaining them in detail". Thus I have only linked to the more detailed references.
2. How do you do your links? For now I have used inline-style relative markdown links.

### mkdocs-material-extensions

I have tested this with [mkdocs-material-extensions](https://github.com/facelessuser/mkdocs-material-extensions) (since it is now referenced on the production site) and it does not pick up on your custom icon sets (a setup can be found [here](https://github.com/diba1013/mkdocs-material-playground/blob/master/mkdocs.yml), note the `:bootstrap-circle-square:` within the bug notice). @facelessuser I know [this was not intended in the first place](https://github.com/squidfunk/mkdocs-material/issues/1498#issuecomment-597678171) and I am happy to file an issue for that. I understand that it is a [pre-release](https://github.com/facelessuser/mkdocs-material-extensions/issues/2) and will be shipped with mkdocs-material eventually.

If this is intentional or not possible, I am happy to remove this notice.

### DIFF
--- a/docs/tutorials/icons.md
+++ b/docs/tutorials/icons.md
@@ -1,0 +1,71 @@
+# Icon Sets
+
+With Material 5.0, it is now possible to customize the icons used in the documentation with `.svg` icons. It ships three icon sets, providing over 5k icons:
+
+1. [Material Design icons](https://material.io/resources/icons/)
+2. [FontAwesome icons](https://fontawesome.com/icons?d=gallery&m=free)
+3. [GitHub's Octicons](https://octicons.github.com/)
+
+However, you can add your own custom icons or third party icon sets to your documentation.
+
+!!! tip
+    Because icons are directly inlined into the html, they must be in the `.svg` vector format. If you have a bitmap (e.g. `.png`), you will need to [trace](https://inkscape.org/doc/tutorials/tracing/tutorial-tracing.html) them using [inkscape](https://inkscape.org) or similar tools so that they can be converted to a `.svg` format.
+
+## Goal
+
+With this tutorial, we will add [Bootstrap Icons][bootstrap-icons], a ready-to-use icon set. These icons will be used to customize the icons available in within `mkdocs.yml`, namely `theme.icon` and the social icons `extra.social.icon`.
+
+## Setup
+
+All icons provided by Material are placed within the `.icons` folder. Each icon set provided is placed within its own sub-folder to avoid naming conflicts for the same icon name. They are accessible using the relative file name (without extension): `fontawesome/brands/github`.
+
+In order to add the Bootstrap Icons, we will need to extend the theme to gain access to the `.icons` folder. For that, we create an folder `theme` (or any other name) within the project root and specify it as `custom_dir`.
+
+```yaml
+theme:
+    name: material
+    custom_dir: theme
+```
+
+!!! warning
+    Theme extension may be used for the following use cases:
+
+    1. Overwrite assets locally provided by Material by specifying the exact relative file name (e.g. `.icons/logo.svg`). 
+    2. Add custom assets to be discoverable by Material.
+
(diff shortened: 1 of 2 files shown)


facelessuser:
@diba1013 I will need you to explain the "bug". Are you referring to using icon syntax `:myicon:` in the yml? If so, that syntax only works in Markdown content as intended.

facelessuser:
I would remove the `bug` note as this is not currently a bug, but an enhancement. It is a beta, and as such, has only implemented official icon support. This was known at the time of the beta's creation.

facelessuser:
It may be reasonable to have a informative note, but I would not advertise it as a bug.

diba1013:
> It may be reasonable to have an informative note, but I would not advertise it as a bug.

Thanks for the feedback. I changed this to an information notice and improved the wording.  Please feel free to improve upon it as you see fit.

docs/tutorials/icons.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+Simply add your `.svg` icons within `${custom_dir}/.icons/${namespace}/`. The `${namespace}` can be empty, but supports any arbitrary folder structure (similar to `fontawesome/brands`).
+
+!!! tip
+    You can overwrite existing icon sets by choosing the same namespace and placing a `.svg` with the same name.
+
+## Usage
+
+You can use the icon within `mkdocs.yml`. For the example below, I assume the namespace `bootstrap` from [bootstrap-icons](https://icons.getbootstrap.com/).
```
squidfunk, line 26: Please don't use first person singular in the documentation.

squidfunk:
Thanks for the writeup! Regarding your question to which degree the tutorial should describe stuff that is already outlined in other parts of the documentation I'd recommend to maybe go in some more detail. A tutorial/howto should be understandable without having to navigate around a lot, and it should be beginner-friendly.

Furthermore, I would think a concrete example is much better than using variables in the text (like `${custom_dir}` which is rather cryptic). Interestingly, many users of this project are technical writers, not developers. Also, as I can read from the results of the survey (which I will release some time soon), people have trouble customizing, as they seem to lack concrete examples. Could we take the Bootstrap icons (which I find to be an excellent use case, as it's a ready-to-use icon set) as an example from the start?

diba1013:
I have rewritten the tutorial from the ground up to be more detailed and to remove the navigation links.

For now, I have removed the integration part with mkdocs-material-extensions. Although it is now possible (thanks to the fast and awesome work of @facelessuser), it [requires a bit more setup right now](https://github.com/diba1013/mkdocs-material-playground/compare/feature/icons-in-markdown) (mostly upgrading to the latest version and adding a stylesheet). I will prepare a corresponding PR (once this is merged) where we can discuss the necessary adaptions.

facelessuser:
@diba1013, stylesheet should not be necessary. You're own custom icons should be styled the same as  all the other icons that material includes...unless your SVG have additional things that must be overridden.

diba1013:
> stylesheet should not be necessary

Yes, I think I was still using `mkdocs-material==5.0.0` which does not style the icons properly (i.e. original size). I will confirm this prior to creating the PR.

squidfunk:
Great! I'll rebase and merge it into the docs branch and we can improve from there on 👍

PR #1621, opened 2020-04-17
Remove unused imports

XhmikosR:


squidfunk:
They are not unused. They are necessary for JSX to work.

squidfunk:
FYI

https://github.com/squidfunk/mkdocs-material/blob/5b8721ff9286a5303219d39f7bf9efc8ca6b191a/tsconfig.json#L9

PR #1626, opened 2020-04-18
fix typos

ofek:


squidfunk:
Thanks! I feel the language can also be improved a little but we can do that when we have some tutorials as a finishing touch.

PR #1632, opened 2020-04-22
Add namespace for icons

Andre601:
I added the actually used name for the icons to the list of supported icons (i.e. Fontawesome (fontawesome): 1.5k icons)

This should prevent future confusion, where it isn't 100% clear, which "prefix" the icon of your choice has. Trust me... this will happen.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -527,9 +527,9 @@ theme:
 All icons are directly inlined as `*.svg` files, so no further requests will be
 made. Icon sets which are bundled with Material for MkDocs:
 
-* [Material Design icons][16]: 3.8k icons
-* [FontAwesome icons][17]: 1.5k icons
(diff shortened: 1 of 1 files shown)


squidfunk:
Please enclose the namespaces in backticks, so they render as code blocks.

Andre601:
Like this?

squidfunk:
Yes, thanks! I don't think it's the optimal solution, but we can improve it later. Ideally, we'd have an auto-generated index for all icons, but putting 5k icons on one page will probably make the docs explode 💥

Andre601:
The point here was not to *somehow* list all available icons.... That would be more the job of the various sites from where they are.

It was just to mention the names used for all the places (in the mkdocs.yml) where you would have `<namespace>/(<path>/)<icon>`, so that people won't ask "What do I use for Octicons or material?"

PR #1639, opened 2020-04-25
WIP – Dark mode

squidfunk:
[screenshot-deploy-preview-1639-mkdocs-material-preview-netlify-app-getting--1587834905922](https://deploy-preview-1639--mkdocs-material-preview.netlify.app/getting-started/)


ofek:
Bottom is light btw

Capture

facelessuser:
Yeah, I don't think he's gone through and adjusted all the colors yet. I imagine he'll update the syntax highlighter, footers, and other things. I think this is just the start.

squidfunk:
Still WIP. Don’t worry, footer will be dark as well.

bilal-fazlani:
Will the dark mode have to be configured in config file or users will be able to switch it on and off according to their comfort?

squidfunk:
You can set it via `mkdocs.yml`, it will probably be under `theme`. Adding a user toggle is super easy (as can be seen in the deploy preview), so I’ll wait until a good approach for adding a general toggle has been found.

ofek:
So it'll be released with a toggle right?

facelessuser:
I think what he was saying is that the act of toggling is easy, but it will not have a toggle until a good toggle (where we put it) can be worked out.

squidfunk:
Exactly. Sorry if that was not clear enough. A toggle implementation will have to change a single attribute on an element, so it’ll be like 2-3 lines of JS. I’m unsure where to put the toggle, that’s my main concern.

ofek:
Can we bind it to a shortcut key too? 😄 In that case we could use it right away as well until toggle

squidfunk:
There won't be a toggle at first. The reason for this is that I'm abstracting dark mode into something more generic called color schemes. This will be another option of the `palette` attribute in `mkdocs.yml` and will allow for the addition of further schemes besides a dark theme like a high contrast theme for people with impaired vision.

This, it's not a binary/boolean option, which makes a toggle impractical. However, later, when some people implemented good toggle solutions, which allow for flexible configuration and switching of schemes, I'm happy to integrate it.

squidfunk:
I think we have a good candidate, which should be quite ready for release now. The dark mode color scheme is called `slate`. Light mode is just the default scheme but will be called `light` for convenience. Some more variables were added:

https://github.com/squidfunk/mkdocs-material/blob/7a12b5ec102071d1917ef83ee5cd62f2a36991a1/src/assets/stylesheets/base/_colors.scss#L64-L73

Thus, admonitions now have a slightly lighter background color for the `slate` theme, as the shadow itself doesn't give enough contrast. Universally increasing the shadow doesn't look good for the `light` color scheme. The footer now has distinct variables for styling.

__Note, that there will always be color combinations that' don't work right out of the box, like for example the `light` scheme with `yellow` as a primary color__. However, it works very well in dark mode, which is why we're going to leave it as-is. With the 21 primary colors, 16 accent colors, and the new `slate` color scheme, there're now  __21 ⨉ 16 ⨉ 2 = 672 combinations__ which can be achieved without any custom CSS. With custom CSS, possibilities are obviously endless. We might add some more variables for distinct styling if we run into customization problems, but for now, it should be enough.

facelessuser:
So does the toggle example you have only work (persistently) because you have instant enabled?

squidfunk:
No, it’s not related to instant loading at all, so no need to enable it.

facelessuser:
So if you switch to a different page with no instant, it'll remember whether you've toggled between light or dark? Forgive my ignorance.

squidfunk:
No, that's up to the author to implement. I may have not read "persistently" in your comment, it can currently only be set from `mkdocs.yml`. The toggles are only to show how they look. It's not an official implementation, yet, so no user-land switching without some JavaScript and localstorage 😀

facelessuser:
Cool, I don't need switching, but I was curious. Thanks!

Mario-Hofstaetter:
@squidfunk I wanted to try the new dark mode since I was awaiting it for weeks, but I can't get it to show.

`mkdocs.yml`:
```yml
site_name: My Docs
theme:
  name: 'material'
  palette:
    scheme: slate
```
Starting in a VM with:
`docker run --rm -it -p 8000:8000 -v ${PWD}:/docs squidfunk/mkdocs-material:5.2.0`

grafik

I tried clearing browser cache and a different browser, Windows System Theme is dark.

What did I do wrong?

hSaria:
I've had the same issue; specifying the primary and/or the accent color fixed it.

```
theme:
  name: material
  palette:
    scheme: slate
    primary: indigo
```

> Indigo is the default

squidfunk:
You may create an issue for that, that’s not intended.

Mario-Hofstaetter:
Thank you @hSaria that fixed it. Are you going to create the issue?

hSaria:
Sure, I'll create one

higker:
> I've had the same issue; specifying the primary and/or the accent color fixed it.
> 
> ```
> theme:
>   name: material
>   palette:
>     scheme: slate
>     primary: indigo
> ```
> 
> > Indigo is the default

thank ❤️

PR #1641, opened 2020-04-26
Fixed sidebar jitter by switching to sticky positioning

squidfunk:
This PR fixes the occasional jitter when scrolling fast across the point where the sidebars snap into place. The behavior can be seen in Chrome, Firefox, Safari, and all other browsers, and stems from the fact that the sidebar positioning is switched from `absolute` to `fixed` from within JavaScript. This behavior was implemented as part of 1.x and was necessary when `position: sticky` support was not good, but now, every major browser supports it. The header was already switched to `sticky` in 5.0, and as adoption is increasing, we can also switch the sidebars to `sticky`.

A further upside of this PR is that positioning of sidebars is now implemented with flexbox, dropping a lot of `margin` hacks, giving better control over what to show where.

__Before__:

before

__After__:

after


MikeyMJCO:
This looks great,

The preview tests ok on the current stable versions of Firefox, Edge Chromium and Google Chrome - and the Beta version of Edge Chromium (all on Windows 10 x64 2004 Preview(

squidfunk:
Thanks for giving it a spin! I'll wait for some more feedback and merge it to `master` then. There're some browsers with <1% browser share that don't support it, but it will degrade gracefully by just not making the sidebars sticky.

squidfunk:
I decided to merge this into `master`, as it works in all major browsers and I don't expect it to be a breaking change, at least for uncustomized installations.

PR #1656, opened 2020-04-27
added gitlab pages tutorial

ayeks:
Here is the tutorial on deploying mkdocs-material to GitLab pages with GitLab-CI.


squidfunk:
Thanks! I'm sorry that this took so long. I'll merge it into the docs branch and we improve from there :-)

PR #1660, opened 2020-04-29
Reorder nav pages

timvink:
In effect this changes nothing the website, but it better reflects the order of the pages that you see on the website:

Screenshot 2020-04-29 at 15 48 37

Context before you think I have OCD 😋: I'm developing a plugin where the ordering of pages in `nav` matters and was using this project's website as a test.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -127,6 +127,10 @@ markdown_extensions:
 nav:
   - Home: index.md
   - Getting started: getting-started.md
+  - Customization: customization.md
+  - Data privacy: data-privacy.md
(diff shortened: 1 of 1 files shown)


squidfunk:
IMHO this is actually counter-intuitive. Customization should come after all extension and plugin descriptions, as customization is a much deeper development knowledge than configuration. Also, the license and contributing stuff should be at the end.

However, I'm thinking about restructuring the stuff into a user and developer guide, but that's something for the future.

timvink:
Fair enough :)

PR #1662, opened 2020-04-30
Losslessly compress banner.png

XhmikosR:
213594 -> 197124 bytes


squidfunk:
I'm actually quite surprised, as I already used `pngquant` for compression. What do you use?

XhmikosR:
I use optipng, zopflipng and pngout :)

I should probably make a repo at some point with my scripts. It's nothing fancy, though.

Issue #1653, opened 2020-04-27
Instant loading doesn't work well when jumping to local .html file


Mi11et:
__I checked that...__

- [x] ... the documentation does not mention anything about my problem
- [x] ... the problem doesn't occur with the default MkDocs template
- [x] ... the problem is not in any of my customizations (CSS, JS, template)
- [x] ... there are no open or closed issues that are related to my problem

## Description

When clicking a link to another local .html file (which is not generated by MkDocs) with instant feature on, the page won't load correctly.

### Expected behavior

Instant loading should ignore that link.

### Actual behavior

The page doesn't load correctly.

### Steps to reproduce the bug

1. Create a new MkDocs project.
``` shell
mkdocs new test
```
2. Set theme to "material" and enable "instant" feature.
``` yaml
theme:
  name: material
  features:
    - instant
```
3. Create `test.html`.
``` html
<html>
    <body>
        <p>
            test
        </p>
    </body>
</html>
```
4. Write a link to that page in `index.md`.
``` markdown
[test](/test.html)
```
5. Run `mkdocs serve` and check that link.

### Package versions

* Python: `3.7.2`
* MkDocs: `1.1`
* Material: `5.1.0`

### Project configuration

``` yaml
site_name: Test

theme:
  name: material
  features:
    - instant
```

### System information

* OS: Windows 10
* Browser: I've tried both Firefox and Chrome

squidfunk:
Just a quick idea for later evaluation: when instant loading is executed and the response is received, we could check for the `generator` meta tag (or maybe others) and perform a full page load, if not found.

squidfunk:
Potential fix in #1668. Note that there are currently some limitations when serving the docs locally, but it should fully work when deployed. Before merging and releasing, we need to collect some feedback.

squidfunk:
No feedback, so I guess I’ll just merge the stuff whenever I find some time and we see if it breaks something after releasing it.

squidfunk:
Released as part of 5.1.6

PR #1668, opened 2020-05-03
Added loading of sitemap to filter instant loading requests

squidfunk:
This PR is a fix attempt for #1653. When instant loading is enabled, the current `master` will intercept all links that are relative to the `base_url` and assume that they can be loaded via XHR. However, as described in the linked issue, static pages may be included as part of the build, but not part of the documentation as such. For this reason, we need a way to distinguish URLs to be loaded via instant loading.

With this PR, before setting up instant loading, the `sitemap.xml` is fetched, which is built by MkDocs and includes all URLs to all documents that are part of the navigation tree. Before dispatching via XHR, the application now checks whether the URL is part of this list. If it is part of the list, the request is loaded via XHR, if it isn't, it is dispatched regularly.

As this is a pretty fundamental change, I'd want to get some feedback before merging this into `master`. Also, locally, the URL must now match the development server host, or instant loading won't work, e.g. when the docs are served with `-a 0.0.0.0`, URLs including `localhost` won't be dispatched. If somebody has a good idea of how to get a canonical URL out of all the different localhost stuff, I'd be happy to integrate it.


squidfunk:
As I received no feedback, we just release it and fix bugs as we go.

sdague:
Something about this change has made it so that instant mode with our site goes completely unresponsive. I downgraded to 5.1.5 and it works again. 

It's so unresponsive that pulling up firefox dev tools doesn't even work, and the browser asks if I want to Wait or Stop the script that's taking the page too long to load.

squidfunk:
Thanks for reporting. Could you open an issue for that including the necessary details? In the meantime you can disable instant loading and everything should be back to normal.

PR #1680, opened 2020-05-10
Add missing gitlab pages in nav and mention GitHub template

Andre601:
Adds the GitLab pages tutorial to the nav in the mkdocs.yml (Was missing) and also mentions the template I made for using MkDocs with GitHub pages.

### DIFF
--- a/docs/tutorials/github-pages.md
+++ b/docs/tutorials/github-pages.md
@@ -75,3 +75,9 @@ Running this command will build the site in the `site` directory.
 If you want a different folder to be used, set the `site_dir` setting in the mkdocs.yml.
 
 After the site is built, you can push these pages to your GitHub repository.
+
+## Template
+A template to use with GitHub Actions can be found at https://github.com/Andre601/mkdocs-template  
+Just click the `Use this template`, choose a name and start with a pre-made repository.
+
+For more information read the `README.md` of the aforementioned repository.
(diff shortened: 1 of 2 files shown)


squidfunk:
Thanks for the PR. However I think we should rather consider adding an `examples` directory to this repo as opposed to linking too many downstream repos, which become out of date more quickly.

Andre601:
> Thanks for the PR. However I think we should rather consider adding an `examples` directory to this repo as opposed to linking too many downstream repos, which become out of date more quickly.

Would be an idea, even tho the chance that a Repo becomes "outdated" would imo be a fairly small one.
Imo could we just note, that the linked repository is not maintained by you and therefore could be outdated.

Additionally should we then also alter the gitlab tutorial as it also mentions a template that can be used.

squidfunk:
> Would be an idea, even tho the chance that a Repo becomes "outdated" would imo be a fairly small one. Imo could we just note, that the linked repository is not maintained by you and therefore could be outdated.

As a background: I'm currently evaluating the survey data and think of restructuring the docs a bit with more examples and a clearer configuration guide, which is why I think integrating it directly into the repository is more natural.

> Additionally should we then also alter the gitlab tutorial as it also mentions a template that can be used.

Yes, I'd also favor this as an example. This could be a good addition to the tutorial section.

Andre601:
I updated the PR to now only update the navigation to add the GitLab page.

I'll later make a PR improving the GitHub Page as I discovered that you don't need a PAT to push changes to the gh_pages branch.

squidfunk:
Thanks!

PR #1686, opened 2020-05-11
Nicer thin scrollbars for pre > code in Firefox

coliff:
Firefox supports `scrollbar-width` styling and `scrollbar-width: thin;` looks much closer to how the custom scrollbars on Chrome look. (REF: https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width)

Note: I haven't tested building the theme with this change, but I have tested using this directly in the browser and it looks great.

Screenshot from Firefox 76 on Windows 10 here with styling applied.
image

### DIFF
--- a/src/assets/stylesheets/base/_typeset.scss
+++ b/src/assets/stylesheets/base/_typeset.scss
@@ -205,6 +205,7 @@ kbd {
       box-shadow: none;
       box-decoration-break: slice;
       touch-action: auto;
+      scrollbar-width: thin;
 
       // Override native scrollbar styles
       &::-webkit-scrollbar {


squidfunk:
Cool, why not! Note that it only seems to affect Windows (or non-macOS).

coliff:
The Netlify deploy preview didn't show the modified scrollbars so I assumed my PR didn't make the necessary changes to the compiled css or something?

squidfunk:
No, the theme is not built as part of CI (that would trigger a loop), so it'll show up on the next build of `master`.

facelessuser:
For that matter, why not use `scrollbar-color` and make them match current chrome styling.

squidfunk:
Seconded!

coliff:
Cool - can take a look and make a new PR

coliff:
actually, I'm not too sure about this. Firefox has dark grey scrollbar on hover which (in my opinion) looks a little bit nicer than Chrome's on-hover scrollbar color blue. I don't think its necessary to make them look exactly the same.

facelessuser:
From a theme perspective, I would think it would make sense that the theme is consistent across browsers. Everywhere that Material themes the scrollbars special, should be themed special in all browsers for consistency. What users do outside of Material for custom styling would be up to them. That's my personal opinion though. It's really up to @squidfunk though.

squidfunk:
> actually, I'm not too sure about this. Firefox has dark grey scrollbar on hover which (in my opinion) looks a little bit nicer than Chrome's on-hover scrollbar color blue. I don't think its necessary to make them look exactly the same.

On `:hover`, the `background-color` is set to the accent color, so I'm also vouching for consistency and settle with @facelessuser here.

coliff:
Agreed :-)  I had a fiddle around with this in Firefox Dev Tools but couldn't get the colors right. I think we want to add `scrollbar-color` [after line 222](https://github.com/squidfunk/mkdocs-material/blob/master/src/assets/stylesheets/base/_typeset.scss#L222) with the variables names - but I'm uncertain of what those variables are. The ones I tested didn't look right (could it be because the variables have transparency so that effects the scrollbar color?). If someone else wants to take a look at this feel free.... 
https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color

facelessuser:
I may take a stab at it. I've used it before, maybe not with transparency. But I think we can accomplish what we want still.

facelessuser:
@squidfunk @coliff 

This is what we needed:

```css
.md-typeset__scrollwrap,
.md-sidebar__scrollwrap,
.md-search__scrollwrap,
.md-typeset pre > code {
  scrollbar-color: var(--md-default-fg-color--lighter) transparent;
  scrollbar-width: thin;

  &:hover {
    scrollbar-color: var(--md-accent-fg-color) transparent;
  }
}
```
This replaced all the scroll bars in firefox. With that said, I noticed that table scrollwraps for whatever reason, even on chrome, weren't styled. That is why I added `.md-typeset__scrollwrap` fixed that.

facelessuser:
I'll create an issue to track this, mainly so I don't forget it. I'll probably put together a pull at some point.

PR #1697, opened 2020-05-15
Fix infinite loop

### LINKED ISSUE #1696: Inifinite loop & browser lock up on v5.1.6
__I checked that...__

- [X] ... the problem doesn't occur with the default MkDocs template
- [X] ... the problem is not in any of my customizations (CSS, JS, template)

## Description

The "check for common prefix" method recently added has a bug when both strings are identical - it spins into an infinite loop and locks up. 

https://github.com/squidfunk/mkdocs-material/blob/master/src/assets/javascripts/index.ts#L386-L387

In our instance, these two strings are both `None` and `None`.  I haven't investigated what this method is actually doing.

Browser locks up when visiting the page (if able to break into the code, you can see it gets stuck on this line)

### Expected behavior

Should not lock up the browser.
 
### Actual behavior

Locks up.

1. Clone https://github.com/soundcloud/intervene
1. Checkout tag `v2.8.5`, which does not have `mkdocs-material` pinned to 5.1.5
1. Ensure mkdocs-material is v5.1.6
1. Run `make mkdocs`
1. Serve the site from the `site` directory locally
1. Open the site in the browser, see it lock up (have devtools open, and pause execution to see spinning on that loop)

### Package versions

v5.1.6
* Python: 3.7.6
* MkDocs: `__main__.py, version 1.1 from /usr/local/lib/python3.7/site-packages/mkdocs (Python 3.7)`
* Material: `Version: 5.1.6`

### Project configuration
```yaml

site_name: intervene
repo_name: intervene
repo_url: https://github.com/soundcloud/intervene

dev_addr: 0.0.0.0:8000

plugins:
  - search

markdown_extensions:
  - admonition
  - codehilite:
      guess_lang: false
  - pymdownx.superfences
  - pymdownx.details
  - toc:
      permalink: 🔗

# ...

theme:
  name: 'material'
  palette:
    primary: 'orange'
    accent: 'deep-purple'
  features:
    - tabs
    - instant
```

### System information

* OS: macos & linux
* Browser: chrome & firefox

bruderstein:
If the URLs being compared were identical, this loop would never finish.

This bounds the loop on the length of the shortest string.

Fixes #1696

### DIFF
--- a/src/assets/javascripts/index.ts
+++ b/src/assets/javascripts/index.ts
@@ -383,7 +383,7 @@ export function initialize(config: unknown) {
 
             /* Determine common prefix */
             let index = 0
-            while (a.charAt(index) === b.charAt(index))
+            while (a.charAt(index) === b.charAt(index) && index < a.length && index < b.length)
               index++
 
             /* Replace common prefix (i.e. base) with effective base */


squidfunk:
Could we short circuit the loop?

``` js
if (a === b)
  index = a.length
else
 while ...
```

bruderstein:
Yeah, absolutely - I haven't had the time to investigate what problem this routine is trying to solve, so there could be many optimisations like this - e.g. is the common prefix likely the whole of `a` or `b`?

I'll add the short circuit.

squidfunk:
The routine just tries to find the common prefix of both strings, which is why it runs in the `undefined` === `undefined` loop if they match exactly 😉Sometimes I make stupid mistakes.

squidfunk:
Bug was fixed in 900c113e59fb1984b8c5cd1fc518c6a1567a088a

PR #1731, opened 2020-05-27
Style Firefox scrollbars

### LINKED ISSUE #1717: Style all scrollbars in Firefox as they are in Chrome
__I checked that...__

- [X] ... the documentation does not mention anything about my idea
- [X] ... to my best knowledge, my idea wouldn't break something for other users
- [X] ... there are no open or closed issues that are related to my idea

## Description

Basically we need the equivalent of (obviously just place in the correct places):

```css
.md-typeset__scrollwrap,
.md-sidebar__scrollwrap,
.md-search__scrollwrap,
.md-typeset pre > code {
  scrollbar-color: var(--md-default-fg-color--lighter) transparent;
  scrollbar-width: thin;

  &:hover {
    scrollbar-color: var(--md-accent-fg-color) transparent;
  }
}
```

This also fixes giant scrollbars in table scroll wraps by styling `.md-typeset__scrollwrap`.

I plan on creating a pull request unless someone gets to it before me.

### Use Cases

Create consistent scroll bars across browsers.

### Screenshots / Mockups

Basically it looks like Chrome scrollbars, but in FireFox. Don't really need a picture. Reference #1686.

facelessuser:
Styles Firefox scrollbars like they are in Chrome.

Exceptions:

1. hover color of thumb occurs when mousing of the object that has the overflow (limitation of Firefox implementation).

2. Scrollbars in macOS don't get styled as mac uses it's own overflow scrollbar. If you disable this feature in mac, it should work. This may be possible through Firefox itself or some other way. At worst, Firefox in macOS looks the same as it always did while Windows and Linux pick up the new styling.

Closes #1717


squidfunk:
Perfect, thanks!

PR #1734, opened 2020-05-29
Add Lighthouse CI

Stanzilla:
This will now run the site against Lighthouse CI. For status check integration, the GitHub app has to be installed, see https://blog.akansh.com/integrate-lighthouse-ci-with-static-site-generators/ for more info.

Alternatively we could check the Netlify preview pages as well https://yashints.dev/blog/2020/04/13/lighthouse-ci

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
           key: ${{ runner.os }}-node-${{ env.NODE_VERSION }}-${{ hashFiles('**/package-lock.json') }}
 
       - if: steps.cache-node.outputs.cache-hit != 'true'
-        run: npm install && npm install -g @lhci/cli@0.4.x
+        run: npm install
 
       # Run linter
       - run: npm run lint
@@ -100,10 +100,7 @@ jobs:
       # Install theme and dependencies
       - run: |
           pip install .
-          pip install mkdocs-minify-plugin>=0.2
-          
-      # Run Lighthouse CI          
-      - run: lhci autorun --upload.target=temporary-public-storage --collect.start-server-command="mkdocs serve" --collect.url=http://localhost:8000/
+          pip install mkdocs-minify-plugin>=0.2    
 
       # Build documentation
       - env:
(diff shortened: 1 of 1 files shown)


squidfunk:
I've seen that Lighthouse has some new monitoring capabilities, but could you explain the necessary steps and gains in more detail? Unfortunately, I'm too short on time to dig into this topic big time, so some directions would be awesome.

squidfunk:
Also note that installation is skipped due to caching:
https://github.com/squidfunk/mkdocs-material/pull/1734/checks?check_run_id=719846326

Stanzilla:
Okay this should be better, can you install https://github.com/apps/lighthouse-ci and set the token secret on the repo?

squidfunk:
Do you have a repo where this is already installed? I'd like to take a look before installing another check suite. Also, as you know, I'm not the biggest fan of Lighthouse, so while we can generate it for every build, I probably won't set it as a mandatory check with thresholds.

.github/workflows/ci.yml
```diff
@@ -100,14 +100,33 @@ jobs:
...
+          pip install mkdocs-minify-plugin>=0.2    
 
       # Build documentation
       - env:
           GOOGLE_ANALYTICS_KEY: ${{ secrets.GOOGLE_ANALYTICS_KEY }}
         run: |
           mkdocs gh-deploy --force
           mkdocs --version
+          
+      - name: Waiting for 200 from the Netlify Preview
```
squidfunk, line 112: Note that it's currently not run for this PR, as the deploy job does solely run on `master`. Thus, it makes no sense to have it here. Also, this will probably increase build time.

Stanzilla:
You can see it in action here https://github.com/JakePartusch/website

.github/workflows/ci.yml
Stanzilla: you can specify how long it waits. and yeah well that depends on where you would want it to run.

squidfunk:
I'm closing this issue, as I do not see much value in running Lighthouse in CI, as it will increase build times with little to no upside. Furthermore, the current score is very good:

Bildschirmfoto 2020-07-27 um 10 01 12

The missing percentages are mainly lost due to "potential savings" in JavaScript files, some of which are external (hello Google Analytics). We should monitor Lighthouse from time to time and improve it when necessary.

PR #1739, opened 2020-06-02
Allow multiple template replacements

Stanzilla:


webpack.config.ts
```diff
@@ -304,7 +304,7 @@ export default (_env: never, args: Configuration): Configuration[] => {
               const manifest = require("./material/assets/manifest.json")
               const template = toPairs<string>(manifest)
                 .reduce((content, [from, to]) => {
-                  return content.replace(from, to)
+                  return content.replace(new RegExp(from, 'g'), to)
```
squidfunk, line 307: Don't wanna be pedantic, but double quotes please ;-) It's configured in TSLint, but it currently doesn't lint the webpack configuration in CI.

Stanzilla: hah, fixed :)

squidfunk:
Great, thanks!

PR #1771, opened 2020-06-22
Fix for Homepage background image gradient on Edge Legacy

coliff:
note the text in the 'Get Started' button is hidden still - but it was completely hidden before anyway (as the whole top was white) - this is still a good improvement.

Screenshots:

Current:
image

With the fix in this PR:
image


squidfunk:
Ah, the `deg` bug again 😁 Thanks!

Issue #1773, opened 2020-06-22
Searching for integers or version numbers does not work


michael-nok:
__I checked that...__

- [x] ... the documentation does not mention anything about my problem
- [] ... the problem doesn't occur with the default MkDocs template
- [x] ... the problem is not in any of my customizations (CSS, JS, template)
- [x] ... there are no open or closed issues that are related to my problem

## Description

When searching for numbers, either on their own, or as part of a word (e.g. SHA256), the search function does not yield any results.

### Expected behavior

I would expect that, by default, searching for numbers (integers) would be supported.  In fact, it seems like it used to be supported, but has been removed or disabled.

### Steps to reproduce the bug

1. Visit https://squidfunk.github.io/mkdocs-material/releases/changelog/ and search for "2020" or "5.3.2".  No results will be found.
2. Visit https://square.github.io/okhttp/changelog/ and search for "2020" or "4.7.1" and the search function will find results.

On my own installation of Material for Mkdocs, I've tested modifying the search.config.pipeline to remove the trimmer and stopWordFilter, but nothing seems to impact the behavior of looking for numbers or digits with decimal points.

squidfunk:
OkHTTP is Running Material 4.x, so I guess this is a regression as there’s no reason why searching for numbers shouldn’t work. It might be related to the transform function that is applied before the query is sent to lunr.js.

squidfunk:
Fixed in a3c03aef. Now that was a super funny bug. It's coming from the search transform's first replacement, which removes rogue control characters:

https://github.com/squidfunk/mkdocs-material/blob/e6e7536356be27ad4da49e8d782f2e875ef323c6/src/assets/javascripts/integrations/search/transform/index.ts#L52

The `-` is not escaped, so the left and right characters, namely `+` and `:` are now forming a range from `+-:`, which includes the characters `+,-./0123456789:`. Everything within this range is filtered from the query, as long as it is either at the beginning of the query, preceded by whitespace. 😁 This is now fixed:

https://github.com/squidfunk/mkdocs-material/blob/a3c03aefb9ea273cc4802e6b451e927345934294/src/assets/javascripts/integrations/search/transform/index.ts#L52

michael-nok:
Amazing, thank you so much for fixing this!  I'm glad you enjoyed the challenge :)

squidfunk:
Released as part of 5.3.3

michael-nok:
[...]

PR #1776, opened 2020-06-25
Update index.ts for #1773

michael-nok:
Additional fix for #1773 to include highlighting integers.


squidfunk:
Hmm, doesn’t work as expected:

7F744DEA-FD55-43A8-A933-696106DC0FFE

michael-nok:
> Hmm, doesn’t work as expected:

I saw that, and I don't know exactly why.  I was able to alter my own local installation to support the highlighting, but changing just this file was insufficient for the deploy-preview.

squidfunk:
Ah you didn’t check in the distribution files. No worries, I’ll do it later. Sorry, was on mobile and had no capacity to check in detail.

PR #1790, opened 2020-07-05
Improve relevance of search results

squidfunk:
Preview

__Try it yourself: [Netlify preview](https://deploy-preview-1790--mkdocs-material-preview.netlify.app/)__

## Problem

Material for MkDocs groups search results by articles (i.e. pages) to give context to each of the section results. This is a significant improvement to the default templates (MkDocs and ReadTheDocs), as it puts sections with the same names (e.g. "Installation") among multiple pages into their respective contexts. This, however, leads to the side effect, that sections with a very low scoring are shown at the top, grouped under the article with the highest score, whereas they may not be as relevant as the first section in the second search result:

- Article 1
  - Section 1 (_Score: 100_)
  - __Section 2 (_Score: 10_)__
  - __Section 3 (_Score: 1_)__
- Article 2
  - Section 1 (_Score: 90_)
  - Section 2 (_Score: 40_)
  - __Section 3 (_Score: 10_)__

## Proposal

The __bold__ sections should be considered _rather_ irrelevant. This PR adds the score to the returned search results and proposes a new technique to ensure more relevant search results above the fold of the search dialog. 

- Render the first section of the first result, as it has the highest score
- Next, check the score of the first section of the next result.
  - If it has a __lower score__ than the second section of the first result, render the second sections.
  - If it has a __higher score__, collapse all remaining sections of the first result under __More on this page__ button
- Repeat for all subsequent pairwise results.

Now, the user can decide to expand and dive into the results of an article, while the top-level search results should be much more concise:

- Article 1
  - Section 1 (_Score: 100_)
  - __More on this page__
    - Section 2 (_Score: 10_)
    - Section 3 (_Score: 1_)
- Article 2
  - Section 1 (_Score: 90_)
  - Section 2 (_Score: 40_)
  - __More on this page__
    - Section 3 (_Score: 10_)

## Progress

- [x] Add `score` to search results returned from index
- [x] Adjust application to handle new search result data structure
- [x] Determine thresholds for grouping and pass them to the rendering function
- [x] Group results when rendering them
- [x] Expand search results
- [ ] Add and use translation
- [ ] Clean up template implementation after feedback


chrieke:
Sounds like a great idea! I guess if there are no "Article 2" results, the full "Article 1" section is rendered? Could you please add a screenshot how the "More on this page" line would be rendered?

squidfunk:
This should show the general idea:

Bildschirmfoto 2020-07-05 um 14 08 40

squidfunk:
@chrieke the _More results on this page_ button will just expand the remaining search results, so they are exactly rendered like the other results. The only difference now is that they're grouped under this button. They will just be rendered inline into the search results. I've prototyped the grouping and button (still not functional), should be available in the preview in a few minutes.

squidfunk:
The Netlify preview now groups results. __Expansion doesn't work just yet__, but the grouping should be functional. The rendering is still hacky, just wanted to collect some feedback before I go down the wrong road.

squidfunk:
I've got a candidate ready! Check out the preview linked in the OP. Any feedback appreciated!

bruderstein:
Hey, I checked this out on our large installation (following raising #1799 - apologies for missing this PR!)

It doesn't bring the top search result back first. In fact, the document with the best score is the 30th document in the list. The one with the second best score is 27th. The first entry in the list contains 10 matches, but they are all weak matches. It seems to be sorting the documents by the sum, rather than by the top match, which _might_ return reasonable results in some circumstances, but in many cases it's probably worse.  The actual test case i'm using, for example, contains the word "is". So one document has many matches, including words (or acronyms) like `issue`, `ISRC`. 10 matches of 1.0xxx on this page is being ranked higher than a single match of 10.xxx.

Sorting via the best match for the document might give better results (at least in our case).

e.g.
Article 5
  - #foo score 10
  - #bar score 5
Article 3
  - #baz score 8
  - #quz score 3
  - #biz score 1
Article 7
  - #foo score 5
  - #bar score 4
  - #baz score 3
  - #quz score 3
  - #fiz score 1
Article 8
  - #a score 1
  - #b score 1
  - #c score 1
  - #d score 1
  - #e score 1
  - #f score 1
  - #g score 1
  - #h score 1
  - #i score 1
  - #j score 1
  - #k score 1
  - #l score 1
  - #k score 1
  - #m score 1
  - #n score 1
  - #p score 1
   - #q score 1
   

This branch still seems to put Article 8 at the top (total 17), then Article 7 (score 16), then 5 (score 15), then 3 (score 12), when actually, the best hits are 5, 3, 7, in the order shown.

squidfunk:
Could you provide a reproducible example?

bruderstein:
TL;DR:  appending `*` to every term is lowering the search accuracy.  Reproducible searching for `Material for MkDocs selects the (best-)matching stemmer for the given theme language` with the docs for this project in readthedocs vs this theme.

The longer version:

I did some investigation into this last night, and it appears that it's due to the appending of `*` to every word term (which doesn't happen in the readthedocs theme, for example). My assumptions above are therefore not correct.

I `console.log`ged the results from the docs for this project, searching for the phrase `Material for MkDocs selects the (best-)matching stemmer for the given theme language`, where the search page where that is featured appears fourth. The score is however lower than the first result, due to it actually searching for `Material* for* MkDocs* selects*...`.  For my test case, the `is` is converted to `is*`, which then matches `issue` and `ISRC`, which are rarer terms in the corpus, hence return a higher score.

squidfunk:
Thanks for providing those insights! We have to find a way to ensure good accuracy while preserving the type-ahead experience. Without the asterisks at the end, type-ahead will be lost and results will be very erratic while typing, as can be seen on [mkdocs.org](https://mkdocs.org).

I'm closing this PR, since it's obvious that we have to invest more work into this topic. I'll be revisiting search as part of the latest Sponsorware initiative I just started (see the pinned issue).

bruderstein:
What about only starring the last word? Or the word where the cursor is? Just a thought. Thanks for considering these issues.

squidfunk:
That might also yield unstable results if you don't finish typing the word before. However, I think I have a very good idea how this could be solved :-) I'll be posting my progress in the issue you opened.

PR #1813, opened 2020-07-17
Fix instant page load race which causes image 404s

### LINKED ISSUE #1638: Instant loading sometimes causes 404s on assets (e.g. images)
First, thank you for the awesome work!

__I checked that...__

- [X] ... the problem doesn't occur with the default MkDocs template
- [X] ... the problem is not in any of my customizations (CSS, JS, template)

## Description

I do have pages with several pictures on pages which are in subfolders. 

When `features instant` is set in `mkdocs.yml`, some of the images may not display. In the DevTools console, I see 404 errors in the network tab. I also see the 404 errors in `mkdocs` stdout.

Refreshing the page does display the images.

### Expected behavior

Images should display without having to refresh.

### Actual behavior

Images do not display (not systematically), and getting 404 errors in the DevTools Network tab.

1. Have a site structured like this, with instant:
```
docs/
  folder1/
    README.md
    pic1.png
    pic2.png
    pic3.png
  folder2/
    README.md
    pic4.png
    pic5.png
    pic6.png
  README.md
```

2. Serve locally `mkdocs serve`
3. Observe 404 errors

### Package versions

* Python: `python --version`: 3.7.5
* MkDocs: `mkdocs --version`; mkdocs, version 1.1
* Material: `pip show mkdocs-material | grep -E ^Version`: version 5.1.1

### Project configuration

``` yaml
site_name: "Test"
site_description: "Testing"
theme: 
  name: material
  features:
    - tabs
    - instant
  icon:
    logo: material/cloud
    repo: fontawesome/brands/github
  include_search_page: false
  search_index_only: true

  plugins:
  - search
  - awesome-pages:
      collapse_single_pages: true

markdown_extensions:
  - meta
  - admonition
```

### System information

* OS: Windows10, WSL (Linux Subsystem / Ubuntu)
* Browser: reproducible in Chrome, MS Edge, Firefox

gartnera:
I've been digging into #1638. It's hard to reproduce but I've seen it happen occasionally with remote servers. It seems that the new document is added to the page before `history.pushState` is called. This results in urls being calculated with the old url instead of new url.

With this change, the `history.pushState` is called immediately after the ajax request and before the HTML is parsed.

I'm not very familiar with RxJS btw, please advise if you can think of a better way.

Fixes #1638

### DIFF
--- a/src/assets/javascripts/integrations/instant/index.ts
+++ b/src/assets/javascripts/integrations/instant/index.ts
@@ -61,6 +61,7 @@ import {
 interface State {
   url: URL                             /* State URL */
   offset?: ViewportOffset              /* State viewport offset */
+  op: string                           /* Current operation (push or pop) */
 }
 
 /* ------------------------------------------------------------------------- */
@@ -137,7 +138,7 @@ export function setupInstantLoading(
         }
         return NEVER
       }),
-      map(el => ({ url: new URL(el.href) })),
+      map(el => ({ url: new URL(el.href), op: "push" })),
       share<State>()
     )
 
@@ -159,13 +160,15 @@ export function setupInstantLoading(
       filter(ev => ev.state !== null),
       map(ev => ({
         url: new URL(location.href),
-        offset: ev.state
+        offset: ev.state,
+        op: "pop"
       })),
       share<State>()
     )
 
   /* Emit location change */
-  merge(push$, pop$)
+  const locationChange$ = merge(push$, pop$)
+  locationChange$
     .pipe(
       distinctUntilChanged((prev, next) => prev.url.href === next.url.href),
       pluck("url")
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for inquiring into this problem! It's great that you found a way around it! I'm still trying to reproduce the problem, sadly unsuccessfully until now.

You're right, it's not (yet) idiomatic RxJS, but don't worry, I'll look into it as soon as I find some time 😊

gartnera:
I can code it a different way if you have any tips. Here's maybe a bit more insight into what is happening in #1638 (top is earliest, bottom is latest):

Screenshot_2020-07-17 12 32 39_AyZ26b
Screenshot_2020-07-17 12 37 24_w0wCLG
Screenshot_2020-07-17 12 37 41_EvZQjj

squidfunk:
I haven't had the time yet to come back to this, but I definitely will! Thanks for the additional data!

gartnera:
Here's a `.whl` with this change if any lurkers want to use it: https://github.com/gartnera/mkdocs-material/releases/download/5.5.2/mkdocs_material-5.5.2-py3-none-any.whl

squidfunk:
So, I've finally found some time to dive into this. Again, thanks for providing this PR and the insights. I was finally able to reproduce it (sometimes) in Firefox and Safari. In Chrome, everything worked as expected.

The problem is definitely related to the execution order among subscribers. It seems that sometimes, history management and document parsing were executed in reverse order. This should be mitigated by adding a `share` operator to `ajax$` as done in 0a0fee34. The `share` operator will create an internal `Subject` to multicast values among subscribers (which makes sense anyway, as we don't want to execute the `ajax` chain more than once. This `Subject` will emit the values it's fed to its subscribers in-order.

My testing shows that the bug is fixed now. Could you check the latest `master`?

gartnera:
Yeah I think 0a0fee3 fixes it, closing this now.

squidfunk:
Perfect, thanks for the quick feedback! I'll issue a bugfix release straight away.

squidfunk:
5.5.12 is out 😊

PR #1827, opened 2020-07-26
Avoid content shining through clipboard icon

aimeos:
image
instead of:
image

### DIFF
--- a/src/assets/stylesheets/main/layout/_clipboard.scss
+++ b/src/assets/stylesheets/main/layout/_clipboard.scss
@@ -27,12 +27,14 @@
 // Copy to clipboard
 .md-clipboard {
   position: absolute;
-  top: px2rem(8px);
-  right: px2em(8px, 16px);
+  top: 0;
+  right: 0;
   z-index: 1;
-  width: px2em(24px, 16px);
-  height: px2em(24px, 16px);
+  padding: px2em(8px);
+  width: px2em(32px, 24px);
+  height: px2em(32px, 24px);
   color: var(--md-default-fg-color--lightest);
+  background-color: var(--md-code-bg-color);
   border-radius: px2rem(2px);
   cursor: pointer;
   transition: color 125ms;


aimeos:
Please check if this are the correct dimensions:
```
  width: px2em(32px, 24px);
  height: px2em(32px, 24px);
```

squidfunk:
Thanks, that makes sense from a usability perspective. However, I don't really like that the button covers so much of the area and I'm also thinking of adding a dark mode toggle for code blocks, so there would be another button covering up even more space. I fiddled around, and found that the following `background` does work pretty well:

``` css
.md-clipboard {
  background: radial-gradient(var(--md-code-bg-color) 40%, transparent 70%)
}
```

aimeos:
Using your background style still looks strage:
image
But you decide how the original theme should hande that

squidfunk:
Maybe there's another way. Let me think about this a little 😊

facelessuser:
- I think this could work if the copy button filled the upper right corner and only covered the first line. So pad it out and set it to `top: 0` and `right: 0`.
- Then pad the `<code>` element under `<pre>` on the right side.

Then it will only cover the first line, and you can always scroll all the code out from under the icon.

I guess you could always have it padding out to take up the right side of all lines equally as well 🤷 .

facelessuser:
copy

facelessuser:
I think I made the icon smaller, but you get the point

squidfunk:
@facelessuser looks good, could you share the code?

facelessuser:
Yeah, I'll clean it up and post it here. I'll make sure I don't make the icon tiny 🙂 .

squidfunk:
No worries, I can take a look too if you don't want to invest the time.

facelessuser:
Nah, I don't mind. I was just screwing with it in the browser. I can illustrate in something more clean, but if you get to it before me that is fine too. I wanted to get my thoughts down though.

facelessuser:
Here's the basics. You don't even have to pad the top it seems:

```css
.md-typeset pre>code {
  padding-right: 2.5em;
}

.highlight .md-clipboard {

  right: 0;
  width: 2em;
  padding-right: 0.5em;
  background-color: var(--md-code-bg-color);
}
```

With that said, I'd check out code blocks under tabbed containers. I think margins get adjusted, so you would need to update some positioning there potentially. I think it would be about `right: px2rem(-12px);`. I can't recall if this is because I treat code blocks under tabs differently than you do though. I still only expand mine when the code block is an only child, in all other cases, I treat them as if the tabbed is a container (similar to details). So, just make sure code blocks look good under tabs.

squidfunk:
That was easier than expected. Got a solid implementation in a217be3e. I tested with content tabs, code in Admonitions, etc., seems to work very well. I adjusted the padding a bit to make a little more room to the left, so the horizontal spacing is symmetrical. Note that I normalized a lot of `rem`s to `em`s today in 4b2c0e5a88ac6e73543dd7aa545b95f93cee349f, which made this particularly easy.

squidfunk:
The only thing I've found so far is – it doesn't work well with highlighted lines, due to the padding. I'll look into it tomorrow:

Bildschirmfoto 2020-08-01 um 20 43 24

facelessuser:
Seems like it may be better just to have something that extends the entire right side that the content scrolls under 🤷 .

squidfunk:
I reverted the changes for now, as I need to make a new release tomorrow. I'll be revisiting this topic in the next weeks.

PR #1833, opened 2020-07-27
Update _keys.scss

twardoch:
The symbols in the current version are OK, except for

```
    "arrow-down": "\2193", // ↓
    "arrow-left": "\2190", // ←
    "arrow-right": "\2192", // →
    "arrow-up": "\2191", // ↑
    "tab": "\21E5", // ⇥
```

The symbols above are better than the ones used today. 

The bloody autoformatter in VSCode has reformatted the SCSS :/

### DIFF
--- a/docs/reference/lists.md
+++ b/docs/reference/lists.md
@@ -97,7 +97,7 @@ _Result_:
 
 An ordered list must start with a number immediately followed by a dot. The 
 numbers do not need to be consecutive and can be all set to `1.`, as they will
-be re-numbered when renderer.
+be re-numbered when rendered.
 
 _Example_:
 


facelessuser:
I think key preference is highly subjective. I've found the current arrows read much easier, especially with different fonts. I only comment because this CSS was swiped from the [pymdown-extensions project](https://facelessuser.github.io/pymdown-extensions/extensions/keys/) which I maintain.

As for tab 🤷 , again, subjective, but I have less of an opinion on that one.

squidfunk:
I'm fine with changing the keys, and your changes in general, as I don't very much like the triangles. I can understand @facelessuer's point, but it's a matter of opinion after all.

However, I can't merge this PR like that, as the formatting is jagged. Could you please keep the original formatting?

twardoch:
OK, I’ll fix the PR :) 
I checked the triangles, and they genuinely look confusing — arrow up looks like Greek Delta. It’s true that on macOS keyboards, the arrow keys are triangles, but I think as a generic solution, the arrows are better. Also, I used the same Unicodes as macOS uses when it shows the shortcuts in the menus

twardoch:
@squidfunk @facelessuser I’m using the `keys` stuff extensively in my FontLab 7 documentation. If you look at https://help.fontlab.com/fontlab/7/manual/Release-Notes-7511/ — this is actually a customized version of MkDocs Material (with much less “material” feel ;) ).

facelessuser:
That's fine, I've personally gone back and forth on some of the glyphs myself, arrows included.

twardoch:
Superseded in https://github.com/squidfunk/mkdocs-material/pull/1836

PR #1836, opened 2020-07-28
[_keys.scss] Visualize `arrow-down` as ↓, `arrow-left` as ←, `arrow-right` as →, `arrow-up` as ↑ and `tab` as ⇥

twardoch:
Changed key visualization for 

```
    "arrow-down": "\2193", // ↓
    "arrow-left": "\2190", // ←
    "arrow-right": "\2192", // →
    "arrow-up": "\2191", // ↑
    "tab": "\21E5", // ⇥
```

and fixed a few lowercase hex to be uppercase for consistency. This replaces https://github.com/squidfunk/mkdocs-material/pull/1833 which was failed

### DIFF
--- a/src/assets/stylesheets/main/extensions/pymdownx/_keys.scss
+++ b/src/assets/stylesheets/main/extensions/pymdownx/_keys.scss
@@ -73,10 +73,10 @@
     "right-windows": "\229E",
 
     // Other keys
-    "arrow-down":    "\25bd",
-    "arrow-left":    "\25c1",
-    "arrow-right":   "\25b7",
-    "arrow-up":      "\25b3",
+    "arrow-down":    "\2193",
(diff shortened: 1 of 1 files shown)


twardoch:
This visualization uses the same characters that Apple uses in macOS when showing keyboard shortcuts in menus etc., example:

scr- 2020-07-28 at 18 50 59

facelessuser:
Those arrows in the screenshot look way better than the tiny arrows most fonts render.

squidfunk:
Could we use another (system) font for the arrows?

squidfunk:
So I decided to merge this, because the arrows are not shown alone, but are always accompanied by "Down", "Up", etc, so it doesn't matter that much. I personally dislike the delta-look, but it's super easy to change it back with a few lines of CSS.

PR #1856, opened 2020-08-08
Support for Referrer-Policy metadata

LucidOne:
Provide a means of setting `referrerpolicy` attribute via metadata.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy

_mkdocs.yml_:
```
extra:
  referrer: same-origin
```

### DIFF
--- a/src/base.html
+++ b/src/base.html
@@ -35,6 +35,11 @@
       <meta charset="utf-8" />
       <meta name="viewport" content="width=device-width,initial-scale=1" />
 
+      <!-- Referrer policy -->
+      {% if config.extra.referrer %}
+        <meta name="referrer" content="{{ config.extra.referrer }}" />
+      {% endif %}
+
       <!-- Page description -->
       {% if page and page.meta and page.meta.description %}
         <meta name="description" content="{{ page.meta.description }}" />


squidfunk:
Thanks for the PR. With 5.5.0, we have [deprecated several metadata flags](https://squidfunk.github.io/mkdocs-material/deprecations/), including redirects, as the recommended way now is to extend the theme. There are just too many `meta` tags that we would need to provide all of them via extra variables. Theme extension is very easy.

As an example see the [Open Graph and Twitter Cards](https://squidfunk.github.io/mkdocs-material/reference/meta-tags/#customization) guide, which both use this mechanism.

LucidOne:
That's not unreasonable and locally I'm extending the theme, but maybe it would make more sense to loop through a dict in the template to generate all the meta attributes?

squidfunk:
The Problem is that some meta attributes are set  is more complex logic than plain assignment. See the `base.html` template for some examples.

A second downside of setting it via `mkdocs.yml` is that you can only set them once, side-wide, whereas with theme extension you can set it per-document, if desired.

LucidOne:
Should I reformat this as an example in the docs?

squidfunk:
I generalized the meta tags documentation in c09e5f1c and added Twitter Cards and Open Graph as an example. `Referrer-Policy` albeit useful, is rather rarely used, so we'll leave it at that, as we cannot add every possible meta tag to the documentation.

PR #1903, opened 2020-09-04
Update emoji settings

timvink:
Syntax has been updated, see https://facelessuser.github.io/pymdown-extensions/extensions/emoji/#using-with-mkdocs


squidfunk:
We’re using Material Extensions, which builds atop PymdownX. This is necessary for the icon integration. If you switch to PymdownX, you loose icons. This, the setting is correct 😉

PR #1928, opened 2020-09-15
Add rss plugin to Dockerfile

ericdallo:
At https://emacs-lsp.github.io/lsp-mode we use the docker image and want to add rss support

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,6 +49,7 @@ RUN \
         'mkdocs-git-revision-date-localized-plugin>=0.4' \
         'mkdocs-minify-plugin>=0.3' \
         'mkdocs-redirects>=1.0'; \
+        'mkdocs-rss-plugin>=0.6.1'; \
     fi \
   && apk del .build gcc musl-dev \
   && rm -rf /tmp/*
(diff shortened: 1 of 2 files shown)


squidfunk:
Thanks for the PR, however the plugin doesn’t look like it’s very popular (very few stars), so I think at the current moment it’s not worth increasing the size of the Docker image.

ericdallo:
That's sad, so emacs-lsp would need to stop using the docker image and setup mkdocs-material manually [here](https://github.com/emacs-lsp/lsp-mode/blob/master/.github/workflows/docs.yml#L34)?

squidfunk:
No, you can continue to use the Docker image by extending it and adding the plugins you need:

``` docker
FROM squidfunk/mkdocs-material
RUN pip install ... whatever you need
```

In fact, I'm thinking about removing all plugins from the Docker image, as more and more requests for individual plugins are coming up and we cannot bundle everything.

squidfunk:
I added [a guide](https://squidfunk.github.io/mkdocs-material/getting-started/#with-docker) how to install plugins in the Docker image in e0b53999. Closing this PR for the reasons stated. If in the future, this PR gets many upvotes, we can bundle the RSS plugin, but for now, you seem to be the only one using it.

ericdallo:
It makes sense, thanks for the explanation @squidfunk, I'll do that :)

Issue #1917, opened 2020-09-11
Macros plugin


ajcollett:
__I checked that...__

- [x] ... the documentation does not mention anything about my idea
- [x] ... to my best knowledge, my idea wouldn't break something for other users
- [x] ... there are no open or closed issues that are related to my idea

## Description

I found this plugin recently: https://github.com/fralau/mkdocs_macros_plugin

I know it's supported, but not in the docker image. 

My question is, why is it not in? Is it just new? Or is there a fundamental reason to avoid it? If no, might it be worth adding?

More importantly, is there any reason this Plugin will *not* work with MKdocs Material?

### Use Cases

It would be great to be able to define variables and macros that one can then use all over the documentation. I know there are some plugins already that support some form of snippet or variables, but this seems to be the "one ring to rule them all"?

Specifically, I had a table to create, with a bunch of links to the same website, but with different parameters and paths. The table becomes unwieldy to edit, if say something like the server URL changes. With macros, that would be quite a lot easier.

facelessuser:
The answer is probably more simple than that. The docker image is not meant to contain every possible mkdocs plugin. @squidfunk has included the ones he's deemed most useful. If requirements are needed outside of that, you'd probably have to roll your own or not use the docker image. To be fair, I never use the docker image and have no problem pushing my documents 🤷.

squidfunk:
In the end, there are thousands of MkDocs plugins out there. We try to strike a balance not blowing the image up too much by installing all of them and providing meaningful defaults. If this is useful to more people, we can integrate it. However, note that it's very simple to extend the Docker image to install your own extensions:

``` Dockerfile
FROM squidfunk/mkdocs-material
RUN pip install mkdocs-macros-plugin
```

That should be about it.

ajcollett:
Thanks for the feedback. 

I have quite a few pipelines that make use of the MKDocs Material image, one of which I have done exactly that (what @squidfunk mentioned). It's a little more complex, since it's to create a PDF of the site as well as well as the site, but same idea. Roll your own. 

However, where I can I try stick with the provided image, since it's one less thing to maintain (I have lots of sys admin tasks :-P).

I was also thinking this might be really helpful to lots of people. Thus to maybe integrate it.

However, since the plugin has existed for a while, and no one else has mentioned it, maybe the want/need for it to be built in isn't great. In which case, I think I'll close this. If others come along in the future, maybe this can be the "first port of call" issue?

[...]

PR #1929, opened 2020-09-16
Add reference page for variables and macros (#1917)

fralau:
As requested in #1917 here is the documentation page for the [macros plugin](https://mkdocs-macros-plugin.readthedocs.io/en/latest/).

### DIFF
--- a/docs/reference/variables-macros.md
+++ b/docs/reference/variables-macros.md
@@ -1,0 +1,332 @@
+---
+template: overrides/main.html
+---
+
+# Variables and Macros
+
+The power of the Markdown language comes from its simplicity and
+visual clarity. Its downside is that it far more limited than HTML.
+That is the reason why Markdown extensions have been developed,
+such as [Footnotes](../Footnotes) or [Admonition](../Admonition)
+(for notes and tips).
+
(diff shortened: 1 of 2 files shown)


squidfunk:
Awesome, thanks for taking the time to write this! I'm merging it right away and might adapt it a little to match the other pages, but this is a great start!

squidfunk:
Thanks again for writing this. I decided to strip it down to the barebones of using variables, as I believe it's better to link to your documentation for the more advanced cases like custom macros etc. The setup is quite lengthy and I don't want it to get out of sync, when something changes on your side. Nevertheless, I believe this plugin is a huge value add and love that it's now mentioned on the docs!

fralau:
> Thanks again for writing this. I decided to strip it down to the barebones of using variables, as I believe it's better to link to your documentation for the more advanced cases like custom macros etc. The setup is quite lengthy and I don't want it to get out of sync, when something changes on your side. 

@squidfunk Thanks and the pleasure was mine. 

I do understand about your concern on the length, and was not sure how much I could it.

> Nevertheless, I believe this plugin is a huge value add and love that it's now mentioned on the docs!

Thank you for your kind acknowledgement, this is a great reward 🙂.

PR #1947, opened 2020-09-27
Fix links to deprecations

lgeiger:
The docs currently link to localhost which breaks the links on the deployed version. This PR removes the prefix to fix the links.


squidfunk:
Ha! Clumsy me 😅 Thanks!

PR #1948, opened 2020-09-27
Fix deprecations links (second try)

lgeiger:
I messed up #1947 and the links now point to `upgrading/deprecations/#hero` this should hopefully fix it for real. Sorry about that.


squidfunk:
No worries, thanks for fixing it!

PR #1955, opened 2020-09-29
Fix header link when use_directory_urls is unset

ivoire:
By default use_directory_urls is set to True. But when set to False, the header
link should be "/index.html" and not "/".

### DIFF
--- a/material/partials/header.html
+++ b/material/partials/header.html
@@ -3,7 +3,7 @@
 -#}
 <header class="md-header" data-md-component="header">
   <nav class="md-header-nav md-grid" aria-label="{{ lang.t('header.title') }}">
-    <a href="{{ config.site_url | default(nav.homepage.url, true) | url }}" title="{{ config.site_name }}" class="md-header-nav__button md-logo" aria-label="{{ config.site_name }}">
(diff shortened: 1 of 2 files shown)


squidfunk:
Thanks for the PR: Does this also work when `site_url` is not set in the config? I'd guess that `nav.homepage.url` will also include `index.html` when `use_directory_urls` is set to `false`.

squidfunk:
Thanks again for your PR. However, there are a few problems with it:

1. You forgot to adjust the link in `nav.html`. I went ahead and add them in the commit mentioned below.
2. if `site_url` is set to an external domain, i.e. the base domain if the documentation is hosted under a subdomain, it will be appended to the main domain, potentially breaking the URL. For example, if you specify http://mycompany.com/ as `site_url`, which redirects to https://mycompany.com/index.php, appending `index.html` might break the URL. For this reason, it's not generalizable.
3. Nevertheless, there's the case when `site_url` is not set and `use_directory_urls`is set to `false`, which is necessary to build offline documentation. This case is now handled as part of 454339bd.

ivoire:
Thanks for the correct fix. It's in fact for a local build so `site_url` is not set as you guessed.

PR #1957, opened 2020-10-01
Use dot notation for Markdown extensions

dhoppe:
The documentation should be consistent with your own config, as it is likely to be used as a reference:
* https://github.com/squidfunk/mkdocs-material/blob/master/mkdocs.yml#L104-L112


squidfunk:
That's actually just my personal preference. If you find that confusing, I'd rather switch the stuff in `mkdocs.yml` back to what's mentioned in the docs than the other way round.

squidfunk:
Supersedes by de173678.

PR #2006, opened 2020-10-23
Update variables.md

bs-jena:
* variables do not work in snippets included using Snippets extension
* workaround: use macros plugin to define snippets (as described here https://mkdocs-macros-plugin.readthedocs.io/en/latest/advanced/#including-snippets-in-pages)


squidfunk:
Thanks for writing this up!

PR #2025, opened 2020-11-02
Add Chinese search tips

Writeup007:
Other language such as English has `search.result.initializer` attribute, but Chinese dones't have, so i add it in Chinese language file.
China is also a country worthy of attention. If you want to learn more about Chinese, please let me know and I will help you.

Have a good day

### DIFF
--- a/material/partials/language/zh-Hant.html
+++ b/material/partials/language/zh-Hant.html
@@ -14,7 +14,7 @@
   "search.config.pipeline": "trimmer, stemmer",
   "search.config.separator": "[\，\。]+",
   "search.placeholder": "搜尋",
-  "search.result.initializer": "正在初始化檢索引擎",
+  "search.result.initializer": "正在初始化搜尋引擎",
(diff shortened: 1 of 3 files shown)


squidfunk:
Thanks for the PR! Could you please make your changes in `src` as well? The `material` folder will be overwritten on build, which is why they contain this header:

https://github.com/squidfunk/mkdocs-material/blob/9c47efc57defcd9065dd89ecdb0f2f0234db7d2f/material/partials/language/zh-Hant.html#L1-L3

Writeup007:
> Thanks for the PR! Could you please make your changes in `src` as well? The `material` folder will be overwritten on build, which is why they contain this header:
> 
> https://github.com/squidfunk/mkdocs-material/blob/9c47efc57defcd9065dd89ecdb0f2f0234db7d2f/material/partials/language/zh-Hant.html#L1-L3

Thank you for your reminder, and i will solve it again.

PR #2029, opened 2020-11-04
Add page keywords

Writeup007:
Add page keywords, I think it's helpful for [SEO](https://en.wikipedia.org/wiki/Search_engine_optimization) (Search Engine Optimization)


squidfunk:
I didn't even know that MkDocs supported keywords. Honestly, they're not really relevant anymore, but as MkDocs supports them, we can include it.

PR #2050, opened 2020-11-15
Change the default branch name for GitHub

hwchiu:
GitHub already changed the default branch from the `master` to `main` for a while.

I think it will be better if we can change the branch name we had in the example so that the first-time user can easily try it by copy and paste the GitHub Action workflow yaml files.

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -30,7 +30,7 @@ contents:
     on:
       push:
         branches:
-          - master
+          - main
     jobs:
       deploy:
         runs-on: ubuntu-latest
@@ -50,7 +50,7 @@ contents:
     on:
       push:
         branches:
-          - master
+          - main
     jobs:
       deploy:
         runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


hwchiu:
I found someone still gets the master branch not the main branch, even he/she create the repo today, keep the master is good.

squidfunk:
Ok great, let's keep it simple then 😊

stevemar:
Hi there :wave:, I just ran into this issue. My default branch was `main` and the documented step didn't work at first. It took a few minutes of poking around before I clued in what was going on.

I was going to suggest a PR but saw this one was open. What do y'all think about specifying both branches in the example? and add some wording around picking one? 

```yaml
    name: ci
    on:
      push:
        branches:
          - master
          - main
```

Thanks for maintaining and creating this project @squidfunk !

squidfunk:
I think it’s reasonable to include both in the example, so feel free to create a PR. I wouldn't add too much additional prose, though; just keep it short and concise.

PR #2059, opened 2020-11-22
Add a helpful error message when attempting to call `mike install-extras`

jimporter:
Now that the Material theme has built-in support for mike (thanks for the work, by the way!), I thought it'd be helpful for `mike install-extras` to report an error directing them to Material's documentation if a user attempts to install extras for their Material-themed docs. This patch causes `mike` to report the following error (assuming you're using the just-released v0.6):

```
$ mike install-extras
mike: install-extras not required to use mike with material theme; for more info
rmation, see https://squidfunk.github.io/mkdocs-material/setup/setting-up-versio
ning/#versioning
```

### DIFF
--- a/material/mike.py
+++ b/material/mike.py
@@ -1,0 +1,5 @@
+raise ValueError(
+    'install-extras not required to use mike with material theme; '
+    'for more information, see https://squidfunk.github.io/mkdocs-material/' +
+    'setup/setting-up-versioning/#versioning'
+)
--- a/setup.py
+++ b/setup.py
@@ -64,6 +64,9 @@ setup(
     entry_points = {
         "mkdocs.themes": [
             "material = material",
+        ],
+        "mike.themes": [
+            "material = material.mike",
         ]
     },
     zip_safe = False


squidfunk:
Thanks for the PR! The mike integration ha chrrerky reserved to users of [Insiders](https://squidfunk.github.io/mkdocs-material/insiders/), which means it's not generally available. When the [https://squidfunk.github.io/mkdocs-material/insiders/#bhut-jolokia](funding goal) is reached, it's merged back into master. Thus, I suggest we wait until this is the case (we’re nearly half way there) 😉

jimporter:
Sounds good to me! (I actually didn't realize that's how Insiders worked for this project, so I'll have to think about whether I want to roll back the PyPI release to v0.5.5 or just leave it as-is, since it's not hard to type `pip install mike==0.5.5`...)

squidfunk:
I'm sorry, maybe I could've made that more clear. Regarding the roll back - that's up to you, but my current projections are that we reach that goal in ~6 months, so new users might miss out on newer mike features.

jimporter:
No worries, I should have read the docs! Overall, I consider mike to be pretty much done, so the only new thing in v0.6 was removing the extras for Material (and a tiny improvement to an error message). I just marked v0.6 as a pre-release so people can still get it if they want, but I doubt anyone will actually need 0.6 for anything. :)

jimporter:
Closing this; since https://github.com/jimporter/mike/commit/e017d3a2be4f553ae682999b142216f6407c0dc8, mike doesn't use `install-extras`, so there's nothing to do here anymore. However, note that I've added an MkDocs plugin for mike that you may (or may not) want to mention in your docs. All it'll do for your users is update the `site_url` to include the canonical version. See here: https://github.com/jimporter/mike#initialization

squidfunk:
Thanks for the update. I though that this was already solved by the `mike` executable? Does this mean that currently (i.e. without using the plugin), the `site_url` is incorrect for all versions?

jimporter:
Yes, under mike 0.5.5, if you set `site_url` to `https://example.com/` in your mkdocs.yml, then navigating to `https://example.com/1.0/page/` will have the canonical URL set to `https://example.com/page/`, which is almost certainly wrong. (If you don't set `site_url` in mkdocs.yml, there's no problem.) That's fixed in 0.6.

squidfunk:
Okay, so this means that we should definitely recommend the use of the plugin in our documentation, right? Why is it implemented as a plugin? Since `mike` is a meta-utility, I don't see the reason for that. Am I missing something?

jimporter:
A plugin is the cleanest way to manipulate the user's `mkdocs.yml` config. Otherwise, I'd have to create a temporary file (in the same directory as the original so that relative paths don't get messed up) to do the manipulations I need to do. That's technically doable, but I'd rather avoid it if possible.

If the `mkdocs` command had more options to allow for meta-utilities (e.g. being able to override `mkdocs.yml` settings via a command line option), this could probably be a lot more transparent.

jimporter:
... thinking about it more, there may be some value in adding the mike plugin to the config automatically so that people don't need to do anything in particular to get the `site_url` fix *or* the version selector for the default themes.

jimporter:
Ok, mike now automatically adds its plugin to a temporary `mkdocs.yml` file (unless the user has already explicitly enabled it), so there's nothing left to do on your end. It should all work automatically (at least on the mike side of things).

squidfunk:
@jimporter Material for MkDocs 7 is out which includes version support!

jimporter:
@squidfunk Great! mike v1.0 is pretty close to ready, so hopefully I'll be able to publish it in a week or so. I don't think your theme needs to do anything special to handle mike 1.0, but I'll double-check if there's anything that should be mentioned in your docs.

squidfunk:
@jimporter thanks! Really looking forward to the 1.0 release!

PR #2069, opened 2020-11-24
Add mkdocs-redirects when running a dev environment

ojacques:
`mkdocs-redirects` is not part of `requirements.txt`, yet required when running the development environment. 
Like `mkdocs-minify-plugin`, this PR adds one line in the documentation to install `mkdocs-redirects`.

It is unclear to me why this is not in `requirements.txt`, but this PR will start the discussion!


squidfunk:
This is because `mkdocs-redirects` isn't an essential dependency of the package `mkdocs-material`, but it's used for the official documentation. Thus, it should not be included in the `requirements.txt` packaged with `mkdocs-material`.

PR #2079, opened 2020-11-30
update spotify URL

stevemar:
It seems like the `mkdocs-monorepo-plugin` repo has moved from https://github.com/spotify/mkdocs-monorepo-plugin to https://github.com/backstage/mkdocs-monorepo-plugin and as such the github pages link it now broken and results in a 404.

It's a known issue that when a repo is transferred between organizations and users the github repos automatically redirect but the github pages do not.

This change would update the github pages URL in the README from the `spotify` org to the `backstage` org.


squidfunk:
Thansk! The link in their repo description is still incorrect, though

stevemar:
@squidfunk yep, i noticed. i filed an issue: https://github.com/backstage/mkdocs-monorepo-plugin/issues/33

PR #2080, opened 2020-11-30
Include main and master branches in the sample workflow

stevemar:
As discussed in #2050 it was suggested to create a PR that updates the documentation to include both the `main` and `master` branches. (https://github.com/squidfunk/mkdocs-material/pull/2050#issuecomment-735358616)

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -31,6 +31,7 @@ contents:
       push:
         branches:
           - master
+          - main
     jobs:
       deploy:
         runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. I'm undecided whether to add the notice regarding `main`/`master`. Low/No-tech users should not notice any difference, as it's likely that there will only be a `master` or `main` and tech-savvy users should be able to know the difference.

Thus I'm all for removing the Admonition. We can add it if users trip over this, but otherwise, it's better to keep it as concise as possible.

stevemar:
@squidfunk makes sense. admonition removed.

PR #2105, opened 2020-12-11
Add PWA manifest generation

Stanzilla:
As a potential successor to https://github.com/squidfunk/mkdocs-material/pull/1405, starting with just adding a PWA manifest. Would love to hear options about how to handle color and icons.

### DIFF
--- a/material/manifest.json
+++ b/material/manifest.json
@@ -1,0 +1,10 @@
+{
+  "name": "mkdocs-material",
+  "short_name": "mkdocs-material",
+  "orientation": "any",
+  "display": "standalone",
+  "start_url": "/?utm_source=web_app_manifest&utm_medium=web_app_manifest&utm_campaign=web_app_manifest",
+  "description": "A Material Design theme for MkDocs",
+  "theme_color": "",
+  "background_color": ""
+}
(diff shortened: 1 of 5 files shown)


material/manifest.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "mkdocs-material",
```
squidfunk, line 2: As I understand, everbody who will be using the theme will have these values set? Wasn't the fact that this is not configurable the reason why we rejected a native PWA integration in the first place?

Stanzilla: Yeah! So, I mainly put that as a placeholder for now, need to figure out a way to get the config values from mkdocs.yml into here. If that is not at all possible, then I need to rethink.

squidfunk: A plugin may be a better approach than to encode this into the theme, as you would have to jump through a lot of hoops to extra information from `mkdocs.yml` in anything other than `*.html` and `*.xml` files.

squidfunk:
Is this still relevant? Would you like to re-open this PR when you managed to address the issues discussed?

Stanzilla:
Yeah, third time might be the charm then :p

PR #2106, opened 2020-12-11
Enable esModuleInterop for better type checking support

Stanzilla:
Motivation: ¯\_(ツ)_/¯

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -21,18 +21,18 @@
  */
 
 import * as CopyPlugin from "copy-webpack-plugin"
-import * as EventHooksPlugin from "event-hooks-webpack-plugin"
+import EventHooksPlugin from "event-hooks-webpack-plugin"
 import * as fs from "fs"
 import { minify as minhtml } from "html-minifier"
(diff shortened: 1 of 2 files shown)


webpack.config.ts
```diff
@@ -224,7 +224,7 @@ export default (_env: never, args: Configuration): Configuration[] => {
         ...args.watch ? [] : [
 
           /* FontAwesome icons */
-          new CopyPlugin({
+          new CopyPlugin.default({
```
squidfunk, line 227: That is no idiomatic JS/TS 😅

squidfunk (changes requested):

squidfunk:
Why would we want `esModuleInterop` to be activated? It's meant as a last resort when working with code that would not work otherwise. Everything is working as expected. I use it very sparingly in my projects.

Stanzilla:
Just for type checking, really. I mostly played around with it and figured why not, hence the PR note being a shruggy :)

squidfunk:
Well, as noted I'd rather not enable `esModuleInterop` as it may increase the bundle size unless absolutely necessary.

PR #2126, opened 2020-12-14
Update es.html

javierobcn:
Added missing translations in spanish language


src/partials/language/es.html
```diff
@@ -34,15 +34,21 @@
   "meta.source": "Fuente",
   "nav.title": "Navegación",
   "search.config.lang": "es",
+  "search.config.pipeline": "trimmer, stopWordFilter",
+  "search.config.separator": "[\s\-]+",
   "search.placeholder": "Búsqueda",
   "search.reset": "Limpiar",
+  "search.result.initializer": "Initializing search",
```
squidfunk, line 41: Should we also translate this key?

src/partials/language/es.html
```diff
@@ -34,15 +34,21 @@
   "meta.source": "Fuente",
   "nav.title": "Navegación",
   "search.config.lang": "es",
+  "search.config.pipeline": "trimmer, stopWordFilter",
```
squidfunk, line 37: These keys are not necessary, as they're inherited from `en.html`.

squidfunk (changes requested):
Thanks! Great to see the existing translations catching up. Could you address the requested changes?

src/partials/language/es.html
javierobcn: yes, done, thanks

src/partials/language/es.html
javierobcn: Done, thanks

squidfunk (approved):

PR #2133, opened 2020-12-15
Add support for multiple architectures

dhoppeIT:
I just stumbled across the following GitHub action:
* https://github.com/mhausenblas/mkdocs-deploy-gh-pages

Unfortunately, your Docker image is only built for the linux/adm64 architecture. 

This pull request ensures that your Docker image is also built for the linux/arm64 and linux/arm/v7 architectures. This makes it possible, to run a self-hosted runner on a Raspberry PI. 
* https://github.com/docker/build-push-action#multi-platform-image

In conjunction with the GitHub action mentioned above, this would automatically generate the docs and make them available via GitHub Pages.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -69,6 +69,12 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v1
         with:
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. Is there a reason one would want to run the Material Docker image on a raspberry pi?

dhoppe:
I honestly do not know. Maybe for non-public documentation?

In any case, it would be the easiest variant, because the following GitHub Action does not support arm64 or arm/v7.
* https://github.com/squidfunk/mkdocs-material/blob/master/.github/workflows/documentation.yml#L39-L42

Besides, I have not had my hands on a new MacBook yet ;)

squidfunk:
Yes, the new MacBooks might be a reason to include this, but I'm not seeing Raspberry Pi as a use case. Note that this image is provided for writing, not hosting the docs. We shouldn't slow down the Docker build, compiling for architectures that are rarely or never used. The Dockerfile is part of the repo, so it can easily be compiled and/or extended, if additional architectures are necessary.

Regarding the MacBook I'll leave this PR open until somebody can confirm that the current image is working. I think I read that Docker support just landed a few days ago.

squidfunk:
I'm closing this issue until somebody reports that this is necessary for the new M1 chip. I don't see any other cases that make it necessary to provide those images. Furthermore, building the `Dockerfile` which is contained in this repository for another architecture is easy enough, so if somebody needs this for a Raspberry Pi, it's straight forward to build.

PR #2147, opened 2020-12-23
Fix documentation

jaguililla:
I had to delete the `overrides` directory from the template declaration inside the page front matter for the site to build correctly.

In my case that directory didn't had the default name and maybe that was the cause of the error.

However, skipping the directory and leaving only the template file fixed the build of my site (`hexagonkt/hexagon`).

### DIFF
--- a/docs/deprecations.md
+++ b/docs/deprecations.md
@@ -98,7 +98,7 @@ templates can be shared among multiple pages:
 
     ``` markdown
     ---
-    template: overrides/hero.html
+    template: hero.html
     ---
     ```
 


squidfunk:
Thanks for your PR. However, the template URL is correct, as it's consistent with everything that is written in regard to [overrides in the documentation](https://squidfunk.github.io/mkdocs-material/customization/#extending-the-theme). Sure, the name of the `overrides` directory can be custom, e.g. `theme` or `custom`, but we have to follow a convention so it's easier to understand where everything goes.

I understand that you had troubles customizing and/or migrating, but we can't change it, as we'd lose on consistency. It's likely that this problem is related to your folder structure after all, which may not be the one recommended in the documentation.

jaguililla:
Ok, I'll discard the change then, thanks for your feedback!

PR #2206, opened 2021-01-11
Remove zero width space U+200B in from code snippets

fjp:
When I tried to use `pymdownx.snippets`, and copied the code from the [adding-a-glossary](https://squidfunk.github.io/mkdocs-material/reference/abbreviations/#adding-a-glossary), the abbreviation docs wasn't included.
Instead after serving the site locally I got no abbreviation underlines in the text followed by

```
--8<--​ "includes/abbreviation.md"
```

I then realized that there is a "zero width space" character (U+200B) after the `--8<--` which prevented the snippets extension from correctly parsing the include. 

Notice the red dot after `--8<--` in the following screenshot:

mkdocs-char

To see the hidden character I also used: https://www.soscisurvey.de/tools/view-chars.php

image

This PR removes those "zero width space" characters.

### DIFF
--- a/docs/reference/abbreviations.md
+++ b/docs/reference/abbreviations.md
@@ -81,7 +81,7 @@ _Example_:
     ```` markdown
     The HTML specification is maintained by the W3C.
     
-    --8<--​ "includes/abbreviations.md"
+    --8<-- "includes/abbreviations.md"
     ````
 
 === "includes/abbreviations.md"
--- a/docs/reference/code-blocks.md
+++ b/docs/reference/code-blocks.md
@@ -311,7 +311,7 @@ _Example_:
 
 ```` markdown
 ```
---8<--​ ".browserslistrc"
+--8<-- ".browserslistrc"
 ```
 ````
 


facelessuser:
That may be intentional to keep it from triggering and including snippets. You can use snippets even in code blocks as snippets are run very early on.

facelessuser:
I am guessing you didn't run the documents and view how it looks after you made this change?

fjp:
Up until now, sorry for that. I am getting an error when I try to run `mkdocs serve`:

```
INFO    -  Building documentation... 
INFO    -  Cleaning site directory 
ERROR   -  Error reading page 'reference/abbreviations.md': Snippet at path includes/abbreviations.md could not be found 

Snippet at path includes/abbreviations.md could not be found
```

fjp:
I guess the way you did it [here](https://facelessuser.github.io/pymdown-extensions/extensions/snippets/#snippets-notation) is the way to go:

```
<pre><code>--8&lt;-- "filename.ext"</code></pre>
```

facelessuser:
Yeah, I did this thing where if you add a space at the end, it won't trigger, but then people would copy and paste it as an example without really reading about the whitespace part and then complain it doesn't work, so then I did it like you pointed out so they could copy it. It was just easier. It's a syntax I'd never really use anywhere else except to insert snippets (and obviously the one-time case where I explain how snippets work 🙂).

squidfunk:
Oh my 🤦‍♂️ I thought I'd be clever tricking the Snippets extension into not expanding the macro, but obviously haven't thought about users copy & pasting the code and wondering why it doesn't work. A better way is of course to just use the literal HTML, so Snippets won't expand it. 

However, I've found no way to make this work inside a tabbed container without messing up the formatting. If somebody could make this work without the zero-width whitespace, I'm happy to merge.

Possible explanation for #2154.

facelessuser:
I created a working pull at #2207

squidfunk:
Superseded by 6cb4c86c, see discussion in #2207.

PR #2207, opened 2021-01-12
Remove zero width whitespace from snippets examples

facelessuser:
Fixes #2206

### DIFF
--- a/docs/reference/abbreviations.md
+++ b/docs/reference/abbreviations.md
@@ -78,11 +78,11 @@ _Example_:
 
 === "docs/page.md"
 
-    ```` markdown
-    The HTML specification is maintained by the W3C.
-    
-    --8<--​ "includes/abbreviations.md"
-    ````
+    <div class="highlight">
+    <pre><code>The HTML specification is maintained by the W3C.
+
+    --8&lt;-- "includes/abbreviations.md"</code></pre>
+    </div>
 
 === "includes/abbreviations.md"
 
--- a/docs/reference/code-blocks.md
+++ b/docs/reference/code-blocks.md
@@ -309,11 +309,11 @@ directly into your project documentation.
 
 _Example_:
 
-```` markdown
-```
---8<--​ ".browserslistrc"
-```
-````
+<div class="highlight">
+<pre><code>```
+--8&lt;-- ".browserslistrc"
+```</code></pre>
+</div>
 
 _Result_:
 


squidfunk:
Thanks for the PR! Unfortunately, it doesn't entirely fix the issue, at least "on my machine":

Bildschirmfoto 2021-01-12 um 16 42 26

Somehow the snippets code breaks onto the next line. Also, there's one too many `l` in `highllight`.

facelessuser:
Can you give me details? I can fix the class, but I'm not seeing your issue:

Screen Shot 2021-01-12 at 8 50 35 AM

Let me update some dependencies and see what I get.

facelessuser:
Hold up, I am seeing this issue. It's probably Markdown's new HTML parser...give me a second

squidfunk:
Hmm, weird. I tested locally and it didn't work. Let's just merge it and see what happens then 😅

facelessuser:
Hmm, well, this, unfortunately, looks like an issue with how HTML is handled under these type of structures, this affects Admonitions and others based on the same logic (which both Details and Tabbed are).

I imagine it is a problem with how the blocks are fed to the parser while being under such structures. 

Basically, if you remove the empty line in the code block, it works. This seems to be a problem in 3.2 as well which is prior to the new HTML parser.

Sigh, this is kind of annoying. I'm surprised I've never run into this issue before.

squidfunk:
If you find out that it doesn't work with the new Markdown parser, we can also disable Snippets for the official docs - I'm not using it anyway, only [here](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#embedding-external-files) and that's easily replaced.

facelessuser:
The issue is not with Snippets. It works as expected. The problem is how blocks are parsed under Admonitions and Admonition like structures. Blocks are fed to the parser (blocks are determined by empty new lines, something that causes frustration in other places as well). So if you break the HTML up this way, and feed those pieces one at a time, you get HTML fragments with incomplete context. This causes them to get parsed odd.

squidfunk:
I understand. However, what I meant is that we can mitigate the issue by disabling Snippets. I did exactly that in 6cb4c86c, so this PR is no longer necessary.

facelessuser:
I see. Yes, that mitigates the issue.

I file an issue in Markdown later today. Hopefully, we can solve the issue there and then port relevant fixes down to pymdown-extensions.

squidfunk:
Perfect, thanks!

Issue #2209, opened 2021-01-15
Syntax error in sv.html


allmightyse:
There is an syntax error in: 
material\partials\languages\sv.html", line 30, in template
    "tabs.title"; "Flikar",
jinja2.exceptions.TemplateSyntaxError: expected token ':', got ';'

It would be nice if that cloud been fixed soon...

squidfunk:
Confirmed, would you like to create a PR?

allmightyse:
I can do that, but in that file is Auto generated? And dont now where thats is auto generated from?

allmightyse:
Now there is an PR. Would be nice if that can be merged.

squidfunk:
Released as part of 6.2.5.

PR #2210, opened 2021-01-15
Update sv.html to correct syntax error

allmightyse:
#2209 Correct syntax error from ; to :


squidfunk:
Thanks! I forgot to say, please also update `src/partials/languages/sv.html`, or the change will be overwritten on the next build (see the note on [automatically generated files](https://squidfunk.github.io/mkdocs-material/customization/#development-mode))

allmightyse:
Aha, then that was done.

squidfunk:
Thanks!

PR #2213, opened 2021-01-16
Optimize nested navigation rendering: don't cross template boundary

oprypin:
The number of calls to sub-templates when rendering the navigation is O(n^2) relative to the number of pages. This is the only such instance, which is why I think this is worth optimizing.

The optimization here doesn't improve the complexity, it just removes the overhead of instantiating a new Jinja template for each of these calls.

E.g. for 710 pages on the site, the number of calls to Jinja's `get_template` (implying `new_context` and others) crosses a million ==(710**2)*2. They're not expensive but also not super cheap, and add up to a big percentage of overall site build times.

### DIFF
--- a/material/partials/nav-item.html
+++ b/material/partials/nav-item.html
@@ -2,6 +2,7 @@
   This file was automatically generated - do not edit
 -#}
 {% set features = config.theme.features or [] %}
+{% macro do_nav_item(nav_item, path, level) %}
 {% set class = "md-nav__item" %}
 {% if nav_item.active %}
   {% set class = class ~ " md-nav__item--active" %}
@@ -31,9 +32,7 @@
       <ul class="md-nav__list" data-md-scrollfix>
         {% set base = path %}
         {% for nav_item in nav_item.children %}
-          {% set path = base ~ "-" ~ loop.index %}
-          {% set level = level + 1 %}
-          {% include "partials/nav-item.html"  %}
+          {{ do_nav_item(nav_item, path = base ~ "-" ~ loop.index, level = level + 1) }}
         {% endfor %}
       </ul>
     </nav>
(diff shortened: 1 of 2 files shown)


oprypin:
Additional notes:

* I have already optimized this as best as possible in MkDocs itself: https://github.com/mkdocs/mkdocs/pull/2261 - and my calculations are being done *with* that optimization, and it's still definitely worth it.

* My use case is building sites with a large amount of semi-auto-generated pages, such as https://oprypin.github.io/crystal-book/api/ (this one has 750 pages, and build times go from 35.34sec to 28sec).
  Sure, I can override this myself, but why.

* I acknowledge that MkDocs' own themes don't optimize this (https://gitter.im/squidfunk/mkdocs-material?at=5fe075e8ce40bd3cdbf31f17 -> https://github.com/mkdocs/mkdocs/blob/master/mkdocs/themes/mkdocs/nav-sub.html), but I just think that Material theme is particularly "prone" to be used with more pages than usual, so still more worth it here.

squidfunk:
Thanks for the PR! Could you remove the JS/CSS assets? I don't think they're necessary.

squidfunk:
Also, could you please make sure that all navigation features still work? Especially sections, integrated table of content etc. Please see [the navigation set up guide](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/) for an exhaustive list.

oprypin:
Sure I'll remove assets. Was just following https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md#submitting-a-pull-request-pr

oprypin:
And yes, everything in navigation features works fine.

squidfunk:
LGTM. Thanks!

squidfunk:
Released as part of 6.2.5.

PR #2254, opened 2021-01-29
Add support for the Pull GitHub app

AndreiJirohHaliliDev2006:
By default, when the [Pull GitHub app](https://github.com/apps/pull) detects new commits from the upstream repository (in most cases, from its parent) and when it merged, it'll do a hard reset, causing data loss in some cases.


squidfunk:
Thanks for providing this. However, I'm wondering if this is a sound default applied to all forks? I'd favor some discussion before integrating this into the main repository - it may be better situated in the forks.

squidfunk:
Closing for the reasons stated - this is something the forking party should decide on.

PR #2262, opened 2021-01-31
Update pl translations

lenisko:
I'm not sure about `search.config.*` so didn't touched it. Rest should be up to date.


jaceklaskowski (approved):
LGTM (a Polish native here)

Issue #2152, opened 2020-12-26
Accent current section in integrated ToC


danierutu:
- [x] I've read the [contribution guidelines][1] and agree with them

__I want to suggest an idea and checked that ...__

- [x] ... to my best knowledge, my idea wouldn't break something for other users
- [x] ... the documentation does not mention anything about my idea
- [x] ... there are no open or closed issues that are related to my idea

## Description

Using the regular Table of Contents (right sidebar), as you scroll through the page, the current section in the viewport is highlighted with the accent color, and the sections you leave behind are greyed out. However, using the integrated ToC (from Prairie Fire), it's just static. It'd be cool if it behaved like the regular one.

  [1]: https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md

squidfunk:
Thanks for reporting. Yes, the labels should definitely be blurred/highlighted like when the table of contents is on the right.

squidfunk:
So fixing this is quite tricky, as currently there can only be one component with the same name due to the way the observable architecture is designed. I want to fix this in its entirety, but that may take some time. I'm preparing a refactoring of how components are detected and mounted which is necessary to attach the same behavior to multiple components. Once this is done, fixing this issue is a no-brainer.

squidfunk:
I started with the refactoring and solved the issue with the table of contents:

https://user-images.githubusercontent.com/932156/105685510-70a8ab80-5ef6-11eb-87fc-9b8fe3d9e97d.mp4

It's still on a branch and not ready for merge yet because it's a pretty serious refactoring I'm currently undertaking that will simplify the observable architecture _a lot_. I expect it to be ready in the coming weeks. The refactoring is necessary as I've seen some downsides of the current architecture when iterating on new features. The new, simpler, architecture will allow us to achieve a higher velocity iterating on new features.

danierutu:
Wow, that's fantastic. I didn't expect that such an apparently simple fix would actually take that amount of work in the backend. You never stop impressing me! Keep it up, and don't rush.

squidfunk:
7.0.0b1 was just released, which includes the fixes. What's even better is that now, on mobile, the anchors are also blurred:

<img width="521" alt="Bildschirmfoto 2021-02-12 um 18 01 17" src="https://user-images.githubusercontent.com/932156/107798389-56772600-6d5c-11eb-858f-15aa218e4510.png">

facelessuser:
[...]

PR #2273, opened 2021-02-07
WIP – Material for MkDocs 7.x

squidfunk:
This branch contains a rewrite of the observable architecture, which will mitigate a lot of pain when maintaining and extending Material for MkDocs. The current architecture is too abstract to be useful and has several problems, e.g. there can be at most one component of each kind, resulting in #2152. The new architecture doesn't have these shortcomings and will even probably yield an (even) smaller payload.

This branch will sit here until the refactoring has been completed. Then, a beta release will be issued which will have the same functionality as Material 6.2.x in order to squash some bugs before the final release.


Stanzilla:
feature request: allow disabling search while leaving searchbox intact, for use with 3rd party search engines like Algolia etc.

squidfunk:
The refactoring is finished. The behavior should be the same as before, except that the complete architecture has been refactored and now allows for an arbitrary number of components. More in a dedicated issue.

PR #2285, opened 2021-02-10
Replace deprecated -webkit-details-marker

Stanzilla:
[Deprecation] ::-webkit-details-marker pseudo element selector is deprecated. Please use ::marker instead. See https://chromestatus.com/feature/6730096436051968 for more details.


squidfunk:
Thanks!

PR #2362, opened 2021-02-28
Add `extracopyright` block to footer.html

seandstewart:
I'm using a few assets for a website that require I credit the source in the licensing. This adds a block to the footer to allow for adding content below the `Material for MkDocs` accreditation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@
 /node_modules
 /__pycache__
 /venv
+/.venv
 
 # Build files
 /build
@@ -62,3 +63,4 @@ tmp
 
 # IDEs
 .vscode
+.idea
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -178,26 +178,27 @@ Then, e.g. to override the site title, add the following line to `main.html`:
 
 Material for MkDocs provides the following template blocks:
 
-| Block name   | Wrapped contents                                |
-| ------------ | ----------------------------------------------- |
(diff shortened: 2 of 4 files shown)


squidfunk:
That's actually a great idea, as I've see this use case quite a few times. Is `extracopy` a good name? Maybe `extracopyright`(albeit longer), as this would mirror the key in `mkdocs.yml` (i.e. `copyright`).

seandstewart:
Good call - it's more explicit what the intended use-case is.

squidfunk:
🤦‍♂️ I forgot about this PR. I'm merging it now and release it with the next version. Sorry for the wait!

EdouardThiel:
Hello,

`extracopyright` doesnt work for me:

- I have added `custom_dir: overrides` after the material theme in `mkdocs.yml`
- also added a `overrides/main.html` file, in which after `{% extends "base.html" %}` I redefine the block `extracopyright`

But it outputs nothing, because the block `footer` includes "partials/footer.html" and currently, includes doesn't override blocks.

(if I redefine the block `htmltitle` it works, this one is defined in base.html and not included in a sub-file)

Is it possible to correct the templates with a variable or a macro instead of a block `extracopyright`? or a special way to import footer.html ? 

For instance in footer.html

```
        Made with
        <a href="https://squidfunk.github.io/mkdocs-material/" target="_blank" rel="noopener">
          Material for MkDocs
        </a>
        {{ extracopyright }}
```

PR #2363, opened 2021-02-28
Support the Google analytics 4

cainmagi:
Currently, the theme does not support the newest Google analytics 4 API. I provide an option for enabling it in the config file. In `mkdocs.yml`, add the following line:

```yaml
# Google Analytics
google_analytics_4:
  - <your-google-analytics-4-ID>
```

When setting `google_analytics_4`, the original configs `google_analytics` would not be used.

### DIFF
--- a/material/base.html
+++ b/material/base.html
@@ -71,8 +71,12 @@
       <link rel="stylesheet" href="{{ path | url }}">
     {% endfor %}
     {% block analytics %}
-      {% if config.google_analytics %}
-        {% include "partials/integrations/analytics.html" %}
+      {% if config.google_analytics_4 %}
+        {% include "partials/integrations/analytics-4.html" %}
+      {% else %}
+        {% if config.google_analytics %}
+          {% include "partials/integrations/analytics.html" %}
+        {% endif %}
       {% endif %}
     {% endblock %}
     {% block extrahead %}{% endblock %}
--- a/material/partials/integrations/analytics-4.html
+++ b/material/partials/integrations/analytics-4.html
@@ -1,0 +1,6 @@
+{#-
+  This file was automatically generated - do not edit
+-#}
+{% set analytics_4 = config.google_analytics_4 %}
+<script async src="https://www.googletagmanager.com/gtag/js?id={{ analytics_4[0] }}"></script>
+<script>window.dataLayer=window.dataLayer||[];function gtag(){dataLayer.push(arguments);} gtag('js',new Date());gtag('config','{{ analytics_4[0] }}');</script>
--- a/src/base.html
+++ b/src/base.html
@@ -143,8 +143,12 @@
 
     <!-- Analytics -->
     {% block analytics %}
-      {% if config.google_analytics %}
-        {% include "partials/integrations/analytics.html" %}
+      {% if config.google_analytics_4 %}
+        {% include "partials/integrations/analytics-4.html" %}
+      {% else %}
+        {% if config.google_analytics %}
+          {% include "partials/integrations/analytics.html" %}
+        {% endif %}
       {% endif %}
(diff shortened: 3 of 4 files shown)


squidfunk:
Thanks for your pull request. I'm sorry, but I have to ask: have you tested this code? I'd be surprised if it worked, because in general, `mkdocs.yml` doesn't allow for adding custom code to the top-level of the configuration file. Everyting custom needs to be added to `theme.*` or `extra.*`, which is also the reason why this hasn't been addressed yet, as this would mean redundant configuration.

There are several issues here that discuss GA4 / GTAG. In essence, we'd need to wait for https://github.com/mkdocs/mkdocs/issues/2252 to be merged.

cainmagi:
> Thanks for your pull request. I'm sorry, but I have to ask: have you tested this code? I'd be surprised if it worked, because in general, `mkdocs.yml` doesn't allow for adding custom code to the top-level of the configuration file. Everyting custom needs to be added to `theme.*` or `extra.*`, which is also the reason why this hasn't been addressed yet, as this would mean redundant configuration.
> 
> There are several issues here that discuss GA4 / GTAG. In essence, we'd need to wait for [mkdocs/mkdocs#2252](https://github.com/mkdocs/mkdocs/issues/2252) to be merged.

Thank you! Actually, I am not very familiar with this theme. However, today I use my modification to create a new GA4 application, and I could confirm that this modification works on my website:

image

I am just following your instructions on [this page](https://squidfunk.github.io/mkdocs-material/creating-your-site/). The modified `mkdocs.yml` locates in the root path of my source folder.

```
.
├─ docs/
│  └─ index.md
└─ mkdocs.yml
```

If you feel that my implementation has mistakes, you could close this this issue. I am looking forward to see an official implementation for supporting GA4.
Thank you!

squidfunk:
Thanks for responding so quickly. I'd rather wait for the PR to be merged which will deprecate the current GA implementation. When this is released, we can add support for GA4, as well as GTAG.

cainmagi:
> Thanks for responding so quickly. I'd rather wait for the PR to be merged which will deprecate the current GA implementation. Your implementation is non-standard in respect to MkDocs and may have unintended consequences. When this is released, we can add support for GA4, as well as GTAG.

I understand. Thank you!

squidfunk:
`gtag.js` support just landed in `master` and will be released soon. Feedback appreciated Now, both formats are supported.

__Google Analytics__

``` yaml
extra:
  analytics:
    provider: google
    property: UA-XXXXXXXX-X
```

__Google Analytics 4__ (via `gtag.js`)

``` yaml
extra:
  analytics:
    provider: google
    property: G-XXXXXXXXXX
```

cainmagi:
Thank you! I will keep track of this.

squidfunk:
As there was confusion among users whether this was already released - no, it's not part of 7.1.7. To test-drive the new analytics integration, install Material for MkDocs from git like so:

```
pip install git+https://github.com/squidfunk/mkdocs-material.git
```

squidfunk:
7.1.8 was just released, which contains the changes.

sorenwacker:
Is there a way to verify that this part is working? When I open the source code of the generated page, what element should I look for? I do not see the provided google ID anywhere. And google analytics also does not start tracking.

sorenwacker:
Tried `mkdocs-material` 7.1.8 and 7.3.0 (?).

squidfunk:
`analytics.js` should be loaded, see [this partial](https://github.com/squidfunk/mkdocs-material/blob/master/src/partials/integrations/analytics/google.html). I can verify that it works, so it might be related to your setup or your customizations, if any.

sorenwacker:
Now it works. ;-)

Issue #2370, opened 2021-03-01
[BUG]: Logo is always trying to load `.icons/<path>.svg`


polarathene:
- [x] I've read the [contribution guidelines][1] and agree with them

__I've found a bug and checked that ...__

- [x] ... the problem doesn't occur with the default MkDocs template
- [x] ... the problem is not in any of my customizations (CSS, JS, template)
- [x] ... the documentation does not mention anything about my problem
- [x] ... there are no open or closed issues that are related to my problem

## Description

I'm having trouble getting the [logo feature](https://squidfunk.github.io/mkdocs-material/setup/changing-the-logo-and-icons/#logo) to work.

I have no problem with the `favicon` which uses a path in the same directory. I've tried both SVG and PNG but neither work.

I've tried providing the path without the SVG extension and placing it in a `.icons/` dir outside of `docs_dir` relative to `mkdocs.yml`, within `docs_dir` and within `theme.custom_dir`, only `theme.custom_dir='theme'` with `theme/.icon/dm-logo.svg` and `icon.logo: dm-logo` (without the `.svg` extension) resulted in a successful build.

If I prepend the `.icons/` path prefix to test the other two `.icons/` folders I get failure with `.icons/.icons/dm-logo.svg` output in the console. This doesn't improve by removing the dot from the directory names to become `icons/`, nor with `.png` extension, which outputs `.icons/icons/dm-logo.png.svg`

Further more the SVG itself with the successful case was not visible, it was loaded but not scaled, instead the CSS applied just resulted in a clipped top-left section of the SVG viewbox, which made it look like it was empty. Inspecting the DOM shows the SVG was loaded, just clipped out of view. Additionally the `fill: currentColor` CSS flattened out some SVG details.

### Expected behavior

Supply a logo graphic to use and it should work.

### Actual behavior

A: Fails to run `mkdocs serve`, seems to try load an internal graphic instead.
B: Loads `.icons/dm-logo.svg` if in the `custom_dir` `.icons/` subdirectory and no extension is provided. SVG is not scaled, thus is clipped from view. Docs don't provide expected size constraints for SVG logo.

### Steps to reproduce the bug

1. Add an image for logo.
2. Update config in `mkdocs.yml` to use that image resource.
3. Try `mkdocs serve` (I'm just running the container), process exits with no error but prints out `.icons/assets/dm-logo.png.svg
`

### Package versions

Current Docker image release (7.x)

### Project configuration

```yaml
[...]

PR #2375, opened 2021-03-01
docs: Better clarify logo setup options

polarathene:
In response to [troubleshooting an issue that I misunderstood as a bug](https://github.com/squidfunk/mkdocs-material/issues/2370) (but was the fault of reading docs at 3am).

Related, another user ran into a similar problem: https://github.com/squidfunk/mkdocs-material/discussions/2167

---

Better clarifies the distinct path difference (as they share a very similar but slightly different path), which can be accidentally missed triggering either an unhelpful "error" message to the console: `.icons/<logo value>.svg`.

Additionally switches the order of the tabs to the more common implicit use-case of providing a logo resource vs a more explicit decision to use a bundled theme SVG icon for a logo (which is handled differently compared to `theme.logo` such as scaling).

### DIFF
--- a/docs/setup/changing-the-logo-and-icons.md
+++ b/docs/setup/changing-the-logo-and-icons.md
@@ -16,11 +16,19 @@ additional icons][1] with minimal effort.
 ### Logo
 
 [:octicons-file-code-24: Source][2] ·
-:octicons-milestone-24: Default: `material/library`
+:octicons-milestone-24: Default: [`material/library`](https://github.com/squidfunk/mkdocs-material/blob/master/material/.icons/material/library.svg)
 
-There're two ways to specify a _logo_: it must be a valid path to [any icon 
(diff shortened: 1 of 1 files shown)


polarathene:
I realize now that this theme doesn't seem to do anything with python itself. Just the Jinja templates. So the "error" message isn't something this project is able to address, nor is it likely something upstream would either as the custom logo is a theme feature leveraging the `mkdocs.yml` theme metadata and jinja templating to source it.

AFAIK that would require handling in the python logic at build time for jinja template processing, but as it's generic for themes you're not able to tailor the message appropriately.

SeanTAllen:
I ran into this problem when using mkdocs-material for the first time yesterday. It took me a while to realize that I needed logo "not under icon", so I personally think that this is an excellent clarification that would have saved me about 10 minutes of futzing and searching.

PR #2391, opened 2021-03-06
Fix bug in instant navigation preprocess function

jbms:
This still isn't a correct algorithm for finding the real longest
common prefix, since it only looks at the first two elements, but at
least it is more correct than it was.

### DIFF
--- a/src/assets/javascripts/integrations/instant/index.ts
+++ b/src/assets/javascripts/integrations/instant/index.ts
@@ -104,22 +104,27 @@ function preprocess(urls: string[]): string[] {
   if (urls.length < 2)
     return urls
 
-  /* Compute references URLs */
-  const [root, next] = urls.sort((a, b) => a.length - b.length)
-
-  /* Compute common prefix */
-  let index = 0
-  if (root === next)
-    index = root.length
-  else
-    while (root.charCodeAt(index) === root.charCodeAt(index))
-      index++
+  /* Compute longest common prefix */
+  let prefix = urls[0]
+  for (let url of urls) {
+    // Exclude the portion of url after the last slash.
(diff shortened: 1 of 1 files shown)


jbms:
This is primarily relevant if you turn off directory urls.  Otherwise the existing code should work, but it is just choosing the shortest, it isn't actually finding the longest common prefix, and it also doesn't take / into account.

For example, currently if you have http://server/index.html and http://server/index1.html it will pick http://server/index.html as the prefix which is not a common prefix.  The longest common prefix (not treating / specially) is http://server/index but what we actually want is http://server/, which is the longest common prefix of the parent directory paths.

squidfunk:
Thanks for providing more information! However, I haven’t encountered a case where this happens with MkDocs. Could you post a minimal `mkdocs.yml` that will generate a sitemap which induces this problem?

The algorithm in-place uses some knowledge of the sitemap, it’s by no means a general LCP implementation, but that wasn’t necessary up to this point.

jbms:
To reproduce this issue, you can modify the mkdocs.yml for this repository's own documentation to have use_directory_urls = false and enable navigation.instant.

Then build/serve the documentation from a location that does *not* match the site_url, in order for the preprocess logic to be necessary.  For example, run `mkdocs serve` and then open `http://localhost:8000` rather than `http://127.0.0.1:8000`.  Or just `mkdocs build` and then serve from a local webserver.

You will observe that instant navigation does not work.

squidfunk:
Thanks again for reporting and providing the reproduction. 1d790c90 fixes instant loading for when directory URLs are enabled. I opted for a simpler solution using a regex, but it's essentially the same. Tested it with and without directory URLs.

PR #2392, opened 2021-03-06
Fix navigation.instant to correctly handle version dropdown

jbms:
Without this change, on each instant navigation, an additional version
dropdown is added but the old one also remains in the header, leading
to multiple version dropdowns.

### DIFF
--- a/src/assets/javascripts/integrations/instant/index.ts
+++ b/src/assets/javascripts/integrations/instant/index.ts
@@ -276,7 +276,9 @@ export function setupInstantLoading(
 
           /* Components */
           "[data-md-component=announce]",
-          "[data-md-component=header-topic]",
+          // Replace header-title rather than header-topic in order to avoid problem with
+          // versioning.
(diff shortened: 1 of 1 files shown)


jbms:
To reproduce, turn on instant navigation and version support (use e.g. your versioning example repo).  Add a versions.json file (doesn't matter if the versions listed actually exist).  Then follow an instant navigation link and observe that there is now a duplicate version drop-down next to the original one.

squidfunk:
Thanks for providing the reproduction. Unfortunately, this reproduction will yield to flickering of the version selector when instant navigation kicks in, which is the reason why the site title was put into its own components called `header-topic`. I've got another idea how to fix this, though.

squidfunk:
Fixed in 526fbed3. The `header-title` component was in the wrong group of components.

squidfunk:
d17c488b fixes a regression after the last commit where the header title was not updating correctly. This should now be a more robust implementation.

PR #2393, opened 2021-03-06
Fix instant navigation when clicking on the logo button and logo is svg

jbms:


jbms:
To reproduce, turn on instant navigation for this repository's own navigation, and modify site_url to match the development server (otherwise the home link will point elsewhere and won't be eligible for instant navigation anyway).   The logo is already svg.

PR #2415, opened 2021-03-08
Update css for custom admonition example

Andre601:
It seems to be required to set a `border-color: <color>;` for `.md-typeset .<class> > .adminition-title, .md-typeset .<class> > summary` to make it work with details or it will otherwise display the primary fg color for the left border.

If there is a better way to this, let me know and I'll update this PR.


squidfunk:
Yes, the Admonition code changed to include the border as part of the title to correct for the keyboard focus outline 😊 Thanks for providing the fix!

PR #2421, opened 2021-03-09
remove 'shortcut' from favicon meta tag

coliff:
This isn't needed and was never part of the spec.
Great article here: https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs#did-we-forget-anyone

Also:
https://mathiasbynens.be/notes/rel-shortcut-icon

### DIFF
--- a/src/base.html
+++ b/src/base.html
@@ -58,7 +58,7 @@
       {% endif %}
 
       <!-- Favicon -->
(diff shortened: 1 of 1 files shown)


squidfunk:
I guess I copied it from somebody, who copied it from somebody else. Awesome find, happy to remove it! However, this may impact instant loading. I'll come back to it when I find some time.

Issue #1408, opened 2019-12-25
Link to return to the top of the page


majkinetor:
## Description

I would like to be able to click on "Table of contents" item to return to the top of the page. Currently this is not possible.

### Use Cases

This can be used on every page.

### Details

Typical markdown page looks like this:

```md
# Title

Intro text

## Subsection 1
....
## Subsection 2
...
```

This produces TOC like this:

Table of contest:
- Subsection 1
- Subsection 2

That is, title is missing. If intro is long, you have no option to return to the top of the page via TOC but must click first item then scroll up.

Check out sample here: 
* https://majkinetor.github.io/mm-docs-template/demo

Everything up to the `Customization` section is not reachable via TOC.

squidfunk:
Not a bad idea, but not that trivial. The `toc.html` partial is also used to render the TOC inside the mobile navigation and so the `Table of contents` title is actually a label which triggers the TOC inside the mobile navigation. For this to work we would need to duplicate the label with an `a` element and only show the latter on `screen+`. Needs some work on the CSS and some testing.

majkinetor:
Perhaps something else could be done that is trivial and has same effect. Would adding extra element be easier ?

```
Table of contest [^]
   [Subsection 1]
   [Subsection 2]

[^] Table of contest 
   [Subsection 1]
   [Subsection 2]

Table of contest
   [^]
   [Subsection 1]
   [Subsection 2]

```

Stanzilla:
If it gets too complicated, I have a to top scroll button in my fork, I could PR that as well.

majkinetor:
That is not intuitive - one button to go to top and TOC for everything else. Its better then nothing certainly.

squidfunk:
I like the idea with the clickable TOC, but I don't know if it is more obvious that the go-to-top button. I also thought about switching __Table of contents__ with the contents of the `h1`, but then it's less obvious that it's a TOC. The reason why the `h1` is excluded is that everything under the `h1` would need to be indented to be consistent, and that made no sense to me. However, I'm open to suggestions.

Stanzilla:
I have the extra button in the bottom right and it works quite well for mobile

majkinetor:
>The reason why the h1 is excluded is that everything under the h1 would need to be indented to be consistent, and that made no sense to me.

Exactly. That thing is actually present on some other systems and maximally anoying to the point I stopped using h1 on said systems.

>However, I'm open to suggestions.

[...]

Issue #974, opened 2019-01-28
Adding link to sidebar results in JavaScript error


MrMiaoMiao:
## Description

Adding a `<a>` link on the right nav bar causes the following error message in developer console in chrome when scrolling (does not occur when scrolling down initially)
```
application.ff6f93ee.js:1 Uncaught TypeError: Cannot read property 'offsetTop' of undefined
    at e.t.update (application.ff6f93ee.js:1)
    at update_ (application.ff6f93ee.js:1)
``` 
I am using a clone of the repo and changed main as follows, all other files remain the same:
```html
{% extends "base.html" %}

{% block site_nav %}
{% if nav %}
<div class="md-sidebar md-sidebar--primary" data-md-component="navigation">
  <div class="md-sidebar__scrollwrap">
    <div class="md-sidebar__inner">
      {% include "partials/nav.html" %}
    </div>
  </div>
</div>
{% endif %}
{% if page.toc %}
<div class="md-sidebar md-sidebar--secondary" data-md-component="toc">
  <div class="md-sidebar__scrollwrap">
    <div class="md-sidebar__inner">
      <ul class="document-info-card">
        <li><a href="https://www.some website.com">Insert text here</a></li>
        <li><a href="https://www.some other website.com">Insert more text here</a></li>
      </ul>
      {% include "partials/toc.html" %}
    </div>
  </div>
</div>
{% endif %}
{% endblock %}
```
If `<a>` tags are replaced with `<div onclick="location='link'">` or anything else, there is no error reported

Reported culprit in developer console is this section:
```javascript
Blur: function() {
            function e(e) {
                this.els_ = "string" == typeof e ? document.querySelectorAll(e) : e,
                this.index_ = 0,
                this.offset_ = window.pageYOffset,
                this.dir_ = !1,
                this.anchors_ = [].reduce.call(this.els_, function(e, t) {
                    var n = decodeURIComponent(t.hash);
                    return e.concat(document.getElementById(n.substring(1)) || [])
                }, [])
            }
            var t = e.prototype;
            return t.setup = function() {
                this.update()
            }
            ,
            t.update = function() {
                var e = window.pageYOffset
                  , t = this.offset_ - e < 0;
                if (this.dir_ !== t && (this.index_ = this.index_ = t ? 0 : this.els_.length - 1),
                0 !== this.anchors_.length) {
                    if (this.offset_ <= e)
[...]

Issue #313, opened 2017-05-06
Can you add exclude a header from toc in secondary sidebar?


ngtrian:
## Description
Can you add exclude a header from toc in secondary sidebar? I want add a header 5 to content, but I don't want display header 5 in secondary sidebar (md-sidebar md-sidebar--secondary)

### Package versions

* Python: `2.7.12`
* MkDocs: `0.16.3`
* Material: `1.6.1`

facelessuser:
I believe something like this could help: https://github.com/squidfunk/mkdocs-material/issues/136, but you'd have to adjust the CSS to target your desired level.

ngtrian:
This can not solve my problem. I want to hide heading 5 not level 2 in toc

Because I want to use headings in documents in a certain file, But I do not want them to appear in TOC.

facelessuser:
That isn't what #136 means. The person wanted to hide anything greater than 2 levels deep (this isn't referring to h1).  So they wanted to show h2 and h3, so the solution hides h4+.  If I understand correctly, you want to hide h5+, so you would need to tweak the CSS to start hiding 1 level deeper.

Unless I am still misunderstanding.

ngtrian:
Simple I want to use headings in documents in a certain file, But I do not want them to appear in TOC. 
Here I use heading 5 less used to avoid overlapping with the heading of other markdown files

facelessuser:
At most, all that can be done is excluding a certain header level with CSS.  So the solution above is based on header levels (forth in the nested hierarchy), but not header type (h5). The TOC generator (which is separate from this theme), does not have the ability to exclude a specific header type.  And the TOC in the sidebar does not track the header type, but simply nests them in a logical hierarchy.

So targeting specifically h5 in the sidebar TOC (with the purpose of hiding it) is not possible, but targeting the 4th level nesting is possible.

![screenshot 2017-05-06 23 39 00](https://cloud.githubusercontent.com/assets/1055125/25778223/73ebab60-32b5-11e7-962b-4e567a1aff50.png)

```css
label[for="toc"]+.md-nav__list .md-nav__list .md-nav__list .md-nav__list {
    display: none;
}
```

But if you skipped nesting levels, for example nest h5 directly under h2, then this would not hide h5 as it would not be at the proper nested level.  So if this is not sufficient for you, I don't think what you want is possible

squidfunk:
When you specify your heading in the Markdown file not like this:

``` markdown
##### Heading
```

but like this:

[...]

PR #2426, opened 2021-03-10
Support TOC anchor targets vertical order differing from TOC order.

Benjamin-Dobell:
Prior to this PR, if the TOC anchor targets' vertical page offsets did not match the order of the TOC anchors themselves, then blur would not work correctly.

All this PR does is sort the anchors by their target's offset so that computation of the current partition (driven by the viewport) remains accurate.

The primary motivation for this change is that I've created a Python-Markdown plugin which allows you to manually insert entries into TOC: https://github.com/Benjamin-Dobell/pymdown-toc-ext.

`pymdown-toc-ext` aside, if a user was particularly adventurous, they could also run into the vertical offset issue with CSS that reorders the anchor targets / headers vertically.

P.S. pymdown-toc-ext can be used as an alternate mechanism to address #1408 #313 #974 (probably others too).

P.P.S. I read the CONTRIBUTING doc re: including build artefacts. However, I'm seeing a [_lot_ of unrelated changes](https://github.com/Benjamin-Dobell/mkdocs-material/commit/ca28879c1f1720b6c2edd7aee29f6cb141869db1) when I build. For example, > 100 SVGs have changed for some reason 🤷‍♂️

### DIFF
--- a/src/assets/javascripts/components/toc/index.ts
+++ b/src/assets/javascripts/components/toc/index.ts
@@ -161,6 +161,13 @@ export function watchTableOfContents(
         }, new Map<HTMLAnchorElement[], number>())
       }),
 
+      /* Sort index by vertical offset */
+      map(index => (
+        new Map<HTMLAnchorElement[], number>(
+          [...index].sort(([, offsetA], [, offsetB]) => offsetA - offsetB)
+        )
+      )),
+
       /* Re-compute partition when viewport offset changes */
       switchMap(index => combineLatest([adjust$, viewport$])
         .pipe(


squidfunk:
Thanks for the PR! I didn't know about this extension. Let me evaluate it and come back to you in a few days.

Benjamin-Dobell:
The extension is brand new 🙂 If you've any suggestions, I'm open for feedback.

In our case it solves an issue we were experiencing with API discoverability. Some complex APIs have their own heading (with examples), whilst other simpler APIs just appear in a table. Users have been reporting they're overlooking the existence of the APIs not in the TOC. This plugin ensures we can inject them in the TOC, and have it link back to the table row.


Thanks for all your hard work!

squidfunk:
Could you move the sorting in a separate `map` operator function that comes after the one used in this PR? The code is already very complex and I believe we could increase readability by doing so, as we wouldn't have the double spread at the beginning.

Benjamin-Dobell:
> Could you move the sorting in a separate map operator function that comes after the one used in this PR?

Yep, good call. It's now a much cleaner PR.

> The code is already very complex and I believe we could increase readability by doing so, as we wouldn't have the double spread at the beginning.

The double spread certainly did gave me pause. However, I've not done anything Reactive since 2016 (I do miss it though). I'm very out of touch with accepted practices (re: appropriate usage of `map`).

It took me far longer than I ought to admit to understand how the partitioning behaviour works 😊

squidfunk:
> It took me far longer than I ought to admit to understand how the partitioning behaviour works 😊

Yeah, it's probably the most complex of all components, but it's efficient and omits excessive repaints by only calculating what's necessary. I'm pretty stunned to see a PR on this component! Great work!

PR #2433, opened 2021-03-13
Add trailing slash to version selector URL's

wbekerom:
Add trailing slash to URL's in version selector in order to be in line with other links in generated docs. This ensures that the relative paths to js / css always work. For me they did not work when deploying to Azure blob storage, hosted as "static website".

You can observe the behavior in the example: https://squidfunk.github.io/mkdocs-material-example-versioning/latest/
- hover over version selector -> no trailing slash
- hover over other link -> trailing slash

Note: I did build the theme, but got a lot of build artifacts, so was not sure whether to include them.


wbekerom:
sorry, messed up branches... this is the one: https://github.com/squidfunk/mkdocs-material/pull/2434

PR #2434, opened 2021-03-13
Add trailing slash to version selector URL's

wbekerom:
Add trailing slash to URL's in version selector in order to be in line with other links in generated docs. This ensures that the relative paths to js / css always work. For me they did not work when deploying to Azure blob storage, hosted as "static website".

You can observe the behavior in the example: https://squidfunk.github.io/mkdocs-material-example-versioning/latest/
- hover over version selector -> no trailing slash
- hover over other link -> trailing slash

Note: I got a huge number of changed graphics when building the theme and did not add these, since this is not something I touched.


src/assets/javascripts/templates/version/index.tsx
```diff
@@ -68,7 +68,7 @@ export function renderVersionSelector(versions: Version[]): HTMLElement {
           <li class="md-version__item">
             <a
               class="md-version__link"
-              href={`${new URL(version.version, config.base)}`}
+              href={`${new URL(version.version + "/", config.base)}`}
```
squidfunk, line 71: Thanks for providing the fix! Could you please use template literal syntax? We use that throughout the entire project.

```
new URL(`${version.version}/`, config.base)}
```

squidfunk (changes requested):

src/assets/javascripts/templates/version/index.tsx
wbekerom: Sure, looks much better!

wbekerom: The build artifacts were conflicting with master, so decided not to include them. Please let me know if they should be.

squidfunk:
Thanks!

PR #2437, opened 2021-03-13
Fix broken screenshots and link in documentation

daaf:
This is a fix for two issues I observed in the documentation.

- On **Setup > Setting up navigation > Hide the sidebars**, the _Hide navigation_/_Hide table of contents_/_Hide both_
screenshots were either broken or showing an incorrect image.
- On **Reference** > **Admonitions** > **Customization** > **Custom admonitions**, the "You can even add animations" link was giving a 404


daaf:
Realized my Prettier added a bunch of trivial formatting changes. Will fix and re-open.

daaf:
Errr...I though I was following the submission guidelines by committing everything in the "material" directory, but now seeing the number of files and lines of code changed, I'm not so sure. Please let me know if I need to do something differently.

polarathene:
You should remove the content you accidentally committed with SVG generated assets in `.icons/` which is unrelated to your PR.

The problem this PR resolves could be better avoided from repeating if using text references instead of numbers (which aren't as pleasant to manage imo).

daaf:
Reverted the unrelated changes in `.icons/`.

squidfunk:
Thanks!

Issue #2417, opened 2021-03-08
Inline Admonition Overlap


jacobswe:
- [x] I've read the [contribution guidelines][1] and agree with them

__I've found a bug and checked that ...__

- [x] ... the problem doesn't occur with the default MkDocs template
- [x] ... the problem is not in any of my customizations (CSS, JS, template)
- [x] ... the documentation does not mention anything about my problem
- [x] ... there are no open or closed issues that are related to my problem

## Description

Inline end admonitions are overlapped by regular admonitions when previously they would be well padded and contained.

### Expected behavior

No overlap between admonitions.

### Actual behavior

![image](https://user-images.githubusercontent.com/14115132/110357175-001e9f80-8009-11eb-9bb5-3e42d6170adb.png)

### Steps to reproduce the bug

Markdown:

```markdown
# Admonition Test

!!! info inline end 
    I too am an admonition.
    
!!! warning
    I am an admonition.
```

### Package versions

* Python: 3.7.3
* MkDocs: 1.1.2
* Material: 7.0.5+insiders.2.2.1

### Project configuration

``` yaml
site_name: My Docs
theme:
  name: material
markdown_extensions:
  - admonition
```

### System information

* OS: linux-gnu
* Browser: Chrome

  [1]: https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md

squidfunk:
Thanks for reporting! This was caused by some changes to the CSS that were necessary to allow code block annotations inside of Admonitions. It should be fixed with the latest commit on Insiders - could you check the latest `master`?

jacobswe:
Thank you for this shockingly fast fix! All looks good on my end at this newest commit. I'll go ahead and consider this issue closed then.

polarathene:
Looks like the issue was flipped? This is against the `latest` Docker image with a fresh `docker pull`:

    !!! info inline end "Result:"

        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla et euismod nulla. Curabitur feugiat, tortor non consequat finibus, justo purus auctor massa, nec semper lorem quam in massa.

    **Markdown:**

    ```md
    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla et euismod nulla. Curabitur feugiat, tortor non consequat finibus, justo purus auctor massa, nec semper lorem quam in massa.
    ```

[...]

PR #2438, opened 2021-03-14
docs: Clarify `inline` admonitions

polarathene:
The declaration order matters and the docs were a tad vague about that.

See [here](https://github.com/squidfunk/mkdocs-material/issues/2417#issuecomment-798825441) for some scenarios where the wrong order isn't intuitive (since `inline end` still renders in the expected style/position, just as a new block).

Initially I thought it was due a code block being too long (the docs version had no scrollbar), but this was not the case. The docs also show off the code block on the left in the first example due to `inline end`. I had to look at the source code for the docs page to confirm the declaration order was important.

I switched the example order, so that the implicit order is implied better that it should be declared prior to sibling block content, and added an additional line of text to further clarify that requirement.

### DIFF
--- a/docs/reference/admonitions.md
+++ b/docs/reference/admonitions.md
@@ -231,29 +231,10 @@ _Result_:
 :octicons-beaker-24: Experimental
 
 Admonitions and [Details][11] can also be rendered as inline blocks (i.e.
-sidebars), moving them to the right using the `inline` + `end` modifiers, or
-to the left using only the `inline` modifier.
+sidebars), placing them to the left using the `inline` modifier, or
+to the right with an additional context modifier: `inline end`.
 
-=== "inline end"
-
-    _Example_ / _Result_:
-
-    !!! info inline end
-
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla et
-        euismod nulla. Curabitur feugiat, tortor non consequat finibus, justo
-        purus auctor massa, nec semper lorem quam in massa.
-
-    ``` markdown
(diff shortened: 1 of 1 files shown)


docs/reference/admonitions.md
```diff
@@ -231,50 +231,52 @@ _Result_:
...
 
-=== "inline end"
+Admonitions that use `inline` should be declared prior to the content block you want to place them beside.
+
+=== "inline"
 
     _Example_ / _Result_:
 
-    !!! info inline end
+    !!! info inline
```
squidfunk, line 243: The blocks are swapped (`inline end` coming before `inline`) because in left-to-right languages, which English is one of, it's more common to place sidebar content to the right, as it doesn't interrupt the left vertical axis.

We can and should add the additional note on placement, though! 😊

squidfunk (changes requested):

docs/reference/admonitions.md
polarathene: > because in left-to-right languages, which English is one of, it's more common to place sidebar content to the right

I understand that, but the placement is visualized in the example, and for English presenting left than right variants regardless of context is the usual expectation of order.

The switch was however as cited earlier for better implying the inline admonition must also be declared first, complimenting the text note added above it. If you want further clarity on left or right positioning, that can be better handled in the Tab names appending `(left-side)` or `(right-side)`?

Your docs though, I can remove the switch if you disagree with it being a UX improvement.

squidfunk:
All sorted in 16274399.

PR #2439, opened 2021-03-14
fix: inline code block inherit link color

polarathene:
Removes the direct ancestor requirement. Any content wrapped in an anchor tag is within the content boundaries of representing a link.

I had a markdown link with inline code block as part of the link text. It rendered as expected, until I altered presentation with italics or bold, due to the `<code>` block being wrapped by another element, preventing the CSS from applying. I see no reason for an direct child requirement in the selector rule.

### DIFF
--- a/src/assets/stylesheets/main/_typeset.scss
+++ b/src/assets/stylesheets/main/_typeset.scss
@@ -210,7 +210,7 @@ kbd {
   }
 
   // Ensure link color in code blocks
-  a > code {
+  a code {
     color: currentColor;
   }
 


squidfunk:
Thanksgiving for the PR. As always, please provide a reproduction. It makes it easier to reason about, trying to find all possibly affected cases, saving your and my time 😊

polarathene:
> As always, please provide a reproduction.

[this is a `reproduction`, *this `code block` lost* it's link colour]()

```markdown
[this is a `reproduction`, *this `code block` lost* it's link colour]()
```

Just copy/paste that to reproduce. It's due to `a > code` selector not being able to target the code block with itallics due to being wrapped in `<em>`, you'd need ` a > em > code`.

But as mentioned, there is no reason with an anchor link for the content within it to require direct children only to receive such styles. Any `<code>` regardless of nesting in an `<a>` is an appropriate target for the selector, so drop the `>` and the rendering difference will be resolved.

squidfunk:
Thanks. I think it is safe to widen the descendant relationship. One thing we need to make sure is that code blocks with inline highlighting remain unaffected by that change, which my testing shows they do.

Issue #2466, opened 2021-03-16
Base image of Dockerfile python:3.8.7-alpine3.12 contains vulnerable libraries


BenHimsel:
- [x] I've read the [contribution guidelines][1] and agree with them

__I've found a bug and checked that ...__

- [x] ... the problem doesn't occur with the default MkDocs template
- [x] ... the problem is not in any of my customizations (CSS, JS, template)
- [x] ... the documentation does not mention anything about my problem
- [x] ... there are no open or closed issues that are related to my problem

## Description

Docker image is currently based on python:3.8.7-alpine3.12 which contains outdated packages such as openssh 8.3_p1-r1 which suffers from [CVE-2021-28041](https://nvd.nist.gov/vuln/detail/CVE-2021-28041).

BenHimsel:
[Pull Request 2467](https://github.com/squidfunk/mkdocs-material/pull/2467) submitted with updated python base image.

mkdocs build and mkdocs gh-deploy validated:
```
/ # mkdocs build
INFO    -  Cleaning site directory
INFO    -  Building documentation to directory: /site
INFO    -  Documentation built in 7.23 seconds
/ # mkdocs gh-deploy
INFO    -  Cleaning site directory
INFO    -  Building documentation to directory: /site
INFO    -  Documentation built in 7.12 seconds
WARNING -  Version check skipped: No version specified in previous deployment.
INFO    -  Copying '/site' to 'gh-pages' branch and pushing to GitHub.
INFO    -  Your documentation should shortly be available at: https://...
```

polarathene:
That's not really an exploit that can be taken advantage of with this image, especially in how it's used afaik? (which runs as root user anyway..)

squidfunk:
Fixed in 56e8162a0:

``` diff
-FROM python:3.8.7-alpine3.12
+FROM python:3.9.2-alpine3.13
```

squidfunk:
Released as part of 7.0.7.

polarathene:
@squidfunk The "fix available" label doesn't really apply to this issue?

The image was updated but that didn't update to a newer package of openssh 8.5 did it? I provided details in the related PR but received no response.

Technically if this vulnerability is a concern (it doesn't appear to be imo) and the package isn't able to be updated with the fix, this issue shouldn't be closed?

squidfunk:
In https://github.com/squidfunk/mkdocs-material/pull/2467#issuecomment-807772023 you wrote:

[...]

PR #2467, opened 2021-03-16
Fixes squidfunk/mkdocs-material#2466

BenHimsel:
Update base python image to 3.8.8


jaceklaskowski:
Why not 3.9.2-alpine3.13 since it's the [latest](https://hub.docker.com/_/python)?

BenHimsel:
The short answer would be my lack of knowledge. I did not know the effects of updating a minor version, but assumed that updating the latest patch version would be a safe contribution.

If updating to the latest minor version is the preferred update methodology, I would be happy to update the PR.

squidfunk:
We should update to the latest version. Could you update the PR and re-verify that building, serving, and deploying still works?

squidfunk (changes requested):
Could you please check if we can safely update to the latest version as proposed by @jaceklaskowski?

polarathene:
I don't see this as an exploit that is relevant to the context of the docker image personally.

You already run as the root user within the docker image (and the docker daemon usually is run as root), how exactly is this attack being envisioned of being carried out, with what goal and what present access to perform it?

[Upstream notes it as](https://www.openssh.com/txt/release-8.5):

> ssh-agent(1): fixed a double-free memory corruption that was introduced in OpenSSH 8.2 . We treat all such memory faults as potentially exploitable.
>
> The most likely scenario for exploitation is a user forwarding an agent either to an account shared with a malicious user or to a host with an attacker holding root access.

Ignoring that, you can easily check what version is shipped by Alpine via [their web interface for packages](https://pkgs.alpinelinux.org/package/v3.13/main/x86_64/openssh) (3.13 provides `openssh-8.4_p1-r3`). You would need to use [`edge`](https://pkgs.alpinelinux.org/packages?name=openssh&branch=edge) (which is a bad idea) or wait until `3.14` AFAIK for openssh version 8.5.

So no, updating the image does nothing to resolve this concern even if it were valid.

squidfunk:
Superseded by 56e8162a0.

PR #2470, opened 2021-03-17
Add tip about setting use_directory_urls: false for offline docs

johnthagen:
When packaging up a Material for MkDocs project for offline HTML file system distribution, I initially had trouble finding a reference to solve this issue: https://github.com/mkdocs/mkdocs/issues/526#issuecomment-102064019

The Offline Search is currently the only place I could find that Material for MkDocs discusses this concept, so I added a `tip` to help other users that come across this in the future.


squidfunk:
Thanks! I agree that it's a good idea to add this info!

PR #2471, opened 2021-03-18
added some missing hungarian localizations

koltai75:
Some Hungarian localizations were missing from hu.html, for example search result details.
Added these to hu.html.


squidfunk:
Thanks for providing!

koltai75:
Sure thing! Thank you too, mkdocs-material rocks!

PR #2487, opened 2021-03-25
Fix disqus breaks when instant loading enabled

PegasisForever:
Use [DISQUS.reset](https://help.disqus.com/en/articles/1717163-using-disqus-on-ajax-sites) to update disqus section when navigating to other pages.


squidfunk:
Thanks for the PR! What happens when coming from a page with Disqus to a page without? Is `DISQUS.reset` a no-op?

PegasisForever:
If a page doesn't have disqus, the disqus partial won't be in the html, so the code won't be there and will not run.

Also, I wonder is there a "proper" way to check if the page is loaded by instant loading? (like an event or something), checking `document
readyState` feels like a hack.

squidfunk:
Yes there is:

```
document$.subscribe(function() {
  ...
})
```

The `document$` observable will emit on initial load, and on every subsequent load when using instant loading.

squidfunk:
Superseded by e71b12943. This PR has become stale, so I decided to take the lead here. I could simplify the code a little, and we now have a working Disqus / instant loading integration. Thanks again for your work!

squidfunk:
Released as part of 7.3.1.

Issue #2488, opened 2021-03-25
Buggy header shadow


PegasisForever:
- [x] I've read the [contribution guidelines][1] and agree with them

__I've found a bug and checked that ...__

- [x] ... the problem doesn't occur with the default MkDocs template
- [x] ... the problem is not in any of my customizations (CSS, JS, template)
- [x] ... the documentation does not mention anything about my problem
- [x] ... there are no open or closed issues that are related to my problem

## Description

Header shadow randomly appears when it shouldn't.
This affects almost every site using mkdocs-material, e.g. https://notes.pegas.is/, https://walkccc.me/LeetCode/
This affects both Chrome and Firefox.
However, the official documentation site doesn't seem to be affected. (probably due to it's announcement bar?)

### Expected behavior

Header shadow only appears when the page is scrolled down.

### Actual behavior

Video: https://cloud.pegasis.site/s/Q5H7LFWoKtWJHcd

### Steps to reproduce the bug

It's so random that I can't give exact steps, try repeating those steps multiple times in different orders:
- resize window
- scroll page
- refresh

### Package versions

* Python: `3.9.2`
* MkDocs: `1.1.2`
* Material: `7.0.6`

### Project configuration

``` yaml
site_name: Pega's Notes
docs_dir: public
site_dir: build
theme:
  name: material
  custom_dir: .overrides
  font:
    text: Fira Sans
    code: JetBrains Mono
  palette:
    primary: teal
    accent: teal
  logo: pega_logo_filled.svg
  favicon: favicon.png
  features:
    - navigation.instant
    - toc.integrate
    - search.suggest
plugins:
  - awesome-pages
  - section-index
  - search:
      lang:
        - en
  - git-revision-date-localized:
      enable_creation_date: true
      timezone: Amarica/Toronto
      locale: en
markdown_extensions:
  - pymdownx.snippets:
      check_paths: true
  - pymdownx.superfences
  - pymdownx.magiclink
  - pymdownx.tilde
  - pymdownx.caret
  - pymdownx.emoji
  - pymdownx.mark
  - pymdownx.smartsymbols
  - toc:
      permalink: true
extra:
  homepage: https://pegas.is/
  disqus: pega-notes
  social:
    - icon: fontawesome/brands/github
      link: https://github.com/PegasisForever
    - icon: fontawesome/brands/instagram
      link: https://www.instagram.com/__pegasis/
extra_css:
  - extra.css
extra_javascript:
  - extra.js
[...]

PR #2494, opened 2021-03-26
Fix buggy header shadow

PegasisForever:
Fix issue: #2488

Explained in this comment: #2488 (comment)


squidfunk:
Thanks for investigating. So, in the end, it's only a rounding error, right?

PegasisForever:
Ya it's just a rounding error, my code works with tabs and announcement bar, tested on the official docs site.

squidfunk:
Thanks for the PR and your investigation. It's definitely related to rounding errors. However, the fix doesn't solve the issue in its entirety. When I zoom out, the header shadow is still not visible with your fix applied. The problem stems from the fact that `getElementSize` uses the `offsetWidth` and `offsetHeight` properties, while `watchElementSize` returns `DOMRect` objects. The former is the effective size as rendered (thus, no floating point values), the latter what has been computed during layout calculation (floating point values).

We definitely need to make those computations consistent, but we can't just use `Math.round`, because [different browsers round differently](https://cruft.io/posts/percentage-calculations-in-ie/), so the behavior would not be consistent.

squidfunk:
Superseded by f43ef69ee – I fixed the root cause.

Issue #2508, opened 2021-03-29
block extracopyright not overriden


EdouardThiel:
Hello,

`extracopyright` doesnt work for me:

- I have added `custom_dir: overrides` after the material theme in `mkdocs.yml`
- also added an `overrides/main.html` file:

``` jinja
{% extends "base.html" %}

{% block htmltitle %}
<title>This is a title</title>
{% endblock %}

{% block extracopyright %}
This is my extra copyright
{% endblock %}
```

The title works, but the `extracopyright` outputs nothing; that's because in `base.html` the block `footer` includes `partials/footer.html` and currently, includes doesn't override blocks.

Is it possible to correct the templates with a variable or a macro instead of a block `extracopyright`? or a special way to import footer.html ? 

For instance replace the block in `footer.htm`l by

``` jinja
        Made with
        <a href="https://squidfunk.github.io/mkdocs-material/" target="_blank" rel="noopener">
          Material for MkDocs
        </a>
        {{ extracopyright }}
```

or even better, define `extracopyright` in the config (like `copyright`) and so in `footer.hml` it would be

``` jinja
        Made with
        <a href="https://squidfunk.github.io/mkdocs-material/" target="_blank" rel="noopener">
          Material for MkDocs
        </a>
        {{ config.extracopyright }}
```
without having to write an `overrides/main.html` file.

squidfunk:
> But it outputs nothing, because the block footer includes "partials/footer.html" and currently, includes doesn't override blocks.

That's correct, that's a limitation of Jinja we completely forgot about 😅 I don't think we should add another variable in `mkdocs.yml`, because we already have `copyright` and adding `extracopyright` is kind of confusing. I think the best thing we can do is remove the template block, because currently it has no use. You can override `footer.html`, if you wish to.

EdouardThiel:
But it would be very simple to replace it by

``` jinja
        Made with
        <a href="https://squidfunk.github.io/mkdocs-material/" target="_blank" rel="noopener">
          Material for MkDocs
        </a>
        {{ extracopyright }}
```

and then, in `overrides/main.html`, write

``` jinja
{% extends "base.html" %}

{% block htmltitle %}
<title>This is a title</title>
{% endblock %}

{% set extracopyright %}
This is my extra copyright
{% endset %}
```

It works nice, and the main advantage is that I don't have to broke the footer logic, nor froze the `footer.html` with a modifed copy!

XhmikosR:
[...]

PR #2509, opened 2021-03-29
Update extracopyright in footer (fixes issue 2508)

EdouardThiel:
Replace the extracopyright block,which doesnt' work, by a variable `{{ extracopyright }}` that can be set e.g from `overrides/main.html` with `{% set extracopyright %} ... {% set %}`; see issue #2508.

### DIFF
--- a/src/partials/footer.html
+++ b/src/partials/footer.html
@@ -94,7 +94,7 @@
         >
           Material for MkDocs
         </a>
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! Could you also update the docs and remove the `extracopyright` block from `customization.md`?

squidfunk:
Thanks for the PR!

PR #2529, opened 2021-04-06
Adding Venafi to list of companies

fergven:
Added link to Venafi docs to the list of companies using material for mkdocs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ For other installation methods, configuration options, and a demo, visit
 [Salesforce](https://policy-sentry.readthedocs.io/en/latest/),
 [SoundCloud](https://intervene.dev/),
 [Square](https://square.github.io/okhttp/),
+[Venafi](https://docs.venafi.cloud),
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for your PR. However, I'm still unsure what to do with the list of companies. Obviously several users wanted to add their companies to the list of users, which I find to be a great idea in general, but then, where to draw the line so the list doesn't grow beyond a count of 10-15? For now, I tried sticking to a list of companies that are well known in the tech space. Personally, I've never heard of Venafi before (before our email correspondence), but that might be completely different for other users. Thus, I'm all for keeping the list of users to companies practically everybody knows, like AWS, Microsoft, etc., because I believe that this provides the most value at that point to convince users that Material for MkDocs is a stable and mature solution for technical documentation.

On a side note, I would be very happy to add more companies with a logo and backlink to the list of [premium sponsors](https://github.com/squidfunk/mkdocs-material#premium-sponsors), if they decide to fund the project!

Issue #2550, opened 2021-04-12
Change left navigation menu title with a simple variable


EdouardThiel:
By default, the site name is used as the title for the left navigation menu.

This is not always desirable; for instance if the site name is long, it could be replaced by a short title or simply by "Navigation".

For now, material does not offer a simple solution, and I would like to suggest one:

1) In `material/partial/nav.html`, replace line 16

        {{ config.site_name }}

    with

        {% if nav_title|trim %}{{ nav_title }}{% else %}{{ config.site_name }}{% endif %}

2) The user will then, if he likes to, add in `overrides/main.html`:

        
        {% set nav_title %}My navigation title{% endset %}

I have used `nav_title` because this variable is unused in Material, but another name could fit too, e.g. `extranavtitle`.

Regarding the documentation, the natural place would be in Getting started / customization, in the last table with `extracopyright`.

squidfunk:
Thanks for your suggestion. I'm not sure if this is the right way to implement this behavior. Note that when you set the navigation title like you propose in the linked PRs (thanks for those, by the way), you will also override the title in the drawer which, in my opinion, should definitely show the site title. This is because the sidebar is re-used on mobile (in order to minimize duplicated content) and rendered as a drawer which collapses into the side:

<img width="242" alt="Bildschirmfoto 2021-04-13 um 18 27 23" src="https://user-images.githubusercontent.com/932156/114587406-ed415f80-9c85-11eb-89c0-9eac502699da.png">

Thus, I'm not considering including this feature request / behavior. After all, overriding `nav.html` is fairly easy and the navigation partial didn't change much during the last major updates (last time a year ago with 5.x).

EdouardThiel:
Well this allows short titles, even on mobiles, not all sites have a short title as "material for mkdocs" and this features is actually missing, there are questions about it on stack overflow.This is a simple solution. Another would be to define config.short_nav_title in config. May I submit it?-- Professeur Edouard THIEL                  http://j.mp/ethiel
Département Informatique et Interactions  http://j.mp/dept-ii
UFR Sciences, Aix-Marseille Université    http://j.mp/fs-amu
Laboratoire d'Informatique et Systèmes    http://j.mp/lis-amu

[...]

PR #2559, opened 2021-04-13
Update nav.html to change left navigation menu title, see #2550

EdouardThiel:


squidfunk:
Not considering for merging, see #2550.

PR #2560, opened 2021-04-13
Update customization.md with nav_title, see #2550

EdouardThiel:
I don't know if an example is necessary before the table, since it works exactly like extracopyright:
{% set %}My short title{% endset %} or {% set %}Navigation{% endset %}


squidfunk:
Not considering for merging, see #2550.

PR #2567, opened 2021-04-15
Fix example for hero deprecation

jairbubbles:
- If  the `custom_dir` is `overrides` (as specified  in the documentation) it shouldn't be specified in the template path
- `{% extends "base.html" %}` is needed in the template


squidfunk:
Thanks, good catch!

PR #2568, opened 2021-04-15
Update highlight.js example

jairbubbles:
- Use latest `highlight.js` version that now uses `highlightAll()`
- Fix when instant navigation is enabled (see #2569)


squidfunk:
Thanks!

PR #2572, opened 2021-04-15
Add powershell docker examples

jairbubbles:
Powershell is becoming more and more popular on Windows so I feel it's worth adding a bit of precision. Please note that the commands are the exat same as on Linux so it might feel like overkill to add a new tab. But from a user point of view it might the best, let me know ;-)

### DIFF
--- a/docs/creating-your-site.md
+++ b/docs/creating-your-site.md
@@ -20,10 +20,16 @@ Alternatively, if you're running Material for MkDocs from within Docker, use:
     docker run --rm -it -v ${PWD}:/docs squidfunk/mkdocs-material new .
     ```
 
-=== "Windows"
+=== "Windows (CMD)"
 
     ```
     docker run --rm -it -v "%cd%":/docs squidfunk/mkdocs-material new .
+    ```
+
+=== "Windows (Powershell)"
+
+    ```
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. Don't really like that it's a redundant tab with identical content. Canw e remove `CMD`? Is that even used anymore? I'm not a Windows user.

jairbubbles:
Yes it's still used. It remains the default when you run a command and will be probably for years. But powershell is the future.

Maybe we could say: just run this command if you're on linux or using powershell and have a collapsible block for the CMD command for windows?

Anyways, it's not a big deal I'm ok if you close the PR ;-)

Honestly, I added the powershell tab an then I noticed it was the same command than Linux :-D

squidfunk:
I know it's not perfectly disjoint, but in f8d5f016c I decided to just add Powershell to the first tab, as it's the same syntax 😊 "Unix" is also not very specific, as it demands that you know that macOS is also a Unix derivative, but I guess it's like that:

- __macOS__: well, obviously not Windows
- __Linux__: Windows? Never heard of
- __Windows, Powershell__: ha! Powershell!
- __Windows, CMD__: what is Powershell? I'm going for Windows...

PR #2579, opened 2021-04-19
Add support for git creation date

timvink:
Closes #2551

PR #2627, opened 2021-05-02
Fix source-file include

Stanzilla:
Fixes https://github.com/squidfunk/mkdocs-material/pull/2579#issuecomment-830806927

### DIFF
--- a/material/base.html
+++ b/material/base.html
@@ -169,7 +169,7 @@
                   {% if page.meta.git_revision_date_localized or
                         page.meta.revision_date
                   %}
-                    {% include "partials/source-date.html" %}
+                    {% include "partials/source-file.html" %}
                   {% endif %}
                 {% endif %}
               {% endblock %}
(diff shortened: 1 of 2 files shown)


timvink:
This will break compatibility for people using the `source-date` partial. See discussion here: https://github.com/squidfunk/mkdocs-material/pull/2579#issuecomment-831081158

Another fix available in https://github.com/squidfunk/mkdocs-material/pull/2629

squidfunk:
Actually, it should be safe to merge. If the user hasn't patched `base.html`, it will pick up the new file. In case the user did, and there's a reference to `source-date.html`, we have it covered by including `source-file.html` in `source-date.html`.

material/base.html
```diff
@@ -197,7 +197,7 @@ <h1>{{ page.title | d(config.site_name, true)}}</h1>
         "base": base_url,
         "features": features,
         "translations": {},
-        "search": "assets/javascripts/workers/search.4fa0e4ee.min.js" | url,
```
squidfunk, line 200: This got merged accidentally and broke search. Fixed in 5bc689cdf

PR #2631, opened 2021-05-03
Add title to to-top button

Stanzilla:


material/partials/languages/en.html
```diff
@@ -33,5 +33,6 @@
   "source.file.date.updated": "Last update",
   "source.file.date.created": "Created",
   "tabs.title": "Tabs",
-  "toc.title": "Table of contents"
+  "toc.title": "Table of contents",
+  "topbutton.title": "Back to Top"
```
squidfunk, line 37: Thanks! Could you rename it into `top.title`? I know that's probably a little less self-explanatory, but it matches the style of `toc` and `tabs` with their feature flags:

- `tabs` -> `navigation.tabs`
- `toc` -> `navigation.toc`
- `top` -> `navigation.top`

I'm all for keeping consistency.

squidfunk (changes requested):

material/partials/languages/en.html
Stanzilla: Yup, done!

squidfunk, line 37: One more missing 😅

squidfunk (changes requested):

src/partials/languages/en.html
```diff
@@ -53,5 +53,6 @@
   "source.file.date.updated": "Last update",
   "source.file.date.created": "Created",
   "tabs.title": "Tabs",
-  "toc.title": "Table of contents"
+  "toc.title": "Table of contents",
+  "top.title": "Back to Top"
```
squidfunk, line 57: Sorry, one more thing - I've overseen. It should be "Back to top", not "Back to Top".

squidfunk (changes requested):

squidfunk:
Code review:

10 lines of code – "hmm that variable isn't named appropriately", "you're missing indentation", "no we should indent like this to congey meaning"

500 lines of code – APPROVED. next.

PR #2725, opened 2021-05-31
Add aria-label to language selector button

Stanzilla:
Adds aria-label to language selector buttons since otherwise screen readers are not able to read it.


squidfunk:
Thanks for the PR. I’d say we call it „Select language“ and add i18n support via `language.select`.

squidfunk:
I've done a little cleanup in 11fcf4f0a:

- Changed my mind on the translation key again
- Added an `aria-label` to the version selector.

PR #2726, opened 2021-05-31
Add title to prev/next footer links

Stanzilla:
Adds a title to the pre/next footer links to satisfy accessibility requirements. On mobile only the icon is shown so the text is not providing the needed context for screen readers.

### DIFF
--- a/material/partials/footer.html
+++ b/material/partials/footer.html
@@ -6,7 +6,7 @@
   {% if page.previous_page or page.next_page %}
     <nav class="md-footer__inner md-grid" aria-label="{{ lang.t('footer.title') }}">
       {% if page.previous_page %}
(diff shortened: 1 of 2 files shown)


squidfunk:
Superseded by 0dd483121. Here, `aria-label` is the better choice, as `title` will lead to a duplication, at least with VoiceOver on macOS. `aria-label` fully overrides the content of the link and now reads as:

- Next: "title of next page"
- Previous: "title of next page"

Labels are sourced from the set of translations. I also checked with Lighthouse and the following warning is now gone (which is probably what you were aiming for):

Bildschirmfoto 2021-06-06 um 14 29 40

Furthermore, accessibility score improved from 89 to 91.

PR #2730, opened 2021-06-01
Add aria-label to palette toggle

Stanzilla:
Otherwise screen readers won't be able to tell the different values. 

See:

- https://www.webaccessibility.com/results/?url=https%3A%2F%2Fsquidfunk.github.io%2Fmkdocs-material%2Fsetup%2Fchanging-the-colors%2F
- https://wave.webaim.org/report#/https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/

### DIFF
--- a/material/partials/header.html
+++ b/material/partials/header.html
@@ -32,7 +32,7 @@
         {% for option in config.theme.palette %}
           {% set primary = option.primary | replace(" ", "-") | lower %}
           {% set accent  = option.accent  | replace(" ", "-") | lower %}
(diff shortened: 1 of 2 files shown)


squidfunk:
Superseded by 42500aafa. Just adding the `aria-label` is dangerous, because toggles might not be defined for color palettes, which is why there is a separate check for `option.toggle` below the `input` field. The idea is that people could provide their own toggle implementation or even change it programmatically, so we should definitely keep this flexibility.

Issue #2727, opened 2021-05-31
Light/dark theme disconnected from light/dark scheme


posita:
## Expected behavior

After reading [this](https://github.com/squidfunk/mkdocs-material/blob/4247e3b440644cfcf7219f1b32e147f13a490f13/docs/setup/changing-the-colors.md#system-preference), as well as squidfunk/mkdocs-material#1266, my understanding is:

1. Toggling the theme using the toggle switch will switch the preferred color scheme; and
1. Switching the preferred color scheme will toggle the theme.

However, they are disconnected as far as I can tell.

## Observed behavior

1. Toggling the theme using the switch in the header of this page will update the theme, but not the scheme.
1. Toggling the preferred color scheme (e.g., on MacOS, switching from `Light` to `Dark` in `System Preferences` » `General` » `Appearance`) will update the scheme, but not the theme.

## Detail

I have created [an example](https://posita.github.io/mkdocs-dark-bug/) that illustrates this. (Source can be found at [**posita/mkdocs-dark-bug**](https://github.com/posita/mkdocs-dark-bug).)

Test matrix:

| Theme&nbsp;👉<br />Scheme&nbsp;👇 | Light | Dark  |
| --------: | :---: | :---: |
| **Light** | ![Light/Light](https://posita.github.io/mkdocs-dark-bug/latest/light-theme-light-scheme.png) | ![Dark/Light](https://posita.github.io/mkdocs-dark-bug/latest/dark-theme-light-scheme.png) |
| **Dark**  |  ![Light/Dark](https://posita.github.io/mkdocs-dark-bug/latest/light-theme-dark-scheme.png)  |  ![Dark/Dark](https://posita.github.io/mkdocs-dark-bug/latest/dark-theme-dark-scheme.png)  |

Tested with:

| package                    | version |
| :------------------------- | ------: |
| mkdocs                     | 1.1.2   |
| mkdocs-material            | 7.1.6   |
| mkdocs-material-extensions | 1.0.1   |

squidfunk:
Thanks for reporting. `prefers-color-scheme` uses the system preference, correct. Material for MkDocs can only react to system preference, but not change it. That’s just not possible. Thus, everything is working as expected.

posita:
[...]

PR #2731, opened 2021-06-02
Improve wording of Changing the Colors » System preference

posita:
Motivated by #2727.

### DIFF
--- a/docs/setup/changing-the-colors.md
+++ b/docs/setup/changing-the-colors.md
@@ -216,7 +216,7 @@ To match a color palette to the user's system preference upon the user's first
 visit, provide a media query via the `media` field next to the `scheme` definition
 in `mkdocs.yml`:
 
-``` yaml hl_lines="3 8"
+``` yaml hl_lines="4, 9"
 theme:
   palette:
     - scheme: default
--- a/docs/setup/changing-the-colors.md
+++ b/docs/setup/changing-the-colors.md
@@ -216,16 +216,16 @@ To match a color palette to the user's system preference upon the user's first
 visit, provide a media query via the `media` field next to the `scheme` definition
 in `mkdocs.yml`:
 
-``` yaml hl_lines="4, 9"
+``` yaml hl_lines="3 8"
 theme:
   palette:
-    - scheme: default
-      media: "(prefers-color-scheme: light)"
+    - media: "(prefers-color-scheme: light)"
+      scheme: default
       toggle:
         icon: material/toggle-switch-off-outline
         name: Switch to dark mode
-    - scheme: slate
-      media: "(prefers-color-scheme: dark)"
+    - media: "(prefers-color-scheme: dark)"
+      scheme: slate
       toggle:
         icon: material/toggle-switch
         name: Switch to light mode
--- a/docs/setup/changing-the-colors.md
+++ b/docs/setup/changing-the-colors.md
@@ -5,7 +5,7 @@ template: overrides/main.html
 # Changing the colors
 
 As any proper Material Design implementation, Material for MkDocs supports
-Google's original [color palette][1], which can be easily configured through 
+Google's original [color palette][1], which can be easily configured through
 `mkdocs.yml`. Furthermore, colors can be customized with a few lines of CSS to
 fit your brand's identity by using [CSS variables][2].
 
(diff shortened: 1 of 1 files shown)


docs/setup/changing-the-colors.md
```diff
@@ -208,12 +208,12 @@ The following fields must be set for each toggle:
   [9]: #accent-color
   [10]: https://github.com/squidfunk/mkdocs-material/tree/master/material/.icons
 
-### System preference
+### Inferring the initial setting from system preference
```
squidfunk, line 211: We shouldn't change the headline title, because we'd lose all anchor links. Please change it back.

docs/setup/changing-the-colors.md
```diff
@@ -208,12 +208,12 @@ The following fields must be set for each toggle:
...
 
 [:octicons-file-code-24: Source][6] · :octicons-milestone-24: Default: _none_
 
-In order to automatically set the color palette to the user's system preference,
-a media query can be set as part of the `media` field next to the toggle
+To match a color palette to the user's system preference upon the user's first
```
squidfunk, line 215: I find "to match ..." less intuitive.

squidfunk (changes requested):

docs/setup/changing-the-colors.md
posita: What would you suggest as an alternative?

posita: I kept the "To match …" for now, but word-smithed the remainder a bit. Hopefully that helps?

squidfunk: No, not really.

docs/setup/changing-the-colors.md
```diff
@@ -212,28 +212,28 @@ The following fields must be set for each toggle:
...
-a media query can be set as part of the `media` field next to the toggle
-definition in `mkdocs.yml`:
+To match a color palette to the user's system preference upon the user's first
+visit, provide a media query via the `media` field next to the `scheme` definition
+in `mkdocs.yml`:
 
 ``` yaml hl_lines="3 8"
 theme:
   palette:
-    - media: "(prefers-color-scheme: light)"
```
squidfunk, line 222: I don't know whether I've not seen this in the first review, because you force-pushed (please don't do that, as we loose track of changes). 

Why did you change the order? First of all, now the wrong lines are highlighted, and second of all I prefer `media` to be at the top, because it's more clear that the following settings are tied to it.

squidfunk (changes requested):

docs/setup/changing-the-colors.md
posita: Yeah, this was new, because it dawned on me after discussing "match[ing]" above that the ordering seemed…_off_.

Thanks for reminding me about the highlighting! Should be fixed now.

My suggestion is to put `scheme` first, because it is the primary thing everything else is anchored. Both `media` and `toggle` are subordinate to `scheme` no? (Meaning neither makes sense without `scheme` being present, correct?)

docs/setup/changing-the-colors.md
posita: If my original defense was unpersuasive, what about this?

> To ~match~ _set_ a color palette _according_ to the user's system preference upon the user's first …

FWIW, I really think my original proposal is not only more concise, but more plainly and directly describes what's going on. "To set … according to …" means "match" in this context. Matching is fundamentally what's being done, no?

docs/setup/changing-the-colors.md
squidfunk: You can set any of the primary, accent or scheme values regardless of each other, so there’s no parent relationship with the scheme.

The current order is more intuitive because it read like “for light mode I set primary to ...”

posita: Sorry for the confusion. I have reverted this.

For what it's worth, I was suggesting they were semantically dependent (at least given what was present in the example), not syntactically. I realize that they are peers in the YAML schema. I only meant to highlight that `media` and `toggle` are effectively meaningless without at least one of `primary`, `accent`, or `scheme` (and only `scheme` was present in the example). Consider, for example, the following configuration:

```yaml
theme:
  palette:
    - media: "(prefers-color-scheme: light)"
      toggle:
        icon: material/toggle-switch-off-outline
        name: Switch to some mode
    - media: "(prefers-color-scheme: dark)"
      toggle:
        icon: material/toggle-switch
        name: Switch to some other mode
```

In any event, I backed this out per your request.

docs/setup/changing-the-colors.md
posita: Any suggestions then?

If my original defense was unpersuasive, what about this?

> To ~match~ _set_ a color palette _according_ to the user's system preference upon the user's first …

FWIW, I really think my original proposal is not only more concise, but more plainly and directly describes what's going on. "To set … according to …" means "match" in this context. Matching is fundamentally what's being done, no?

squidfunk:
Honestly, I find the original paragraphs (both) better as I find them easier to understand than the current draft.

posita:
It's your baby. 😊 I'll withdraw the PR. Thanks for the consideration, though!

Not to beat a dead horse, but I think this passage as written is probably accessible to someone who already understands the subject matter (such as yourself). Consider, though, that human beings often have a difficult time empathizing with others who lack similar context or expertise.

The passage—particularly the phrase, "… automatically set the color palette to the user's system preference …"—left me with a fundamental misunderstanding of even the desired behavior. So much so that I went through the effort to file #2727 thinking the behavior was somehow incorrect, and then followed up trying to addressing the source of my confusion in this PR. That being said, I'm a sample size of one, and not so arrogant to assume my experience is representative, either. I'm open to the possibility that I am an outlier.

No worries if you don't think my edits are helpful. I appreciate the time, attention, and consideration. Please consider revisiting this topic if similar confusion arises from another source.

squidfunk:
Thanks for your understanding! IIRC you're the first one being tripped up by the wording, so I think we're not really solving a problem here. However, we'll definitely revisit if the need arises.

PR #2739, opened 2021-06-07
Fix a possible conflict with gh-pages

Nthily:
today, i ran into this [issue](https://github.com/mkdocs/mkdocs/issues/2370) tonight

But the strange thing is that when I didn't write this command before, it was working fine.

By the way, without adding this command, `- git-revision-date-localized` will not work properly

image

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -37,6 +37,8 @@ contents:
         runs-on: ubuntu-latest
         steps:
           - uses: actions/checkout@v2
+            with:
+              fetch-depth: 0
           - uses: actions/setup-python@v2
             with:
               python-version: 3.x
@@ -59,6 +61,8 @@ contents:
         if: github.event.repository.fork == false
         steps:
           - uses: actions/checkout@v2
+            with:
+              fetch-depth: 0
           - uses: actions/setup-python@v2
             with:
               python-version: 3.x


aahnik (approved):
good!

squidfunk:
Thanks for the PR. This problem is exclusive to the `git-revision-date-localized` plugin and should be mentioned in their documentation. For users that don't use this plugin, the changes will lead to longer build times because now [all history for all branches and tags is fetched](https://github.com/actions/checkout#checkout-v2):

> Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set fetch-depth: 0 to fetch all history for all branches and tags. Refer here to learn which commit $GITHUB_SHA points to for different events.

Thus, I don't see why this should appear in the publishing guide.

Nthily:
Makes sense, I should change these in 'git-revision-date-localized' document？

aahnik:
Thanks @squidfunk for the explanation. I now understand. I will also use a force push, from now on.

vabene1111:
interestingly my build is failing since yesterday as well without any changes made anywhere (neither in docs nor actions nor the mkdocs.yml)
https://github.com/vabene1111/recipes/runs/2776791066?check_suite_focus=true

i do not use the plugin described above, any idea what might be broken because using --force should eliminate the issue ?

If you want me to open a real issue just say so and i will create a proper report

Nthily:
> interestingly my build is failing since yesterday as well without any changes made anywhere (neither in docs nor actions nor the mkdocs.yml)
> https://github.com/vabene1111/recipes/runs/2776791066?check_suite_focus=true
> 
> i do not use the plugin described above, any idea what might be broken because using --force should eliminate the issue ?
> 
> If you want me to open a real issue just say so and i will create a proper report

i dont know, i got exactly the same issue as you, also last night, i just updating the .md file, and it got this error

aahnik:
> interestingly my build is failing since yesterday as well without any changes made anywhere (neither in docs nor actions nor the mkdocs.yml)

Last day there was an internet outage. Github, gitlab, stackoverflow, and many other popular sites may be affected. The build may be failing due to some component server being down.

jaceklaskowski:
It does not seem to be a coincidence that it surfaced just two days ago - see https://github.com/mkdocs/mkdocs/issues/2448 and more importantly https://github.com/mkdocs/mkdocs/issues/2448#issuecomment-857624041. HTH

PR #2747, opened 2021-06-09
Add ability to override homepage title (config.extra.homepage_title)

kellymenzel:
While you can change homepage URL, I was wondering if it made sense to also allow for the hompage title to be changed. Allowing this will prevent you from having to override the header.html partial and goes along with the homepage URL.

### DIFF
--- a/docs/setup/changing-the-logo-and-icons.md
+++ b/docs/setup/changing-the-logo-and-icons.md
@@ -48,6 +48,7 @@ with the following configuration:
 ``` yaml
 extra:
   homepage: https://example.com
+  homepage_title: Example
 ```
 
 ### Favicon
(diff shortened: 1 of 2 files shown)


squidfunk:
Thanks for the PR. Why not just use the site name?

kellymenzel:
The home page for us is our internal Docs website that essentially is a catalog or index for all our mkdocs-based websites (each for different projects or teams that have their own repositories and builds). The site that the homepage override points to is a different site than the current docs site. 

I understand if we don't really want this for everyone, so feel free to close the PR. I was able to achieve this using the header.html partial override, but since we have probably 20 or 30 mkdocs repositories, it is quite a bit of code and files I have to remember to copy to each repository and keep in sync if it changes. Let me know your thoughts

squidfunk:
While I understand your use case, I don't want to add more and more indirections as it increases the maintenance burden and makes behavior less predictable. As a rule of thumb, link titles (which this is about) should be deduced from other variables if possible, with social links being the exception as there is no title to use.

PR #2751, opened 2021-06-12
Fix Chinese translation for "source.file.date.created"

jfcherng:
Used in https://github.com/squidfunk/mkdocs-material/blob/9da2afef912f90da736b75c4f42bf5f8a32e6ca1/material/partials/source-file.html#L12


squidfunk:
Thanks!

PR #2772, opened 2021-06-19
Update `name` info in config docs

Stanzilla:
This can actually stay as `material` and does not have to be `null`. Benefit is that even when using a fork/material as submodule, plugins can rely on the name from the config for feature detection.

### DIFF
--- a/docs/creating-your-site.md
+++ b/docs/creating-your-site.md
@@ -56,7 +56,7 @@ slightly different:
 
     ``` yaml
     theme:
-      name: null
+      name: material
       custom_dir: mkdocs-material/material
 
       # 404 page


squidfunk:
That diverges from the official MkDocs recommendation and my guess is that it only works if you have some version of Material installed alongside your clone. More specifically, it might currently work because `mkdocs-material-extensions`, which is installed when you install the `requirements.txt` will also install the latest `mkdocs-material`. There’s currently a cyclic dependency which we already intended to remove (although it hasn’t happened yet), and with that, this PR won’t work anymore.

Stanzilla:
The docs do say it is fine though https://www.mkdocs.org/dev-guide/themes/#creating-a-custom-theme maybe it's worth copying that hint they have there?

squidfunk:
I'd consider this as an edge case. I really don't want to overcomplicate the customization guide. Also, this behavior is going away as soon as `mkdocs-material-extensions` drops the cyclic dependency.

PR #2790, opened 2021-06-27
Use GitHub's Issue forms

Andre601:
For full context, read here:
- #2477

GitHub made their issue forms available as public Beta.
This PR implements those for this repository.

Noteworthy changes:

- The checkbox for agreeing to the contributing guidelines is its own section
- The other checkboxes are NOT required due to the possibility that it may not be doable
- "Configuration" in bug report and "Translation" in translate are automatically rendered as Codeblocks with their respective render (yaml and jinja)
- All text fields are required to be filled out.
- Descriptions and the "Warning" block at the top won't be displayed in the opened issue, but the title's of the sections will.

More info about issue forms:
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema

# Screenshots
- ## Bug Report
  image
  image
- ## Feature Request
  image
- ## Translation
  image


squidfunk:
Wow, thanks! That looks great! Is it still possible to post screenshots? This would be necessary, especially for feature requests.

Andre601:
> Wow, thanks! That looks great! Is it still possible to post screenshots? This would be necessary, especially for feature requests.

Text areas act as normal markdown fields (open the toolbar and preview thing) so people should be able to upload images.
Withthat said is there a suggestion open to add a type for uploading files.

squidfunk:
Thanks again!

squidfunk:
I've polished the issue templates a little. With the mandatory fields, we can get rid of the teaser text that reminded the user to fill out the template and not delete it. As this is now obligatory, there's no more need.

PR #2801, opened 2021-06-30
Point `keys` reference to `keys`, not `snippets`

NJAldwin:
I noticed in the docs that the link to browse the key codes actually links to the snippets extension.

I believe this was due to an accidental switch of the numbers used as link references.

Might I suggest at some point converting this page to use textual keys (e.g. `[keys-extension]` instead of `[14]`) to prevent such inadvertent swaps in the future?


squidfunk:
Thanks!

> Might I suggest at some point converting this page to use textual keys (e.g. [keys-extension] instead of [14]) to prevent such inadvertent swaps in the future?

Yes, you're right, but I couldn't settle on a naming scheme yet that is as small and concise as numbers 😅

PR #2804, opened 2021-07-03
Add OkHttp to list of successful OSS projects

LouisCAD:
OkHttp being over 40K stars, and well known among Java and Kotlin
developers, it makes sense to have it in the list.


squidfunk:
Thanks for your suggestion! It's already been listed for a long time under _Square_:
https://github.com/squidfunk/mkdocs-material/blame/master/README.md#L153

PR #2843, opened 2021-07-18
Add Spanish and Galician translation

flanciskinho:
added tags on es.html and gl.html translations files

- fixed "search.placeholder" on Galician file. Changed by the normative word from Galician. [Word in the Galician Royal Academy](https://academia.gal/dicionario/-/termo/procurar)
- fixed "source.file.date.created" on Spanish file. It has been changed because it was not translated.


squidfunk:
Thanks!

PR #2852, opened 2021-07-20
Adds missing translation for french language

orobardet:
Some strings were missing in the french translation, this PR adds them.


orobardet:
The `npm run build` removed the file `material/overrides/assets/javascripts/iconsearch_index.json` and added a `\n` at the end of `material/.icons/material/LICENSE`.
Since it does not seems to have any relation to the new translation string, I did not commit these changes.

But I can if needed.

squidfunk:
Thanks!

> The npm run build removed the file material/overrides/assets/javascripts/iconsearch_index.json and added a \n at the end of material/.icons/material/LICENSE.
Since it does not seems to have any relation to the new translations, I did not commit these changes.

Those changes don't belong here, that's correct. The icon index is only built when the twemoji database is found in the `venv` folder, which obviously isn't there on your installation:

https://github.com/squidfunk/mkdocs-material/blob/ec4de0d822c046253c5e13a8c37fe90f012bfd16/tools/build/index.ts#L250-L261

It's specific to Material for MkDocs' own site, and is not necessary for the distribution. __TL;DR: nevermind.__

PR #2855, opened 2021-07-22
Add DockerHub credentials instructions

sixhobbits:
This is needed because of lines 46-50 in the `.github/workflows/publish.yml` file

```
      - name: Login to DockerHub
        uses: docker/login-action@v1
        with:
          username: ${{ secrets.DOCKER_USERNAME }}
          password: ${{ secrets.DOCKER_PASSWORD }}
```

Another option is to remove those lines. I tried both options and they worked. I'm not really sure why it is necessary to log into DockerHub if we are only using GHCR, but I assume that those steps are there for a reason.

### DIFF
--- a/docs/insiders/getting-started.md
+++ b/docs/insiders/getting-started.md
@@ -59,9 +59,10 @@ comfortable self-hosting:
     2. Click on [Generate a new token][5]
     3. Enter a name and select the [`write:packages`][10] scope
     4. Generate the token and store it in a safe place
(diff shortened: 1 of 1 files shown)


sixhobbits:
fixed by https://github.com/squidfunk/mkdocs-material-insiders/pull/22

squidfunk:
Ah yeah, @fkorotkov was right, Docker Hub isn't even necessary 😊

PR #2875, opened 2021-07-27
Updated Korean translations

J-H-Cho:
Fixed `source.file.date.created` and added some missing Korean translations.


squidfunk:
Thanks! Great to see more languages add support for the new translation keys 😊

PR #2880, opened 2021-07-29
Create ci.yml

Zfqwb:


squidfunk:
We already have this workflow in place:
https://github.com/squidfunk/mkdocs-material/blob/master/.github/workflows/documentation.yml

PR #2903, opened 2021-08-02
Documented the use of line ranges in hl_lines

vedranmiletic:


squidfunk:
A great addition to the docs, thanks!

PR #2915, opened 2021-08-08
Prototype: scrollable content tabs

squidfunk:
__Improvement on the current content tabs implementation as described in https://github.com/facelessuser/pymdown-extensions/issues/1415__

The revised HTML I proposed in the issue is more complex, but now allows for making the tab labels container scrollable, which fixes tabs breaking into multiple lines on mobile. Here's a demo:

https://user-images.githubusercontent.com/932156/128626605-3997b3f6-2bdc-473b-a509-4e7710576551.mov

The new implementation uses more complex selectors, which increases the total payload. However, as selectors are co-located, they are expected to compress very well. In fact, the size of the new implementation is 4,1kb (716b gzipped), which is pretty good, as it's a compression ratio of ~83%. The current implementation weights 1,5kb (525b gzipped), which could be removed when we settle on the new implementation. Thus, the net increase in transfer size is currently around 200b.

The new tabs can be activated with:

``` yaml
markdown_extensions:
  - pymdownx.tabbed:
      alternate_style: true
```

And here's how it currently looks (i.e. why it needs to be improved):

Bildschirmfoto 2021-08-08 um 11 08 19


requirements.txt
```diff
@@ -22,5 +22,8 @@
 mkdocs>=1.1
 Pygments>=2.4
 markdown>=3.2
-pymdown-extensions>=7.0
+# pymdown-extensions>=7.0
 mkdocs-material-extensions>=1.0
+
+# Use feature branch temporarily
+git+https://github.com/facelessuser/pymdown-extensions.git@feature/tabbed-alternate
```
facelessuser, line 29: Can you change this to?

```
pymdown-extensions @ git+https://github.com/facelessuser/pymdown-extensions.git@feature/tabbed-alternate
```

facelessuser: Having trouble installing with `pip`

squidfunk: Fixed, I wasn't aware of that syntax.

facelessuser: Yeah, it's new-ish. I ran into it with a project at my work.

facelessuser: Setuptools doesn't like it in the old way.

squidfunk:
After the last discussion in https://github.com/facelessuser/pymdown-extensions/issues/1415, this PR seems to be nearly ready to be merged.

Things to be done:

1. Wait for next `pymdownx` release to include the new implementation
2. Adjust CSS to revised markup (rename `.tabbed-subcontent` to `.tabbed-block` as discussed in https://github.com/facelessuser/pymdown-extensions/issues/1415#issuecomment-895644094)
3. Issue new release and update documentation

If all goes well, the legacy tabs implementation will be removed in the next major release.

squidfunk:
I went over the source again and did a little clean up and think we have a working implementation. Since this is hidden behind a configuration flag of the new Tabbed implementation, it's safe to merge into `master`. When the next version of Pymdown Extensions is released, we can update the `requirements.txt` and the documentation and advertise the new implementation.

The old implementation will from then on be considered deprecated and is planned to be removed in v8. It's not expected that v8 arrives any time soon, though. v7 is pretty stable and v8 will probably be a deprecation-only release to remove some long-standing deprecations.

Note that the current implementation doesn't include the arrow indicators, but I'll revisit this topic in the coming weeks.

squidfunk:
Did some more refactoring, merging now.

PR #2971, opened 2021-08-24
Change default CI name in the docs

victorbnl:
I changed the name `ci` to `Deploy` in the section explaining how to publish the site on GitHub. I think it's a lot clearer. If you prefer lowercase, you can, of course, change it to `deploy`.


squidfunk:
Thanks for the PR. However, I don't see why this change is necessary. There is a code annotation explaining that the name can be changed to your liking, so it's pretty explanatory.

victorbnl:
It's true but I don't think it's just a matter of taste saying that `deploy` is clearer than `ci` and would therefore be much better. And while you don't see why `deploy` should be used instead of `ci`, I don't see why `ci` should be used neither. But as you want!

PR #2994, opened 2021-09-03
Update zh-Hant.html

Aqours:
Fix source.link.title


squidfunk:
Thanks for the PR. Could you elaborate why the new version is better? How do both values translate to English?

Aqours:
image

The repo maybe GitHub, GitLab or private git repo, it should not include immutable GitHub brand in the title.
English version is correct.

squidfunk:
Thanks, that was my first guess but it's always good to ask a native speaker. Please include an explanation like this in the PR description the next time, as this will save us both time 😊

PR #2996, opened 2021-09-03
Documented the displaying of metadata in page body

vedranmiletic:
Hopefully this is another useful addition to the documentation.

### DIFF
--- a/docs/reference/meta-tags.md
+++ b/docs/reference/meta-tags.md
@@ -85,6 +85,25 @@ extra:
 
 ## Customization
 
+### Displaying the metadata
+
+Sometimes it's useful to be able to display the medatada in the page body, e.g.
+print the name of the page author at the bottom of the page. To achieve that,
+you can [extend the theme][7] by adding the following contents to `main.html`:
+
+``` html
+{% extends "base.html" %}
+
+{% block content %}
+  {{ super() }}
+  {% if page and page.meta and page.meta.author %}
+    <p><small>Author: {{ page.meta.author }}</small></p>
+  {% endif %}
+{% endblock %}
+```
+
+  [7]: ../customization.md#extending-the-theme
+
 ### Custom meta tags
 
 #### on all pages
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! I see two issues with this:

1. It should go under the __customization__ section, because it involves theme extension.
2. Could you come up with a simpler example that doesn't involve so many lines? This would make it easier for users to grasp, which will reduce maintenance on my side.

squidfunk (changes requested):
See above.

vedranmiletic:
I can reduce the example, but add a warning admonition that directs the user to copy the contents of the block from `main.html` and then modify it in order to keep all the features of theme. Would that be acceptable?

squidfunk:
Another idea might be to just extend the content block and put stuff at the bottom. It might not be perfect, but it's easier to understand, like so:

``` jinja
{% block content %}
  {{ super() }}
  <!-- ... your additions ... -->
{% endblock %}
```

vedranmiletic:
Neat! Didn't know that trick. Let's do it like that then.

squidfunk:
LGTM, thanks!

vedranmiletic:
@squidfunk Was this intentionally removed from the docs? I think it's still an useful addition and doesn't duplicate the instructions in [metadata in headers section](https://squidfunk.github.io/mkdocs-material/reference/#using-metadata-in-templates).

squidfunk:
Yes, the meta tags reference page was moved to the beginning of the reference. The new [Using metadata in templates](https://squidfunk.github.io/mkdocs-material/reference/#using-metadata-in-templates) explains how to use custom front matter in templates, which is exactly what this PR included, albeit it's another example.

vedranmiletic:
The idea was to demonstrate how to display the metadata *in the page body*, not the header. It also showed how to use the `super()` so I don't think it's redundant for non-experts (your perspective as the author of the project who is very familiar with the code is understandably different).

squidfunk:
I understand the different use cases, but as it uses exactly the same principles, namely theme extension and template overrides, I'm considering those the same. I really don't want to provide redundant information, because it blows up the docs. If more users have trouble with this, I will reconsider adding it back to the docs.

PR #3086, opened 2021-10-05
Update docs link for web app manifests

ofek:
BTW, that option does not create a `manifest.json` for me, is it supposed to? Also, is this only relevant for `navigation.instant`?


squidfunk:
No, it’s only a link to a file that needs to be provided by the user. Im actually thinking of deprecating and removing it in favor of theme customization through `extrahead`.

ofek:
Ah okay 👍

Issue #3182, opened 2021-11-04
Please change Greek language code to "el"


alkisg:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the problem doesn't occur with the `mkdocs` or `readthedocs` themes
- [X] ... the problem persists when all overrides are removed, i.e. `custom_dir`, `extra_javascript` and `extra_css`
- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

Hi, the country "Greece" (or Hellas) has a domain of "gr".
The language "Greek" (or H**el**lenic), is used in both Greece and Cyprus, and has a language code of "el".
Reference: https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes

Please rename this file: https://github.com/squidfunk/mkdocs-material/blob/master/src/partials/languages/gr.html
to: https://github.com/squidfunk/mkdocs-material/blob/master/src/partials/languages/el.html
Also update its line "25" to "el".
Finally, update the documentation accordingly: https://squidfunk.github.io/mkdocs-material/setup/changing-the-language/

### Expected behaviour

The Greek language code should be "el".

### Actual behaviour

The Greek language code is "gr".

### Steps to reproduce

Select the "gr" language in the settings.

### Package versions

- Python: `python --version`
- MkDocs: `mkdocs --version`
- Material: `pip show mkdocs-material | grep -E ^Version`

### Configuration

```yaml
theme:
    name: material
    language: el
```

### System information

Ubuntu 20.04.

squidfunk:
Thanks for reporting. Could you craft a PR for this?

squidfunk:
Greek translations were added in #986. @cakrit – do you know why you've chosen `gr` as a language code?

alkisg:
Sure, I filed a PR in #3183.

Thanks for mkdocs-material, I haven't started using it yet; [my](https://ltsp.org) [sites](https://epoptes.org) are currently using just-the-docs, and I was considering to migrate to mkdocs or sphinx. When I saw mkdocs-material, I made up my mind very quickly!

I'm closing this issue, I think we can comment in the PR from now on...

squidfunk:
[...]

PR #3183, opened 2021-11-04
Use el ISO 639-1 code for Greek language

alkisg:
Hi, this PR renames the Greek language code from "gr" to "el", as defined in https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes.

The code "gr" is for the country (Greece, e.g. DNS TLD), while "el" is for the language (Greek, H**el**lenic), which is also spoken in Cyprus.

Relevant issue: #3182

### DIFF
--- a/docs/setup/changing-the-language.md
+++ b/docs/setup/changing-the-language.md
@@ -35,6 +35,7 @@ The following languages are supported:
 - `cs` – Czech
 - `da` – Danish
 - `de` – German
+- `el` – Greek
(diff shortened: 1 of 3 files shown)


squidfunk:
Thanks for providing the PR! This is a breaking change, so we should provide a fallback. I'll merge it anyway and add back the fallback, which will then be removed in the next major version.

squidfunk:
Added in f3c27fdaf.

Issue #3169, opened 2021-11-03
Send search analytics for Google Analytics 4


garrettheel:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I want to suggest an idea and checked that ...

- [X] ... to my best knowledge, my idea wouldn't break something for other users
- [X] ... the documentation does not mention anything about my idea
- [ ] ... there are no open or closed issues that are related to my idea

### Description

It'd be great if mkdocs-material would send search queries to GA4.

https://github.com/squidfunk/mkdocs-material/issues/1658 was the previous implementation of this for Universal Analytics (UA-*)

### Use Cases

This would help to understand how documentation is being used and improve the ranking of pages.

garrettheel:
@squidfunk I understand that it's a little tougher to set up on the Google Analytics side, but I think it would still be nice if mkdocs-material fired a custom event for search so that the data is somehow available. I suppose it can be hacked around with custom javascript, but would be a pretty common use case.

squidfunk:
If you would like to work on this, we can integrate it. Unfortunately, I currently don't have the bandwidth to do so. We would need a snippet to send the data, similar to the following code:

https://github.com/squidfunk/mkdocs-material/blob/b405149efd4118f6e02b2d639b784a7e9917e425/src/partials/integrations/analytics/google.html#L73-L81

And documentation, which explains how to configure Google Analytics to pick up the query. If you can provide both in a PR, I'm happy to merge it.

garrettheel:
@squidfunk created https://github.com/squidfunk/mkdocs-material/pull/3184

PR #3184, opened 2021-11-04
Send search analytics events for GA4

garrettheel:
Send a `search` event when using Google Analytics 4 properties. The docs are probably overkill since I am fairly sure that this works out of the box, but there's a long delay between sending an event and being able to see it in the GA UI.

Related to https://github.com/squidfunk/mkdocs-material/issues/3169

One related bug I noticed is that this event handler (for both UA and GA4) fires twice on every search (doubling the events produced). I didn't attempt to fix that here though.

### DIFF
--- a/docs/setup/setting-up-site-analytics.md
+++ b/docs/setup/setting-up-site-analytics.md
@@ -49,7 +49,8 @@ following lines to `mkdocs.yml`:
 
     Besides page views and events, [site search] can be tracked to better
     understand how people use your documentation and what they expect to find.
-    In order to enable site search tracking, the following steps are required:
+    In order to enable site search tracking, the following steps are required
+    for Universal Analytics (UA-XXXXXXXX-X) properties:
 
     1.  Go to your Google Analytics __admin settings__
     2.  Select the property for the respective tracking code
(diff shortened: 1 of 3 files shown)


docs/setup/setting-up-site-analytics.md
```diff
@@ -49,20 +49,24 @@ following lines to `mkdocs.yml`:
...
     1.  Go to your Google Analytics __admin settings__
     2.  Select the property for the respective tracking code
     3.  Go to the __view settings__ tab
     4.  Scroll down and enable __site search settings__
     5.  Set the __query parameter__ to `q`
 
-    Note that currently, site search tracking is not supported with Google
-    Analytics 4 due to the more complicated manual setup. If you want to set up
-    site search tracking yourself, [this tutorial][tutorial] is a good start.
+    For Google Analytics 4 (G-XXXXXXXXXX) properties:
```
squidfunk, line 61: Could you please move both guides into content tabs? Similar to the configuration section above the admonition 😊

src/partials/integrations/analytics/google.html
```diff
@@ -42,6 +42,14 @@
 
     /* Register virtual event handlers */
     document.addEventListener("DOMContentLoaded", function() {
+      if (document.forms.search) {
```
squidfunk, line 45: So I read that you verified this works, correct? If so, LGTM.

squidfunk (changes requested):
Many thanks for the PR! Just some minor docs changes.

src/partials/integrations/analytics/google.html
garrettheel: Yep, this worked for me.

docs/setup/setting-up-site-analytics.md
garrettheel: Sure thing

garrettheel: Screen Shot 2021-11-04 at 3 55 01 PM

squidfunk: Looks great!

squidfunk (approved):

squidfunk:
@garrettheel I've tried to access the site search data in GA4, but I can't find where the data ends up. Site search is activated and I'm seeing incoming events, but I don't see any site search dashboard. Could you please clarify what needs to be done to access the site search data in a meaningful way?

garrettheel:
@squidfunk I believe there's a significant delay between when you begin emitting events and when the "Search Terms" tab is populated (this seemed to be ~12-24h for me). Had that amount of time elapsed before you checked?

squidfunk:
Not yet, I'll keep waiting 🐌🤓

PR #3191, opened 2021-11-07
Update Greek translation

alkisg:
Hi, I updated the Greek translation.
There were some missing strings, but most importantly the next/previous words were translated as previous/next.


alkisg:
@squidfunk, I need a bit of help for the Greek search function.
My attempt above ^ fails with:

> WARNING  -  [18:45:12] "GET /assets/javascripts/lunr/min/lunr.el.min.js HTTP/1.1" code 404

If I change `"search.config.lang": "en",` instead of `el`, then I can search for English strings but not for Greek strings.

I might be able to send some patch for `el` to lunr in the following months, but for now, could we somehow have Greek search working even without stemming?

alkisg:
Got it, with this additional commit Greek search works now (without stemming).

src/partials/languages/el.html
```diff
@@ -34,9 +34,7 @@
   "meta.comments": "Σχόλια",
   "meta.source": "Πηγή",
   "nav.title": "Πλοήγηση",
-  "search.config.lang": "el",
-  "search.config.pipeline": "trimmer, stopWordFilter",
-  "search.config.separator": "[\s\-]+",
+  "search.config.pipeline": " ",
```
squidfunk, line 37: I think the `trimmer` and `stopWordFilter` should not be removed. My testing showed it also works with them.

squidfunk (changes requested):

alkisg:
By replacing:
> "search.config.pipeline": " ",

with:
> "search.config.pipeline": "trimmer, stopWordFilter",
  
Greek search stops working for me. I'll experiment more, but if you have any hints, they'll be most welcome. :)

alkisg:
With this, it does work:

>   "search.config.pipeline": "stopWordFilter",

alkisg:
As a test case, if I have the string:
"english ελληνικά"

then typing `en` in Search always matches the text, but typing `ελ` only works if `trimmer` has been removed.

squidfunk (approved):

PR #3207, opened 2021-11-10
Source branch for GitHub Pages has to be set

SchulteMarkus:
Using your awesome workflow, a branch gh-pages is being created, which serves as GitHub Pages source. This has to be set once in the repository settings.

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -91,6 +91,9 @@ Now, when a new commit is pushed to either the `master` or `main` branches,
 the static site is automatically built and deployed. Push your changes to see
 the workflow in action.
 
+After the first successful build, visit `github.com/<username>/<repository>/settings/pages`
+and set `Source (branch)` to `gh-pages`.
+
 Your documentation should shortly appear at `<username>.github.io/<repository>`.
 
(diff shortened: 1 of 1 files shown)


docs/publishing-your-site.md
```diff
@@ -91,6 +91,9 @@ Now, when a new commit is pushed to either the `master` or `main` branches,
 the static site is automatically built and deployed. Push your changes to see
 the workflow in action.
 
+After the first successful build, visit `github.com/<username>/<repository>/settings/pages`
```
squidfunk, line 94: Is this really necessary? When I tested it, it always automatically set the default branch, unless you changed it.

SchulteMarkus: Well I just had my first time setup and had to do this setting. I can replay it one a fresh repository, but I assume you want to check yourself before (if) applying this change.

squidfunk: I see. Could we maybe reformulate the instructions saying "if the page doesn't show up after a few minutes, check your `gh-pages` setting here...", so maybe more like troubleshooting advice.

squidfunk (approved):

squidfunk:
Thanks!

PR #3208, opened 2021-11-10
Some clarifications when using Docker

SchulteMarkus:
I chose "Docker-style" for first time setup, and noticed some missing and/or misleading information (at least for my understanding), such as:
- "PowerShell" is the Windows one?
- File permissions have to be taken care of
- For "build" command, C+P docker instructions had been missing

### DIFF
--- a/docs/creating-your-site.md
+++ b/docs/creating-your-site.md
@@ -14,13 +14,14 @@ mkdocs new .
 
 Alternatively, if you're running Material for MkDocs from within Docker, use:
 
-=== "Unix, Powershell"
+=== "Unix"
 
     ```
     docker run --rm -it -v ${PWD}:/docs squidfunk/mkdocs-material new .
+    sudo chown -R $UID:$UID ./*
     ```
 
-=== "Windows"
+=== "Windows, Powershell"
 
     ```
     docker run --rm -it -v "%cd%":/docs squidfunk/mkdocs-material new .
@@ -158,13 +159,13 @@ mkdocs serve # (1)
 
 If you're running Material for MkDocs from within Docker, use:
 
-=== "Unix, Powershell"
+=== "Unix"
 
     ```
     docker run --rm -it -p 8000:8000 -v ${PWD}:/docs squidfunk/mkdocs-material
     ```
 
-=== "Windows"
+=== "Windows, Powershell"
 
     ```
     docker run --rm -it -p 8000:8000 -v "%cd%":/docs squidfunk/mkdocs-material
@@ -183,9 +184,24 @@ Point your browser to [localhost:8000][live preview] and you should see:
 When you're finished editing, you can build a static site from your Markdown
 files with:
 
-```
+``` sh
 mkdocs build
 ```
+
+If you're running Material for MkDocs from within Docker, use:
+
+=== "Unix"
+
+    ```
+    docker run --rm -v ${PWD}:/docs squidfunk/mkdocs-material build
+    sudo chown -R $UID:$UID site/
+    ```
+
+=== "Windows, Powershell"
+
+    ```
+    docker run --rm -v "%cd%":/docs squidfunk/mkdocs-material build
+    ```
 
 The contents of this directory make up your project documentation. There's no
 need for operating a database or server, as it is completely self-contained.
--- a/docs/creating-your-site.md
+++ b/docs/creating-your-site.md
@@ -4,7 +4,7 @@ template: overrides/main.html
 
 # Creating your site
 
-After you've [installed] Material for MkDocs, you can bootstrap your project 
+After you've [installed] Material for MkDocs, you can bootstrap your project
 documentation using the `mkdocs` executable. Go to the directory where you want
 your project to be located and enter:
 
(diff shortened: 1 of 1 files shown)


docs/creating-your-site.md
```diff
@@ -14,13 +14,14 @@ mkdocs new .
...
-=== "Unix, Powershell"
+=== "Unix"
 
     ```
     docker run --rm -it -v ${PWD}:/docs squidfunk/mkdocs-material new .
+    sudo chown -R $UID:$UID ./*
     ```
 
-=== "Windows"
+=== "Windows, Powershell"
```
squidfunk, line 24: I'm not a Windows user, but as I understand there's WIndows Command Prompt and Windows Powershell, the latter of which has the same behavior as Unix (macOS, Linux)

docs/creating-your-site.md
```diff
@@ -183,10 +184,25 @@ Point your browser to [localhost:8000][live preview] and you should see:
...
 mkdocs build
 ```
 
+If you're running Material for MkDocs from within Docker, use:
+
+=== "Unix"
+
+    ```
+    docker run --rm -v ${PWD}:/docs squidfunk/mkdocs-material build
+    sudo chown -R $UID:$UID site/
```
squidfunk, line 197: Why is the `chown` necessary at this point? I'd say we drop it because otherwise some users might be confused.

squidfunk (changes requested):
Thanks for the PR! Please see my comments.

docs/creating-your-site.md
SchulteMarkus: Well I have to say: Don't know.
But if I'm not wrong, there is no such thing as "Powershell" at Unix.

docs/creating-your-site.md
squidfunk: I've just tested it, and Docker it looks like the user/group running the Docker command is just forwarded to Docker, see:

```
> docker run --rm -it -p 8000:8000 -v ${PWD}:/docs squidfunk/mkdocs-material build
INFO     -  Cleaning site directory
INFO     -  Building documentation to directory: /docs/site
INFO     -  Documentation built in 0.81 seconds
> ls -lha site                         
total 80
drwxr-xr-x  11 squidfunk  staff   352B 10 Nov 16:11 .
drwxr-xr-x  13 squidfunk  staff   416B 10 Nov 16:11 ..
-rw-r--r--   1 squidfunk  staff   9,6K 10 Nov 16:11 404.html
drwxr-xr-x   5 squidfunk  staff   160B 10 Nov 16:11 assets
drwxr-xr-x   4 squidfunk  staff   128B 10 Nov 16:11 bar
-rw-r--r--   1 squidfunk  staff   279B 10 Nov 16:11 custom.css
drwxr-xr-x   3 squidfunk  staff    96B 10 Nov 16:11 foo
-rw-r--r--   1 squidfunk  staff    13K 10 Nov 16:11 index.html
drwxr-xr-x   3 squidfunk  staff    96B 10 Nov 16:11 search
-rw-r--r--   1 squidfunk  staff   609B 10 Nov 16:11 sitemap.xml
-rw-r--r--   1 squidfunk  staff   197B 10 Nov 16:11 sitemap.xml.gz
```

docs/creating-your-site.md
squidfunk: No there isn't, that's correct. It might be a little misleading actually, but I didn't want to copy the same command 4 times for: Linux, macOS, Windows, Windows Powershell. That's overkill. The first tab reads as "Unix OR Powershell"

docs/creating-your-site.md
SchulteMarkus: Hmm, different here. I am using Linux Mint, which is an Ubuntu fork.

``` bash
$ docker run --rm -v ${PWD}:/docs squidfunk/mkdocs-material build # fresh+clean project
INFO     -  Cleaning site directory
INFO     -  Building documentation to directory: /docs/site
INFO     -  Documentation built in 0.19 seconds
$ ls -l site/
insgesamt 68
-rw-r--r-- 1 root root  8092 Nov 10 16:14 404.html
drwxr-xr-x 5 root root  4096 Nov 10 16:14 assets
-rw-r--r-- 1 root root 10813 Nov 10 16:14 index.html
drwxr-xr-x 2 root root  4096 Nov 10 16:14 search
-rw-r--r-- 1 root root   234 Nov 10 16:14 sitemap.xml
-rw-r--r-- 1 root root   188 Nov 10 16:14 sitemap.xml.gz
```

docs/creating-your-site.md
SchulteMarkus: Ahhh, I got it. Well, at least I got it wrong.
Maybe change to "Unix/Powershell | Windows". Or maybe reduce to Unix command.

squidfunk:
After a short private chat, we decided to leave it as it is.

SchulteMarkus:
Btw., Linux/Docker users should be aware of file permissions problem when mounting into a container.

squidfunk:
Using Linux, I'd say that Linux users should be well aware of file permissions 😉 I'd rather try to keep the docs as short and concise as possible. If it appears to be a repeating problem, we can add something to the docs, but until further notice, I'll leave the `chown` and `chmod` commands as an exercise to the user. I'm actually quite astonished that Docker doesn't forward permissions and sets them to `root`!? Maybe something to raise as a problem to Docker?

SchulteMarkus:
I totally agree.
Having file ownership set to root (UID 0) is correct, because the user within the docker container (which shouldn't know anything about the host including the executing user) is root (unless changed in Dockerfile). I'm more kind of surprised, that Docker under MacOs behaves differently. As "run container with mounted host volume from non-root" is quite a usual workflow, there should be something in Docker supporting this without workarounds. There are tons of workarounds, but not an elegant/built-in solution to this issue as far as I know.

Issue #3219, opened 2021-11-13
🚀 Material for MkDocs 8 – Beta


squidfunk:
__With the help of my awesome sponsors, I'm happy to announce that the '[Ghost Pepper](https://squidfunk.github.io/mkdocs-material/insiders/#4000-ghost-pepper)' funding goal has been reached, which means that [code annotations](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#adding-annotations), [anchor tracking](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#anchor-tracking) and [versioning warnings](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/#version-warning) are finding their way into this repository!__

Since those are some pretty nifty features, it's a good time to wrap them up into a new major release and drop some long-standing deprecated features. This issue will track the progress on the road from the beta to the final release of version 8.

__If you experience any problems, for now, please report them as a comment in this issue__.

## Installation

```
pip install mkdocs-material==8.0.0b2
```

Note that this is a beta release.

### Added

- [__Code annotations__](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#code-annotations): this is probably the most anticipated new feature. After finishing the merge and deprecations, I'll revisit the [placement issues on mobile](https://github.com/squidfunk/mkdocs-material/issues/2475). Those need to get fixed before we push out the major release.

- [__Version warning__](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/#version-warning) and [__Anchor tracking__](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#anchor-tracking)
### Improved

- __Content partials__: there's now a new partial, which allows fpr customizing the appearance of the Markdown content without overriding `base.html`. This makes it easy to add a further button to the top of a page or to add arbitrary content to the footer of a page:

    ``` jinja2
    {% block content %}
      {% include "partials/content.html" %}
    {% endblock %}
    ```

- __Copyright partial__: the copyright has been moved into a separate partial, which lifts the need to override the `footer.html` template when changing the copyright information.

    ``` jinja2
    <div class="md-footer-meta md-typeset">
      <div class="md-footer-meta__inner md-grid">
        {% include "partials/copyright.html" %}

        <!-- Social links -->
        {% if config.extra.social %}
          {% include "partials/social.html" %}
        {% endif %}
      </div>
    </div>
    ```

[...]

PR #3224, opened 2021-11-14
added missing Italian localization

aborruso:


src/partials/languages/it.html
```diff
@@ -41,5 +41,22 @@
   "source.revision.date": "Ultimo aggiornamento",
   "source.file.date.updated": "Ultimo aggiornamento",
   "source.file.date.created": "Creata",
-  "toc.title": "Indice"
+  "toc.title": "Indice",
+  "direction": "ltr",
```
squidfunk, line 45: Could you please put the keys in the same order as in the `en.html`?

squidfunk (changes requested):

src/partials/languages/it.html
aborruso: Sure. Thank you

https://github.com/squidfunk/mkdocs-material/blob/c67f6007de983764edfafaa6b0b908af68986a21/src/partials/languages/it.html

src/partials/languages/it.html
```diff
@@ -23,23 +23,40 @@
...
   "edit.link.title": "Modifica",
   "footer.previous": "Precedente",
   "footer.next": "Prossimo",
+  "footer.title": "Piede",
+  "header.title": "Intestazione",
   "meta.comments": "Commenti",
   "meta.source": "Sorgente",
+  "nav.title": "Navigazione",
   "search.config.lang": "it",
+  "search.config.pipeline": "trimmer, stopWordFilter",
```
squidfunk, line 38: Almost perfect :-) This line and the next are not necessary, as they are the same as in `en.html`. They can be removed. Otherwise, this is ready for merge!

squidfunk (changes requested):

src/partials/languages/it.html
aborruso: Done, thank you

squidfunk:
Perfect, thanks again!

aborruso:
> Perfect, thanks again!

it is a pleasure and an honor

aborruso:
> Perfect, thanks again!

A question: I have a [GitHub action](https://github.com/ondata/guidaVisiData/blob/master/.github/workflows/ci.yml) to deploy in GitHub. If I run it I will have these new localizations? Must I force this in some way?

Thank you

squidfunk:
If you clone from git, yes. Otherwise no, because the changes are not released yet. As we're on our way to version 8, I'm going to release those changes in `8.0.0b2`, which should follow this weekend. In the mean time you can install from `git`:

```
pip install git+https://github.com/squidfunk/mkdocs-material.git
```

aborruso:
> If you clone from `git`, yes. Otherwise no, because the changes are not released yet. As we're on our way to version 8 (see #3219), I'm going to release those changes in `8.0.0b2`, which should follow this weekend. 

@squidfunk we have the `8.0.0b2`, but I have still the "Back to top" English label. Is it normal?

This is my [github action](https://github.com/ondata/guidaVisiData/blob/master/.github/workflows/ci.yml).

Thank you

image

squidfunk:
It's still only in `master`, not in the `v8` branch. I'll merge it as soon as that is possible. In the meantime you can either use `master` when you want to use 7.3.6, or use custom translations when you want to use 8.0.0b2.

aborruso:
I don't think I understand. Sorry. In my github action I install simply ` mkdocs-material ` via pip. In this case I guess then the master is installed. You say that it's still only in the master, but I do not have these new Italian labels on my site?

What I did not understand? What am I doing wrong?

Thank you

martinbira:
> I don't think I understand. Sorry. In my github action I install simply `mkdocs-material` via pip. In this case I guess then the master is installed. You say that it's still only in the master, but I do not have these new Italian labels on my site?
> 
> What I did not understand? What am I doing wrong?
It is in the master for version `8.x.x`, which is not released yet, so it is not available in the latest `7.3.6` release that you get from installing `mkdocs-material`. It will be available when version 8 is released.

squidfunk:
The changes were commited to `master` after the 7.3.6 release, but `master` was not merged into `8.0.0` yet. Thus, you have to either install from git or use custom translations for the time being.

PR #3231, opened 2021-11-16
Update zh.html

jasonren0403:
Fix missing Simplified Chinese translations (for mkdocs-material v8).
Noticed the missing translation problem after viewing the discussion #3223 .
I've finished most of the translations except the `search.result.term.missing` key, I haven't got an idea what it is use for.


squidfunk:
Thanks! `search.term.missing` is printed before terms not found in the current result:

Bildschirmfoto 2021-11-16 um 10 12 12

jasonren0403:
> Thanks! `search.term.missing` is printed before terms not found in the current result:
> 
> Bildschirmfoto 2021-11-16 um 10 12 12

All done. 😆

PR #3236, opened 2021-11-20
Minimal configuration is not enough for a build

MarlinMr:
The build fails due to the lack of `site_name`

### DIFF
--- a/docs/creating-your-site.md
+++ b/docs/creating-your-site.md
@@ -62,6 +62,7 @@ slightly different:
 === ":fontawesome-brands-git-alt: git"
 
     ``` yaml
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. However, the [section](https://squidfunk.github.io/mkdocs-material/creating-your-site/#minimal-configuration) states that you should **add** the lines to `mkdocs.yml`, because in [the step before](https://squidfunk.github.io/mkdocs-material/creating-your-site/) a minimal `mkdocs.yml` was already generated. It's in fact the minimal configuration for the theme to work, not for MkDocs itself.

PR #3252, opened 2021-11-23
Account for empty versions

ofek:
Essentially, I really want the default version to be at the root and the versioned docs to live alongside it.

I'd manage `versions.json` myself so it (and dropdown) would look like:

```json
[
    {"version": "dev", "title": "dev", "aliases": []},
    {"version": "", "title": "1.1", "aliases": []},
    {"version": "1.0", "title": "1.0", "aliases": []}
]
```

Then when `1.2` is released:

```json
[
    {"version": "dev", "title": "dev", "aliases": []},
    {"version": "", "title": "1.2", "aliases": []},
    {"version": "1.1", "title": "1.1", "aliases": []},
    {"version": "1.0", "title": "1.0", "aliases": []}
]
```

### DIFF
--- a/src/assets/javascripts/templates/version/index.tsx
+++ b/src/assets/javascripts/templates/version/index.tsx
@@ -76,10 +76,12 @@ export function renderVersionSelector(versions: Version[]): HTMLElement {
   const config = configuration()
 
   /* Determine active version */
-  const [, current] = config.base.match(/([^/]*)\/?$/)!
+  const [, current] = config.base.match(/([^/]+)\/?$/)!
   const active =
     versions.find(({ version, aliases }) => (
       version === current || aliases.includes(current)
+    )) || versions.find(({ version }) => (
+      version === ""
     )) || versions[0]
 
   /* Render version selector */
--- a/src/assets/javascripts/templates/version/index.tsx
+++ b/src/assets/javascripts/templates/version/index.tsx
@@ -76,7 +76,7 @@ export function renderVersionSelector(versions: Version[]): HTMLElement {
   const config = configuration()
 
   /* Determine active version */
-  const [, current] = config.base.match(/([^/]+)\/?$/)!
+  const [, current] = config.base.match(/([^/]*)\/?$/)!
   const active =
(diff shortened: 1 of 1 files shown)


ofek:
If accepted, could this get in 8.0.0? 🙂

squidfunk:
Thanks for the PR! So if I understand correctly, this will essentially just redirect to the root, right? Is this supported by `mike`? This won't be included in `8.0.0` – I'll merge and evaluate it once the next major version stabilizes. I expect quite some issues because there are so many changes, so I'd kindly ask you for some more time.

squidfunk:
I'm converting this to a draft, since you still seem to be working on this.

ofek:
Hey sorry all done, I just fixed the logic so it both works now and is more explicit. This is not yet supported by `mike` so it'd be manual for now but this would be a prerequisite for such support in both cases.

By next version do you mean adding this feature to 8.1.0?

squidfunk:
So `mike` won't add support for it – I guess this PR can be closed then?

- https://github.com/jimporter/mike/issues/83

ofek:
It would still allow users to manually do this, I'm writing a script for this

ofek:
I got the build working https://ofek.dev/hatch/ then go to `/dev/` and notice the dropdown appear!

So, just waiting on this now 😄

ofek:
Hmm, I tried copying the new code to `site-packages/material/assets/javascripts/bundle.b1047164.min.js.map` but it doesn't make the dropdown appear

squidfunk:
I'm sorry, but I'm currently not intending to merge this. The thing is that you'll need a custom workflow (= your custom script) to patch `versions.json` and it's not the canonical way that `mike` supports, so you're effectively fighting `mike`. However, `mike` is currently the only versioning solution we support, and there's no mention that we support custom flows. This might change in the future, though. At the present time, it's not something I'd want to maintain. I beg for your understanding. As an alternative, as @jimporter already suggested, you might use webserver-level rewrites to achieve almost the same thing.

PR #3267, opened 2021-11-29
Update node to 16 and npm to 7+ lockfile

Stanzilla:
Updates the `engine` property to node v16 which is now LTS and generates a new npm lockfile in the version 7+ format.


squidfunk:
Superseded by e585216.

PR #3270, opened 2021-11-29
add meta data to require python >=3.6

ipmsteven:
[MkDocs has dropped support for Python 3.5 in 1.2](https://www.mkdocs.org/about/release-notes/#other-changes-and-additions-to-version-12)


squidfunk:
Thanks! Good idea!

PR #3276, opened 2021-11-30
top.title localization was missing in ru.html

arterm-sedov:
The "Back to top" button was displaying in English in the Russian theme

### DIFF
--- a/material/partials/languages/ru.html
+++ b/material/partials/languages/ru.html
@@ -21,4 +21,5 @@
   "source.file.date.updated": "Последнее обновление",
   "source.file.date.created": "Созданный",
   "toc.title": "Содержание"
+  "top.title": "К началу"
 }[key] }}{% endmacro %}


squidfunk:
Thanks! Could you also provide the translations for the other missing keys? The best idea is to check which translations are missing from `en.html`:

https://github.com/squidfunk/mkdocs-material/blob/ff4cf19f8d72234945251001006d9acbf1541e22/src/partials/languages/en.html#L24-L60

squidfunk (changes requested):
Please also provide the other missing translations.

arterm-sedov:
> Please also provide the other missing translations.

Will do a bit later. Thanks for the hint.

arterm-sedov:
> Thanks! Could you also provide the translations for the other missing keys? The best idea is to check which translations are missing from `en.html`:
> 
> https://github.com/squidfunk/mkdocs-material/blob/ff4cf19f8d72234945251001006d9acbf1541e22/src/partials/languages/en.html#L24-L60

Sure, I will translate them

arterm-sedov:
I've committed the requested changes in a new pull-request today

squidfunk:
Thanks! Superseded by #3298.

Issue #3170, opened 2021-11-03
"Was this page helpful" does not work with Google Analytics 4 properties


garrettheel:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the problem doesn't occur with the `mkdocs` or `readthedocs` themes
- [X] ... the problem persists when all overrides are removed, i.e. `custom_dir`, `extra_javascript` and `extra_css`
- [X] ... the documentation does not mention anything about my problem
- [ ] ... there are no open or closed issues that are related to my problem

### Description

When using GA4 analytics properties, the "Was this page helpful?" feature does not work. Reading the code, it looks like this is currently only implemented for Universal Analytics.

Also mentioned in https://github.com/squidfunk/mkdocs-material/issues/3011

### Expected behaviour

I expect to see the widget at the footer of the page.

### Actual behaviour

The div is `hidden`, since there is no implementation for this analytics provider.

### Steps to reproduce

1. Use a GA4 analytics provider and switch on the feedback feature

2. Observe that there is no "was this page helpful" widget displayed

### Package versions

- Material: `7.3.6+insiders.3.1.5`

### Configuration

```yaml
# ...
extra:
  analytics:
    provider: google
    property: G-...
    feedback:
      title: Was this page helpful?
      # ... 
```

### System information

N/A

squidfunk:
I've added the sending of the `feedback` event data for GA4 in [`bfdd5c670`](https://github.com/squidfunk/mkdocs-material-insiders/commit/bfdd5c67030112c87824abec20ec0939dd1f9593), but I yet have to find out how to build a sensible custom report from this.

squidfunk:
I had to reissue the latest release (by fault), so this has been released as part of 7.3.6+insiders-3.2.2. I'll let this open until we find out how to build a sensible custom report from this.

squidfunk:
@garrettheel did you manage to create a custom report in GA4?

garrettheel:
> @garrettheel did you manage to create a custom report in GA4?

Not yet, still contending with the confusing UI and lag in data becoming available. I'll keep trying and report back

garrettheel:
[...]

PR #3277, opened 2021-11-30
Add GA4 docs for page helpful

garrettheel:
Related: https://github.com/squidfunk/mkdocs-material/issues/3170

Adds docs for setting up a GA4 report to explore feedback data collected via mkdocs-material


squidfunk:
Many thanks!

PR #3279, opened 2021-11-30
fix translations for sv and permissions for pl

NeilHanlon:
Hi folks - ran into this while deploying the Rocky Linux docs site.

I checked over the other files with a hacky one-liner to parse out the JSON and the rest seem okay!

Thank you,
Neil


squidfunk:
Thanks! Polish seems to be the only one being 755 – weird.

NeilHanlon:
hah yeah, I ran an `ls` and said "weird" :) 

Thanks for the quick merge!

PR #3329, opened 2021-12-10
*try* to add giscus tutorial

Andre601:
This is a - in my personal opinion really bad - approach at adding a tutorial for how to Integrate Giscus into Material for MkDocs.

The main issues I have with this is, that you still heavily rely on the Giscus site itself to get values such as a Repo and Discussion ID, which I feel like isn't really an optimal approach.

I also don't think that it is well made in terms of structure and readability, so any help on improving this further is greatly appreciated.

Related discussion: #3266

### DIFF
--- a/docs/setup/adding-a-comment-system.md
+++ b/docs/setup/adding-a-comment-system.md
@@ -5,22 +5,22 @@ template: overrides/main.html
 # Adding a comment system
 
 Material for MkDocs allows to easily add the third-party comment system of your
-choice to the footer of every page by using [theme extension]. As an example,
-we'll be integrating [Disqus] a wildly popular comment provider, but others
-can be integrate with the same principles
+choice to the footer of every page by using [theme extension]. On this page do
+we show two examples using [Disqus] and [Giscus] respectively.
 
   [Disqus]: https://disqus.com/
+  [Giscus]: https://giscus.app/
 
 ## Customization
 
-### Disqus integration
+### Extending the theme
 
 In order to integrate a third-party comment provider offering a JavaScript-based
 solution, follow the guide on [theme extension], copy the contents from the
 [`content.html`][content partial] partial and create a file at the same location
 in the `overrides` folder:
 
-=== ":octicons-file-code-16: overrides/partials/content.html"
+=== ":octicons-file-code-16: Disqus integration"
 
     ``` html
     <!-- Add copied contents from original content.html here -->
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! I'll try to look into this when I find some time 😊 If you manage to streamline the setup in the meantime, feel free to update the PR.

Andre601:
> Thanks for the PR! I'll try to look into this when I find some time 😊 If you manage to streamline the setup in the meantime, feel free to update the PR.

Yeah... Sadly can't think of a good solution. Only alternative I can think of would be having like this in the mkdocs (and by extension in frontmatter):
```yaml
extra:
  giscus:
    repo: 'someUser/repo'
    repo-id: '....' # Needs a reliable way to get the "id" giscus generates/uses.
    # ... All the other options
```
But issue I see here would be to have a ton of setters and ifs for each value that can be set...
If there was a more easy way to retrieve the actual values, would I be really happy for it, but as of right now is it rather tedious...

Also, small fun-fact I guess? But when decoding the Base64 for `data-repo-id` for my Repository (Andre601/blog) did I receive the following value:
```
010:Repository323123424
```
And testing it with `giscus/giscus` returned
```
010:Repository351958053
```

So my guess here is, that the "id" is just `010:Repository` followed by the Repository ID from GitHub? Iirc does GitHub have numerical Ids right?

Kanaduchi:
Can we hame more common usage in mkdocs and single pages?

For example:
in mkdocs.yaml
```
extra:
  comment-system:
    type: 'disqus' or 'giscus'
    name: 'short-name' # Needs for disqus
    repo: 'someUser/repo' # Needs for giscus
    repo-id: '....' # Needs a reliable way to get the "id" giscus generates/uses
    # ... All the other options
```

on a single page:
```
comment-system: show / hide
```
or the same as in mkdocs if user needs some personal settings

Andre601:
> Can we hame more common usage in mkdocs and single pages?
> 
> For example: in mkdocs.yaml
> 
> ```
> extra:
>   comment-system:
>     type: 'disqus' or 'giscus'
>     name: 'short-name' # Needs for disqus
>     repo: 'someUser/repo' # Needs for giscus
>     repo-id: '....' # Needs a reliable way to get the "id" giscus generates/uses
>     # ... All the other options
> ```
> 
> on a single page:
> 
> ```
> comment-system: show / hide
> ```
> 
> or the same as in mkdocs if user needs some personal settings

You could try and add that for yourself through the theme extension system...

I can't think of a proper and efficient way here, mainly because I barely use Jinja2 to begin with and don't want to touch it when not needed, and honestly don't have any real motivation in trying to find something on my own... Still waiting on squidfunk's input here...

squidfunk:
I'll get to it eventually. The TODO list for this project is currently pretty packed.

squidfunk:
[I've just set up Giscus](https://squidfunk.github.io/mkdocs-material/blog/2021/search-better-faster-smaller/#whats-next) for the blog section of the documentation and am pretty happy with the result. Using [giscus.app](https://giscus.app) for generating the snippet is the simplest way, so there's no need to add configuration, as comment systems need theme extension anyway. Furthermore, all comment systems have different configs, so there's no point in abstracting it.

Here's all that I did:

https://github.com/squidfunk/mkdocs-material/blob/5842b0a7b36e298de56740d834dc8bd76bf62965/src/overrides/blog.html#L26-L46

I'll restructure the "adding a comment system" documentation for Giscus and throw out Disqus.

Andre601:
> [I've just set up Giscus](https://squidfunk.github.io/mkdocs-material/blog/2021/search-better-faster-smaller/#whats-next) for the blog section of the documentation and am pretty happy with the result. Using [giscus.app](https://giscus.app) for generating the snippet is the simplest way, so there's no need to add configuration, as comment systems need theme extension anyway. Furthermore, all comment systems have different configs, so there's no point in abstracting it.
> 
> Here's all that I did:
> 
> https://github.com/squidfunk/mkdocs-material/blob/5842b0a7b36e298de56740d834dc8bd76bf62965/src/overrides/blog.html#L26-L46
> 
> I'll restructure the "adding a comment system" documentation for Giscus and throw out Disqus.

IMO does this only add extra work for the person. Whenever they add a new page would this automatically require them to add another extension to add giscuss, or require them to make sure they use the right template here... Not really optimal imo.

Tho, one solution for them could be to just have the same template twice, but one with and one without the integration and then define the template through frontmatter... Not sure how user friendly this one would be in the end...

Andre601:
Also, by that logic could the same be applied to Disqus. Why go do all this abstraction when you can just theme-extend it anyway?
Most people won't have more than one Disqus instance anyway, so no point in adding all the frontmatter stuff for it...

This could simplify the setup guide in the long run by having the same setup for both with the difference being the extra settings on Giscus' end.

squidfunk:
> Also, by that logic could the same be applied to Disqus. Why go do all this abstraction when you can just theme-extend it anyway?

We removed the Disqus integration for that exact reason - you _must_ use theme extension to add it. Theme extension is much more flexible than maintaining tons of flags in `mkdocs.yml`. You can decide yourself whether you want it on every page (just extend `main.html`) or on specific pages (create a new template that extends from `main.html`).

squidfunk:
9511569d0 adds dark mode support, reloading the page when the toggle is clicked. I'm not sure there's another way to reload the Giscus integration, so reloading it is.

Andre601:
> > Also, by that logic could the same be applied to Disqus. Why go do all this abstraction when you can just theme-extend it anyway?
> 
> We removed the Disqus integration for that exact reason - you _must_ use theme extension to add it. Theme extension is much more flexible than maintaining tons of flags in `mkdocs.yml`. You can decide yourself whether you want it on every page (just extend `main.html`) or on specific pages (create a new template that extends from `main.html`) like I did with the new `blog.html` template in `overrides`.

My point is, that your current guide of Disqus integration recommends the previous/old system you yourself say isn't that good, which kinda is weird tbh.

I need to think of restructuring the Guide... Probably make Disqus and Giscus their own sections on the page, rather than tabs. Otherwise could it become a bit much.

> Additionally, as you realized yourself, we cannot infer the Giscus `data-*` attributes, some of them look very cryptic or are base64 encoded. Thus, referring the user to use the configuration tool they offer, and just copy & paste the result is much simpler and leads to less maintenance on our side.

The Giscus dev mentioned to me, that this data stuff is now pretty much pointless to have since it is now handled server-side on his end for the main Giscus app, so there might be a future update that removes this need altogether, which can be beneficial. But there is no guarantee for this right now...

squidfunk:
> My point is, that your current guide of Disqus integration recommends the previous/old system you yourself say isn't that good, which kinda is weird tbh.

As I said – I'm about to restructure the documentation. Disqus will be removed, as it became too horrible, IMHO. So it's still a work in progress.

> The Giscus dev mentioned to me, that this data stuff is now pretty much pointless to have since it is now handled server-side on his end for the main Giscus app, so there might be a future update that removes this need altogether, which can be beneficial. But there is no guarantee for this right now...

I used the guide on their page which added the `data-*` attributes. Let's go with the official way. If that changes, we can change it.

Andre601:
I tried to simplify the docs a bit... Not 100% happy with it, so feel free to tweak it...

squidfunk:
Superseded by f9e941846.

PR #3331, opened 2021-12-11
Update images to match examples

Andre601:
I updated the images in the reference section to actually reflect the examples given. I also changed the image colour to a brighter grey, using the colour palette mentioned [here](https://material.io/design/color/the-color-system.html). This was to make the image more noticeable because while it can be seen easily on the dark theme, is it barely any different from the white background on the light theme.

I hope the colour combinations are fine the way they are...

### DIFF
--- a/docs/reference/images.md
+++ b/docs/reference/images.md
@@ -42,12 +42,12 @@ respective alignment directions via the `align` attribute, i.e. `align=left` or
     _Example_:
 
     ``` markdown
-    ![Placeholder](https://dummyimage.com/600x400/eee/aaa){ align=left }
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. I'd consider this a pure matter of taste, and to be honest, I'd rather leave it as it is.

Andre601:
Okay. Your choice after all. Just found it a bit bad that the examples only use the `<width>x<height>` but the result uses `- image -`.

squidfunk:
Yes, because it's a placeholder image after all. Remember that there are many non-technical users out there who might have a difficult time understanding what the additional parameters mean, yet they are irrelevant to the problem at hand.

Issue #3361, opened 2021-12-18
Section index pages seems not work


aborruso:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the problem doesn't occur with the `mkdocs` or `readthedocs` themes
- [X] ... the problem persists when all overrides are removed, i.e. `custom_dir`, `extra_javascript` and `extra_css`
- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

I have set these features [here](https://github.com/ondata/guidaPraticaPubblicazioneCSV/blob/main/mkdocs.yml#L33-L40):

```
  features:
    - navigation.top
    - navigation.tabs
    - navigation.sections
    - search.highlight
    - search.share
    - search.suggest
    - navigation.tracking
 ```

And this is [my navigation](https://github.com/ondata/guidaPraticaPubblicazioneCSV/blob/main/mkdocs.yml#L47-L57):

```
nav:
    - Home: README.md
    - Guida:
      - Introduzione: introduzione/README.md
      - Formato di dati tabellari: formato_dati_tabellari/README.md
      - Dizionario: dizionario_dati/README.md
      - Linee guida pubblicazione:
          - linee_guida_pubblicazione_file/index.md
          - Singola riga intestazione: linee_guida_pubblicazione_file/P01_unica_riga_intestazione.md
    - Partecipa: partecipa.md
    - Note: note_redazione.md
```

Why `Linee guida pubblicazione` is not linked to the section index file, and I have the index file below the section name in the tree?

![image](https://user-images.githubusercontent.com/30607/146643377-0d88cfbc-7fa6-4c54-9525-0a07561fbbb3.png)

### Expected behaviour

To have the section in tree linked to index file

### Actual behaviour

The section is not linked to the index file

### Steps to reproduce

I have inserted all my source file.

This is the repo https://github.com/ondata/guidaPraticaPubblicazioneCSV

### Package versions

- Python: `python --version` 3.9.2
- MkDocs: `mkdocs --version` 1.2.3
- Material: `pip show mkdocs-material | grep -E ^Version` 8.1.0

### Configuration

```yaml
site_name: Guida per la pubblicazione di CSV
site_url:  https://ondata.github.io/guidaPraticaPubblicazioneCSV/
[...]

PR #3362, opened 2021-12-19
Added a warning about index file and sections

aborruso:
Related to:

> Material for MkDocs only supports a single index.md (or README.md which is the same for MkDocs). If you rename the files of those pages they will reappear.

https://github.com/squidfunk/mkdocs-material/issues/3361#issuecomment-997282688


squidfunk:
Thanks for the PR! Superseded by f3926bf44 fixes.

Issue #2435, opened 2021-03-13
Docker vulnerabilities in the docker alpine base image used - easier to upgrade the dist 


tan31989:
- [x] I've read the [contribution guidelines][1] and agree with them

__I've found a bug and checked that ...__

- [x] ... the problem doesn't occur with the default MkDocs template
- [x] ... the problem is not in any of my customizations (CSS, JS, template)
- [x] ... the documentation does not mention anything about my problem
- [x] ... there are no open or closed issues that are related to my problem

## Description

The Docker has some vulnerabilities which seem to fail our build while running the scan. The best way to fix such high vulnerabilities is to keep the dist upgraded by default. We use a known scanner [trivy](https://github.com/aquasecurity/trivy)

### Expected behavior

When we run a trivy scan with the following command:

```sh
# docker build -t mkdocs-image .
# trivy image mkdocs-image
```

Then we should get 0 vulnerabilities:

`Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)`

### Actual behavior

The current OpenSSL and other libraries are not upgraded as per the base alpine image used. So we get some high vulnerabilities.

### Steps to reproduce the bug

```sh
# docker build -t mkdocs-image .
# trivy image mkdocs-image
```
You will find the below errors:

```
mkdocs-image (alpine 3.12.3)
====================
Total: 6 (UNKNOWN: 0, LOW: 2, MEDIUM: 2, HIGH: 2, CRITICAL: 0)

+--------------+------------------+----------+-------------------+---------------+---------------------------------------+
|   LIBRARY    | VULNERABILITY ID | SEVERITY | INSTALLED VERSION | FIXED VERSION |                 TITLE                 |
+--------------+------------------+----------+-------------------+---------------+---------------------------------------+
| libcrypto1.1 | CVE-2021-23840   | HIGH     | 1.1.1i-r0         | 1.1.1j-r0     | openssl: integer                      |
|              |                  |          |                   |               | overflow in CipherUpdate              |
|              |                  |          |                   |               | -->avd.aquasec.com/nvd/cve-2021-23840 |
+              +------------------+----------+                   +               +---------------------------------------+
|              | CVE-2021-23841   | MEDIUM   |                   |               | openssl: NULL pointer dereference     |
|              |                  |          |                   |               | in X509_issuer_and_serial_hash()      |
[...]

PR #3375, opened 2021-12-20
Reducing Docker image by compiling to self-contained binary

SchulteMarkus:
# MkDocs (including Material theme) as binary for Docker image 

So I asked myself, if MkDocs (including Material theme) can be compiled to a self-contained binary. Actually, this is possible.

The backbone of doing so is https://www.pyinstaller.org/, which can create a standalone binary from a Python program.

In my opinion, using a binary is the better choice inside the delivered Docker image for MkDocs including material theme
1. Docker image size is reduced, as Python environment can be omitted. 
```bash
ms/mkdocs-material          latest             c27f7d13c656   21 minutes ago   48.9MB
squidfunk/mkdocs-material   latest             bd43b9c59ff4   24 hours ago     98.8MB
```
2. Quite naturally a Docker multi stage build can be used. Because of this, the Dockerfile does not have to cleanup caching etc. any more.

From what I saw while developing, performance of 1. MkDocs as Python 2. MkDocs as binary is similar, although 2. uses more RAM while serving.

## How to proceed

If you like this idea, I think next step would be to test and cleanup what I've done so far. As I've almost rewritten the Dockerfile, I'm pretty sure I missed or have broken something.
I am not a Python dev. Maybe/probably the used `PyInstaller` command can be improved, creating a smaller binary and/or a binary with better performance.
Feel free with any question, as well as committing to this PR itself. 

# Background

I discovered https://github.com/xvik/gradle-mkdocs-plugin, which make MkDocs available via Gradle. But it requires Python to be installed locally. Better choice might be using Docker instead (see https://github.com/xvik/gradle-mkdocs-plugin/issues/17) - but in my opinion, best choice would be having a MkDocs binary available somewhere.
If this PR is to be merged into mkdocs-material, I think the produced MkDocs binary could be made available as part of GitHub release assets, being available for direct download and execution; of course, this would not work in Windows yet.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,8 @@
 # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
 # IN THE SOFTWARE.
 
-FROM python:3.9.2-alpine3.13
+### 1. Docker build stage: Create MkDocs binary inluding material theme
+FROM python:3.10-alpine AS builder
 
 # Build-time flags
 ARG WITH_PLUGINS=true
@@ -38,44 +39,29 @@ COPY README.md README.md
 COPY requirements.txt requirements.txt
 COPY setup.py setup.py
 
-# Perform build and cleanup artifacts and caches
-RUN \
-  apk upgrade --update-cache -a \
-&& \
-  apk add --no-cache \
-    git \
-    git-fast-import \
-    openssh \
-&& \
-  apk add --no-cache --virtual .build \
-    gcc \
-    musl-dev \
-&& \
-  pip install --no-cache-dir . \
-&& \
-  if [ "${WITH_PLUGINS}" = "true" ]; then \
-    pip install --no-cache-dir \
-      "mkdocs-minify-plugin>=0.3" \
-      "mkdocs-redirects>=1.0"; \
-  fi \
-&& \
-  apk del .build \
-&& \
-  for theme in mkdocs readthedocs; do \
-    rm -rf ${PACKAGES}/mkdocs/themes/$theme; \
-    ln -s \
-      ${PACKAGES}/material \
-      ${PACKAGES}/mkdocs/themes/$theme; \
-  done \
-&& \
-  rm -rf /tmp/* /root/.cache \
-&& \
-  find ${PACKAGES} \
-    -type f \
-    -path "*/__pycache__/*" \
-    -exec rm -f {} \;
+# Perform build
+RUN apk add --no-cache binutils upx \
+ && apk add --no-cache --virtual .build gcc musl-dev
+RUN python -m pip install --upgrade pip
+RUN pip install --ignore-installed pyinstaller PyYAML \
+ && pip install --no-cache-dir . \
+ && if [ "${WITH_PLUGINS}" = "true" ]; then \
+      pip install --no-cache-dir \
+        "mkdocs-minify-plugin>=0.3" \
+        "mkdocs-redirects>=1.0"; \
+    fi
 
-# Set working directory
+# Create self-contained MkDocs binary inluding material theme
+RUN pyinstaller  --recursive-copy-metadata mkdocs --collect-all mkdocs \
+  --copy-metadata mkdocs-material --collect-all material --onefile /usr/local/bin/mkdocs
+
+### 2. Docker build stage: Serving MkDocs using binary from
+FROM alpine
+COPY --from=builder /tmp/dist/mkdocs /usr/local/bin/mkdocs
+
+RUN apk add --no-cache git git-fast-import openssh
+
+# Set build directory
 WORKDIR /docs
 
 # Expose MkDocs development server port


squidfunk:
Thanks for the PR!

> The backbone of doing so is https://www.pyinstaller.org/, which can create a standalone binary from a Python program.

I haven't used PyInstaller, but I'm aware of its existence. I would like to loop someone in with more Python background than I have – @facelessuser would you care to comment on PyInstaller and the overall approach in general?

> Docker image size is reduced, as Python environment can be omitted.

What's the impact of the omitted Python environment? Can the user still use/install additional MkDocs plugins? What if plugins have extra Python dependencies? The Docker image provides only a bare bone environment including two plugins, so I guess at least some users would want to install further plugins. Docker provides an easy way of extension – just use the official image as a base image and add your plugins, [as listed in the documentation](https://squidfunk.github.io/mkdocs-material/getting-started/#with-docker). Does that work with PyInstaller?

> If this PR is to be merged into mkdocs-material, I think the produced MkDocs binary could be made available as part of GitHub release assets, being available for direct download and execution; of course, this would not work in Windows yet.

We could do that, but I'm unsure if I want to maintain this, as I'm not a user of PyInstaller. However, if we don't decide to add this to the main repository and you want to maintain it, I'm happy to link to your repository/image from the documentation. We've done the same thing with the [`armv7`/`arm64` images](https://github.com/afritzler/mkdocs-material) which are built by @afritzler due to the significant increase in build time.

afritzler:
I haven't used PyInstaller myself but reducing the image size by 50% sounds pretty good to me. Also coming from a golang world I also like the one binary approach to packaging images. @SchulteMarkus are there any known compatibility issues?

Let me try out if my `arm` builds are working fine using this approach.

SchulteMarkus:
> What's the impact of the omitted Python environment? Can the user still use/install additional MkDocs plugins? What if plugins have extra Python dependencies? The Docker image provides only a bare bone environment including two plugins, so I guess at least some users would want to install further plugins. Docker provides an easy way of extension – just use the official image as a base image and add your plugins, [as listed in the documentation](https://squidfunk.github.io/mkdocs-material/getting-started/#with-docker). Does that work with PyInstaller?

This might be a killer argument of not using this approach, at least not as the main one.
The current approach of extending the base image will not work, because - as you said - no Python environment means no additional Python plugins. 
Can you imagine changing the approach of building custom mkdocs-material Docker images from extending to "Use existing Dockerfile, define additional plugins by [Docker build-time variables](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg) and (still) build yourself"? Of course, this would break existing custom builds.

> We could do that, but I'm unsure if I want to maintain this, as I'm not a user of PyInstaller. However, if we don't decide to add this to the main repository and you want to maintain it, I'm happy to link to your repository/image from the documentation. We've done the same thing with the [`armv7`/`arm64` images](https://github.com/afritzler/mkdocs-material) which are built by @afritzler due to the significant increase in build time.

I will provide a small single repository of doing so, if you decide not to change to current approach for building mkdocs-material Docker image.

squidfunk:
> Can you imagine changing the approach of building custom mkdocs-material Docker images from extending to "Use existing Dockerfile, define additional plugins by Docker build-time variables and (still) build yourself"?

I'm not keen on adding additional indirections to the build process, and switching to the approach you suggested would mean that a user would need to copy an existing Dockerfile and use it as a template instead of just using the extension mechanism. This would also mean that the user would not participate in updates to the Dockerfile, e.g. when fixing CVEs by updating the alpine base image as has happened in #2435 or adding dependencies in major version bumps.

I understand the use case for smaller images and always try to keep the size minimal while retaining functionality, but switching to PyInstaller would severely limit extension capabilities. For this reason, I'm not considering adding this to the main repository. If you want to provide this as part of a separate repository that is maintained by you, we can add a link to your repository to the documentation explaining that you provide sealed images that might lead to faster builds.

However, caching Docker layers during CI might also mitigate the issue at hand.

SchulteMarkus:
Totally makes sense. User experience in case of own plugins would definitely be reduced / kind of broken, and the improvement is possibly not worth it.

So I gonna make my thoughts how to publish this myself.
An idea here: Maybe create a GitHub organization "mkdocs-material", where https://github.com/squidfunk/mkdocs-material/ would itself be a repository of. And maybe https://github.com/afritzler/mkdocs-material would become an additional mkdocs-material repository "mkdocs-material-special-builds", where my stuff could be integrated into (what do you think @afritzler, of putting my stuff into your repo?). You could even extract https://github.com/squidfunk/mkdocs-material/tree/master/docs to an own repository inside that organization. Or, getting totally nuts, ever thought about putting https://github.com/squidfunk/mkdocs-material at https://github.com/mkdocs ?

squidfunk:
> An idea here: Maybe create a GitHub organization "mkdocs-material"

Definitely something worth considering, but I'm afraid that the ship is sailed, as there's no way to move the repository to the organization and also keep all current sponsorships for the project. If that's possible in the future, I will reconsider.

afritzler:
@SchulteMarkus if you like you can contribute the pyinstaller based setup into my repo. I can create a second publishing workflow for a e.g. `mkdocs-material-minimal` image. I think that would be the least invasive change here.

PR #3434, opened 2022-01-06
Fix "Changing the color" and "Changing the language" document javascript code

hafeoz:
Currently the [Color Setup](https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/) and [Language Setup](https://squidfunk.github.io/mkdocs-material/setup/changing-the-language/) page's button is not working correctly (at least on my Firefox 95.0.2 64-bit Arch Linux browser). When pressing buttons, incorrect text is changed, as illustrated in the screenshot:

image

(The "indigo" text in the code box should be changed, but the "palette" text is changed instead).

This is because the "primary" text is the 13th of the element in the code block and the javascript uses a hard-coded 7-element offset instead.

In this PR, I attempted to loop over the elements in code block to find the text that we need to modify in order to provide better compatibility over the hard-coded offset.

### DIFF
--- a/docs/setup/changing-the-colors.md
+++ b/docs/setup/changing-the-colors.md
@@ -44,7 +44,14 @@ Click on a tile to change the color scheme:
     button.addEventListener("click", function() {
       var attr = this.getAttribute("data-md-color-scheme")
       document.body.setAttribute("data-md-color-scheme", attr)
-      var name = document.querySelector("#__code_1 code span:nth-child(7)")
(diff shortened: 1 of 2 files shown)


squidfunk:
Ah yes, thanks for the PR. It broke when we switched on line anchors. Please just correct the offset then, because the code you provided also relies on offsets (`i + 3`) so isn't necessarily more stable but more complex 😉

hafeoz:
Got it. I'll start another PR then

Issue #2767, opened 2021-06-16
Please Add `plantuml-markdown` To The Docker Image


loganknecht:
Hello!

# Gratitude

Thanks for the wonderful docker image! It has been super helpful in working with my organization's documentation and `GitLab Pages`!

# Existing Discussion

https://github.com/squidfunk/mkdocs-material/discussions/2389

# Description/Use Cases

I'm working with `mkdocs` to generate consistent documentation across my organization's team.

The issue is that we use `GitLab` `Markdown` syntax and by default they support `PlantUML`

I've managed to figure out a way to configure our setup like so

```yaml
markdown_extensions:
    # Code block highlighting
    - pymdownx.highlight
    # Nesting code blocks
    - pymdownx.superfences
    # Insert code from other files
    - pymdownx.snippets
    # Inline code block highlighting
    - pymdownx.inlinehilite
    - plantuml_markdown:
          server: http://www.plantuml.com/plantuml
```

And then I can generate locally like this
```bash
docker run \
		--rm \
		-it \
		--workdir /app \
		--volume $(MAKEFILE_DIRECTORY):/app \
		--entrypoint /bin/sh \
		squidfunk/mkdocs-material \
		-c "pip install plantuml-markdown; mkdocs build --verbose --site-dir documentation_preview"
```

But it would be pretty cool to not obligate me to install pip like this, and instead have an image that allows me to use plantuml with minimal friction.

Because this isn't here I'm obligated to `pip install plantuml-markdown` every-time, whether it's local or in CI/CD.

# Request

Can we please add `plantuml-markdown` to the docker image?

squidfunk:
Thanks for suggesting! We have decided not to add further plugins to the Docker image unless they are as unopinionated as the minify and redirects plugin which are both bundled and provide significant value to the majority of users. We already have the [Mermaid.js integration](https://squidfunk.github.io/mkdocs-material/reference/diagrams/#superfences) which offers UML diagrams with the help of the already bundled SuperFences extension and I don't see why _everybody_ should need the `plantuml-markdown` extension.

That said, I'm not considering adding this plugin. Docker image extension is super easy after all, as described in the documentation, see the collapsed admonition in [the installation guide](https://squidfunk.github.io/mkdocs-material/getting-started/#with-docker). You could build a Docker image based on the one offered by this project and add your plugins with a few lines. You could also push your image to DockerHub to make it available to others.

Issue #2349, opened 2021-02-26
[FEAT]: Add `awesome-pages` into the Docker image


polarathene:
- [x] I've read the [contribution guidelines][1] and agree with them

__I want to suggest an idea and checked that ...__

- [x] ... to my best knowledge, my idea wouldn't break something for other users
- [x] ... the documentation does not mention anything about my idea
- [x] ... there are no open or closed issues that are related to my idea

## Description

Would it be possible to consider adding back the [`awesome-pages` plugin](https://github.com/lukasgeiter/mkdocs-awesome-pages-plugin)?

This was deprecated back in [5.5 and removed in 6.0](https://squidfunk.github.io/mkdocs-material/deprecations/#bundled-plugins) with the reasoning of size concerns. While this is valid with `mkdocs-git-revision-date-localized-plugin` (31MB layer), `mkdocs-awesome-pages-plugin` only adds 323KB as layer weight.

### Justification

If size is a concern, it looks like there is plenty of opportunity to reduce the Docker image further.
- For example the `lunr` package includes a 4 `rust_stemmer.cpython.*.so` files, 3 appear to be diffentiated by `35m`, `36m` and `37m`... which I assume might be relevant to Python version support? These are files built during install not source files, they have a weight of 2.7MB each, two can be discarded and the container would work as intended still. Furthermore the 4th file is a binary for `darwin` (macOS), 345KB right there that could be swapped out for `awesome-page`.
- `nltk` which seems to be a supporting package for `lunr` I think is a 9MB package, there's 1.2MB `test` directory that I doubt is required for this project?
- `python3.8` in total is 52MB, with `site-packages` at 48MB, I don't know Python well enough to know if anything else could be cleaned up from there.
- `git` and related dependencies also eat up around 20MB in size (this layer is 77MB). Technically a slimmer image could be used if a minimal/slim base image was useful for most users (`serve` and possibly `build`?), those that would use the image to deploy could use a `deploy` variant, but I imagine most would be fine deploying the build output via other means.
- The earlier layers from the python image you extend is about 40MB in addition to all this. Total image size is 126MB.

These insights were provided via the CLI tool [`dive`](https://github.com/wagoodman/dive). The alternative git revision date plugin `mkdocs-git-revision-date-plugin` weighs 1.3MB btw if you were interested.

### Use Cases

[...]

PR #3457, opened 2022-01-14
Include revision-date(*) plugins in Dockerfile

weeblr:
These are supported plugins, mentioned in the documentation, it'd be rather convenient to have them included in the Docker image without having to build your own derived image.

I am aware of #1455 but altough you seemed keen on accepting these 2 plugins. However the discussion derived into another one on improving the Dockerfile structure, which was merged, but inclusion of these 2 plugins was "lost in translation".

This PR is about bringing them back.


squidfunk:
Thanks for the PR. However, I'm not interested in doing so. Please see:

- #2349
- #2767

PR #3460, opened 2022-01-16
Add a popular combination of dark theme toggle

wyfo:
This combination is presented in material documentation:
https://material.io/design/color/dark-theme.html#behavior


squidfunk:
Thanks, great idea!

PR #3473, opened 2022-01-17
Remove ToC setup instructions

oliversalzburg:
The ToC appears to be enabled by default. Also, the instructions are for enabling permalinks, which are a distinct feature already listed further below.


squidfunk:
Permalinks are not enabled by default and the block you removed is the recommended minimal configuration (including permalinks). Since permalinks are not enabled by default, this line is in fact necessary, and thus should stay in the docs.

PR #3480, opened 2022-01-19
Added Google Tag Manager support

gergoradeczki:
Imported the JavaScript code from https://tagmanager.google.com/ with the modification to use the property id from mkdocs.yml.


squidfunk:
Thanks for the PR. You made the changes in the `material` folder, which is overwritten during build, so this won't work. Please make your changes in the `src` folder and then build the theme. This is explained in the [theme development guide](https://squidfunk.github.io/mkdocs-material/customization/#theme-development), and mentioned in the [contribution guidelines](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md#submitting-a-pull-request-pr) 😉

squidfunk (changes requested):
See above

squidfunk (changes requested):
Site search and instant loading integration seem to be missing. Please add.

gergoradeczki:
I might be wrong but, as I understood the differences between Google Analytics and Google Tag Manager, GTM doesn't require you to explicitly add these events in code. Instead, you would have to add it through their website.

squidfunk:
Oh, okay. I'm not familiar with tag manager. So to better understand whether this PR is a good fit for the project, how would you implement site search tracking via GTM? is that possible? How does a GTM setup look like? I'd be curious to learn some more before I can decide whether we can merge this or not.

gergoradeczki:
This is what I ended up doing in order to enable tracking of users:

image

In the Tags tab, I created **Google Analytics GA4 Configuration** Tag and added a Page View trigger called All Pages which from its name, fires on all pages when loaded. Then created a **History Changes** trigger in the Triggers and added it to the triggers list for the **Google Analytics GA4 Configuration**.

For the search field, I don't think I have the best solution. My solution would be to track the URL changes, then extract the ?h=string query using a variable.

image

squidfunk:
Thanks for the explanation. I think we can add support for GTM, but this PR still needs some work before it can be merged:

- Can we improve the docs? They only explain how to add GTM, but not what to make of it, e.g. how to achieve the same stuff as with the other integrations. We don't have to add guides explaining everything in detail, but some guidance for users in the form of recommended readings or some instructions would be beneficial. If we don't have this in the docs, users will open issues asking for them, which is more work for me in the future. Thus, please think of how we can improve the documentation so that users know what to make of GTM.

- Please format the source code so that it matches the style of the other integrations. I know that this is generated from a snippet, but I'd rather have readable, maintainable code.

docs/setup/setting-up-site-analytics.md
```diff
@@ -34,6 +34,15 @@ following lines to `mkdocs.yml`:
         property: G-XXXXXXXXXX
     ```
 
+=== ":material-google-analytics: Google Tag Manager"
```
squidfunk, line 37: Please add documentation, see my last comment.

src/partials/integrations/analytics/google.html
```diff
@@ -90,6 +90,27 @@
     })
   </script>
   <script async src="https://www.google-analytics.com/analytics.js"></script>
-{% endif %}
 
+<!-- Google Tag Manager (GTM-XXXXXXX) -->
+{% elif property.startswith("GTM-") %}
+  <!-- Google Tag Manager -->
+  <script>
+    (function(w, d, s, l, i) {
```
squidfunk, line 98: Please format the sources properly. Those are no good variables names.

squidfunk (changes requested):

squidfunk:
@gergoradeczki could you please make the requested changes?

gergoradeczki:
I'm not confident enough with the search query tracking solution I found and because of this I don't think the addition of GTM is a good idea.

squidfunk:
Okay, never mind, thanks for your time!

pstavirs:
@squidfunk Any chance/plans to support Google Tag Manager? Even if search query tracking is not possible with GTM, it would be good to have it as an option for those who want to use GTM instead of GA4.

minhluan259:
@pstavirs any update on this? Thanks.

pstavirs:
See https://github.com/squidfunk/mkdocs-material/discussions/5607

squidfunk:
@minhluan259 We currently have no plans to add it to core, but [it can be achieved with customization in 5 minutes](https://github.com/squidfunk/mkdocs-material/discussions/5607#discussioncomment-7323367).

PR #3542, opened 2022-02-06
Add custom language code to the config schema

notpushkin:
This resolves the `Value is not accepted` warning when [validating mkdocs.yml](https://twitter.com/squidfunk/status/1487746003692400642).

Technically, any `language` value would be valid here as long as user provides the relevant `overrides/partials/languages/{{ language }}.html` file, but `"custom"` is the one [suggested in the docs](https://squidfunk.github.io/mkdocs-material/setup/changing-the-language/#custom-translations) so it is what most users would probably stick with.

### DIFF
--- a/docs/schema/theme.json
+++ b/docs/schema/theme.json
@@ -46,6 +46,12 @@
       "markdownDescription": "https://squidfunk.github.io/mkdocs-material/setup/changing-the-language/",
       "oneOf": [
         {
+          "title": "Site language: Custom",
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. Yes, we really have to strike a balance between being strict enough to perform validation and at the same time allowing for theme extension. I guess `custom` should work. I'll add a link to the documentation for that value in the next commit.

PR #3564, opened 2022-02-08
Update fontawesome URL

Andre601:
Fontawesome seems to have shipped their next version which comes with a new website.

Ignoring the god awful way to find anything (And the fact that clicking the button to see all icons is extremely unreliable) does it seem like the URL has slightly changed as the current one is no longer really usable.

This PR updates it from https://fontawesome.com/icons?d=gallery&m=free to https://fontawesome.com/search?m=free which works.


squidfunk:
Thanks!

PR #3592, opened 2022-02-17
Fix outdated fontawesome icon name

Andre601:
So..... apparently `laugh-wink` has changed to `face-laugh-wink` in FA v6... nice.

On another small note, I consider also taking the time to add an info-box regarding Octicons and that their names require the pixel size to be appended too... (Also, are the 12px variants included too? Not really important for me, but just curious.)


squidfunk:
Thanks! I don't think we need an infobox, as the [icon search](https://squidfunk.github.io/mkdocs-material/reference/icons-emojis/) will list all octicons variants. Try searching for `octicons` using the icon search. Regarding the `12px` variants – we source the icons from the official `@primer/octicons` package – if they are part of this package, they will also be included with Material for MkDocs.

PR #3654, opened 2022-03-02
add GitHub URL for PyPi

andriyor:
Warehouse now uses the project_urls provided to display links in the sidebar on [this screen](https://pypi.org/project/requests/), as well as including them in API responses to help the automation tool find the source code for Requests.

Docs: [packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,6 +41,9 @@ setup(
     name = "mkdocs-material",
     version = package["version"],
     url = package["homepage"],
+    project_urls = {
+        "Source": "https://github.com/squidfunk/mkdocs-material",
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! Could we also add https://squidfunk.github.io/mkdocs-material/ as Homepage and Documentation?

andriyor:
I think it's redundantly and produces duplicates

PR #3666, opened 2022-03-04
Add mdpo plugin JSON schema

mondeja:
Not sure if the plugins are alphabetically placed there.


squidfunk:
Thanks for providing this! Awesome to see more plugins being added. Yes, the plugins should be ordered alphabetically.

PR #3688, opened 2022-03-09
Use comma as thousand separator in documentation

iBug:
According to [Wikipedia][1], all major [English-speaking countries][2] use dot as decimal point, instead of comma. This makes it desirable to follow the convention in an English context.

This PR changes two occurrences of dot used as digit *grouping separator* (a.k.a. thousand separator) into comma, so as not to be confused with decimal point.

  [1]: https://en.wikipedia.org/wiki/Decimal_separator#Current_standards
  [2]: https://en.wikipedia.org/wiki/List_of_countries_and_territories_where_English_is_an_official_language


squidfunk:
Thanks! In German it's the other way round, which is why I messed this up.

PR #3762, opened 2022-03-23
Tags Plugin: raise exception when tags.md file doesn't exist

jdoiro3:
I think it would be beneficial to raise an exception when the file for the tags plugin doesn't exist. This would prevent *None* from being added to the `Files` collection in mkdocs, since `get_file_from_path` returns *None* if the `file` isn't in the dict object.

*relevant MkDocs code*
https://github.com/mkdocs/mkdocs/blob/f725e225adf9b6cfab45042e44637f137f56d1bd/mkdocs/structure/files.py#L26-L32


squidfunk:
Superseded by dfba03f15.

PR #3770, opened 2022-03-27
Mention config iheritance for social cards

PEZ:
Hello, I think it is good to know of the config inheritance option when setting up social cards. So I added a few words and links about it.

### DIFF
--- a/docs/setup/setting-up-social-cards.md
+++ b/docs/setup/setting-up-social-cards.md
@@ -59,7 +59,8 @@ The following configuration options are available:
 
 :   :octicons-milestone-24: Default: `true` – This option specifies whether
     to generate social card images. If you want to switch the plugin off, e.g.
-    for local builds, you can use an [environment variable]:
+    for local builds, you can either use MkDocs [configuration inheritance],
+    as described in [built-in-plugins], or use an [environment variable]:
 
     ``` yaml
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks! However, I'd find it to be confusing in this section. The Insiders installation guide includes a [section on configuration inheritance](https://squidfunk.github.io/mkdocs-material/insiders/getting-started/#built-in-plugins), which IMHO is a better place, as we now have several Insiders-only plugins.

PEZ:
Yes, I do link to that section. Basically I just want to help people avoid the mental trap I got caught in trying to solve my use case with that ENV thing.

I now made another attempt that possibly is less confusing. You think that is better? Or is it better to not mention this option at all, maybe?

squidfunk:
Let me think about this and come back to you. I'm in the process of doing a minor docs restructuring, so maybe I can come up with a better place than the one you suggested, but let's see. In general, I try to balance reading flow with the locality of information, which is always hard. It's not possible to have all information local.

PEZ:
Cool. I agree it is a trade-off between flow and offering enough help for enough user types.... I'm taking the opportunity to learn from how you reason around issues like this.

squidfunk:
Superseded by 27efd1dd6.

PR #3783, opened 2022-03-30
Use key value syntax for Insider plugins example

cicdw:
I noticed that the official Insiders documentation used hyphenated list syntax when enumerating plugins, but this is not compatible with MkDocs configuration inheritance requirements when merging config files; I assume most people won't click through and read carefully, so this PR updates the documentation to use syntax that will "just work" along with a note.


squidfunk:
Thanks for the PR. I don't think it's a good idea to change the syntax here, because it would be the only part of the documentation using that syntax, which might confuse users. I understand your requirement for merging configurations, though, as you probably don't want to duplicate the plugin setup. Moving forward, I would suggest that we put a note into an admonition explaining the alternate syntax below the configuration inheritance example.

cicdw:
@squidfunk updated 👍

docs/insiders/getting-started.md
```diff
@@ -185,4 +185,17 @@ pipeline), you can build your documentation project with the following lines:
 mkdocs build --config-file mkdocs.insiders.yml
 ```
 
+!!! warning "Alternative syntax"
+    
+    If you plan to split `plugins` or `markdown_extensions` between the two files, you must use
+    alternative key / value syntax for these sections.  The above insiders file would then look like:
+    ``` yaml
+    INHERIT: mkdocs.yml
```
squidfunk, line 193: Could we maybe provide two code blocks each for each `mkdocs.yml` file (base + Insiders overrides) which just adds the `social` plugin? The other plugins are non-Insiders anyway. This would be a more self-contained example for the user to understand how to do overrides while keeping the plugin configuration of the base configuration. I say we could just add the `social` plugin as an example. Otherwise, users might start to mix syntaxes, or is that possible?

squidfunk (changes requested):
Thanks! I think we should make the example a little more explicit, see my comments.

docs/insiders/getting-started.md
squidfunk: So essentially something like:

`mkdocs.yml`

``` yaml
plugins:
  search: {}
  tags: {}
```

`mkdocs.insiders.yml`

``` yaml
INHERIT: mkdocs.yml
plugins:
  social {}
```

Please use the content tabs pattern for this, as in [other parts of the documentation](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#custom-admonitions).

cicdw: Thanks for your patience with my slow updates - I just pushed the requested changes!

PR #3811, opened 2022-04-11
recommand image caption plugin-> img2fig

Jackiexiao:


Jackiexiao:
this plugin has a [bug](https://github.com/stuebersystems/mkdocs-img2fig-plugin/issues/4)

PR #3824, opened 2022-04-20
Update Mermaid.js to version 9.0.0

vectorsli:
This update brings `gitGraph` to stable. Since the support for `gitGraph` wasn't advertised and this is the reason for the major version bump, it should be safe to upgrade.

**Note**: I used search + replace to change `8.13.3` to `9.0.0` for the mermaid dependency. The development environment I created locally didn't work as expected and I didn't want to debug.


squidfunk:
Thanks for the PR. However, without testing and ensuring that all graphs still work, I'm unable to merge. I might look into this in the future, but if you want your PR merged, we should make sure that this PR doesn't break anything. In essence, please provide screenshots of all graphs on the diagrams page after upgrading to 9.0.0.

squidfunk (changes requested):
As noted.

vectorsli:
I added an example to the documentation:
image

The other diagrams look fine, too (from bottom up):
image
image
image
image
image


## Development setup

These images were created using the following dev workflow:
``` sh
npm start
mkdocs serve --watch-theme
```

Building the theme using `npm run build`/`npm run build:all` didn't work for me. The `iconsearch_index.json` wasn't generated and for some reason the html templates were referencing the non-minified versions of the assets (`bundle.js` etc.). I'm not sure what's wrong with my setup, maybe it's because I'm on Windows?
Anyway, I don't think it's useful to commit my generated files, so I'll leave the PR as it is. I hope that's okay.

docs/reference/diagrams.md
```diff
@@ -260,3 +260,43 @@ erDiagram
 </div>
 
   [entity-relationship diagram]: https://mermaid-js.github.io/mermaid/#/entityRelationshipDiagram
+
+### Using Gitgraphs
```
squidfunk, line 264: Please remove this section – we're not officially supporting this type of diagram.

squidfunk (changes requested):
Thanks for making the changes! However, we don't want to advertise support for Git Graphs, because they have several problems when rendered, particularly sizing issues that don't work well on mobile. This is explained [here](https://facelessuser.github.io/pymdown-extensions/extras/mermaid/#practical-diagrams). Thus, the section should be removed from the documentation, or users will start to create issues because Git Graphs do not work as advertised.

docs/reference/diagrams.md
squidfunk: I've added a note for the other diagram types in 314bab53c.

squidfunk:
Superseded by 24025f42. I went ahead and updated to Mermaid.js 9.0.1, the latest version. Thanks again for testing the diagrams! I've discovered two more problems with arrows in dark mode that I fixed in the same commit.

vectorsli:
Thanks for bringing this to the latest release!

Issue #3840, opened 2022-04-24
pymdownx.highlight extension causes crash at startup


jstafford5380:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the problem doesn't occur with the `mkdocs` or `readthedocs` themes
- [X] ... the problem persists when all overrides are removed, i.e. `custom_dir`, `extra_javascript` and `extra_css`
- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

I'm trying to get code highlighting to work but when including the documented option to enable, the site crashes

```yaml
markdown_extensions:
  - pymdownx.highlight:
      anchor_linenums: true
```

### Expected behaviour

The site should start up and code blocs should be highlighted

### Actual behaviour

When serving the site, the crashes

```sh
mkdocs serve
INFO     -  Building documentation...
INFO     -  Cleaning site directory
ERROR    -  Error reading page 'index.md': format() argument must be a formatter instance, not a class
Traceback (most recent call last):
  File "C:\Python310\lib\site-packages\pygments\__init__.py", line 61, in format
    formatter.format(tokens, realoutfile)
  File "C:\Python310\lib\site-packages\pygments\formatter.py", line 94, in format
    return self.format_unencoded(tokensource, outfile)
  File "C:\Python310\lib\site-packages\pygments\formatters\html.py", line 981, in format_unencoded
    source = self.wrap(source)
TypeError: BlockHtmlFormatter.wrap() missing 1 required positional argument: 'outfile'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Python310\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Python310\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\Python310\Scripts\mkdocs.exe\__main__.py", line 7, in <module>
  File "C:\Python310\lib\site-packages\click\core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "C:\Python310\lib\site-packages\click\core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "C:\Python310\lib\site-packages\click\core.py", line 1657, in invoke
[...]

PR #3862, opened 2022-04-30
Fix issues related to latest Pygments and Pymdown Extensions

facelessuser:
Fixes remaining issues related to #3840

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,5 +23,5 @@ jinja2>=2.11.1
 markdown>=3.2
 mkdocs>=1.3.0
 mkdocs-material-extensions>=1.0.3
-pygments>=2.10,<2.12
-pymdown-extensions>=9.0,<9.4
+pygments>=2.12
+pymdown-extensions>=9.4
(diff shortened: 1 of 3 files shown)


facelessuser:
The environment would need to be rebuilt, I didn't want to commit all of that noise.

squidfunk:
Thanks!

squidfunk:
Unfortunately, there's still some work to do. Tables now have rounded borders in tabs because the hierarchy is entirely different. I need some more time to polish this and wrap it up, but it's tedious work.

squidfunk:
I think I got all cases. There was some regression with rounded borders for full-width code blocks on mobile, and with code blocks nested under tabs. Fixes in bdb5ba5a3. I need to merge this into Insiders tomorrow, then I can issue a release.

Unfortunately, some selectors need to be overly specific.

facelessuser:
Ah, yeah, I missed that apparently.

squidfunk:
Let's hope the syntax doesn't change again. Code blocks and content tabs and their combinations are quite fragile 😅

facelessuser:
Yeah, I hope they don't change again too. It caught me off gaurd.

squidfunk:
Merge done. Also, https://github.com/squidfunk/mkdocs-material/commit/1eeb8b7c0ce30a2f3e1870e606be730d1fa3596a removes some more unnecessary stuff. I'll issue a release later or tomorrow.

PR #3877, opened 2022-05-02
Update python version requirement to 3.7

### LINKED ISSUE #3866: mkdocs-material@8.2.12 requires python >= 3.7 instead of python 3.6
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the problem doesn't occur with the `mkdocs` or `readthedocs` themes
- [ ] ... the problem persists when all overrides are removed, i.e. `custom_dir`, `extra_javascript` and `extra_css`

### Description

Installing `mkdocs-material@8.2.12` on an environment with python 3.6 would fail because one of the dependencies `pymdown-extensions` is now required to be `>=9.4`. However, since `pymdown-extensions@9.2`, it has dropped support for python 3.6 and now requires python 3.7 or above. 

Here's the error message:
```
  Could not find a version that satisfies the requirement pymdown-extensions>=9.4 (from mkdocs-material) (from versions: 1.0.0, 1.0.1, 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 1.6.1, 1.7, 1.8, 2.0, 3.0, 3.1, 3.2, 3.2.1, 3.3, 3.4, 3.5, 4.0, 4.1, 4.2, 4.3, 4.4, 4.5, 4.5.1, 4.6, 4.7, 4.8, 4.9, 4.9.1, 4.9.2, 4.10, 4.10.1, 4.10.2, 4.11, 4.12, 5.0, 6.0, 6.1, 6.2, 6.2.1, 6.3, 7.0b1, 7.0b2, 7.0rc1, 7.0rc2, 7.0, 7.1, 8.0, 8.0.1, 8.1, 8.1.1, 8.2, 9.0.dev0, 9.0, 9.1)
No matching distribution found for pymdown-extensions>=9.4 (from mkdocs-material)
```

### Expected behaviour

Install mkdocs-material on python 3.6 environment would succeed (i.e. it installs mkdocs-material@8.2.11, which is the last version that support python 3.6)

### Actual behaviour

It fails with the error message mentioned in description

### Steps to reproduce

On an environment with python 3.6, run
```
pip3 install mkdocs-material
```

### Package versions

- Python: `python3 --version`: 3.6.9
- MkDocs: `mkdocs --version`: unrelated
- Material: `pip show mkdocs-material | grep -E ^Version`: unrelated

### Configuration

```yaml
Unrelated
```

### System information

- Operating system: Ubuntu 18.04
- Browser: Unrelated

DennisPan:
With `pymdown-extensions` required version recently changed to `9.4`, `mkdocs-material` effectively requires at least python 3.7 now

This fixes #3866

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,7 @@ setup(
     packages = find_packages(exclude = ["src", "src.*"]),
     include_package_data = True,
     install_requires = install_requires,
-    python_requires='>=3.6',
+    python_requires='>=3.7',
     entry_points = {
         "mkdocs.themes": [
             "material = material"


squidfunk:
Thanks! LGTM!

PR #3883, opened 2022-05-03
Doc: Description tag now incorporates per page meta when present 

AjaniBilby:
Previously OpenGraph and Twitter meta tags would use the site-wide description in their content, however, this change means they use the page specific description when present on the page.


AjaniBilby:
Wait....
I think there is actually a bug created by this which was actually in the original as well...

AjaniBilby:
I'm getting a weird bug where if you use the search to go to a page, the title of any page you then visit from that point on will override the title in your header bar... Appears to be unrelated and an issue with my environment.

squidfunk:
This will make the example noisier and more complicated. I'd really like to keep it as short and concise as possible. However, I invite you to share your customization as a discussion, so users can find it 😊

PR #3910, opened 2022-05-11
Update README.md

linmq:


squidfunk:
Thanks for the PR! However, this space is reserved for very popular projects (checkout out how many stars/forks FastAPI has). The reason for this is that popular projects adopting Material for MkDocs is a good way to show that this project is mature and stable. For this reason, I'm currently not considering merging this PR.

linmq:
OK, thanks.

PR #3930, opened 2022-05-17
fixed the water formula typo

Minhao-Zhang:
Found a small typo in formatting.md about the super and subscript. The formula for water should be H2O not H20. 

This is the first time ever I have done a PR, I don't know I am doing this correctly or not.


squidfunk:
Ha, good catch! LGTM. Thanks!

Issue #3931, opened 2022-05-18
Tabs inside header support


yinkh:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I want to suggest an idea and checked that ...

- [X] ... to my best knowledge, my idea wouldn't break something for other users
- [X] ... the documentation does not mention anything about my idea
- [X] ... there are no open or closed issues that are related to my idea

### Description

Current tab navigation add a new nav below header,shall we support put nav inside the header?
when site_name or page name(navigation.tabs.sticky is config)  is overlength, we can truncatewords.

### Use Cases

Maximize the size of content.
Put the navigation bar inside header is one more common way.

### Screenshots / Mockups

![image](https://user-images.githubusercontent.com/8170059/168949754-230f1ca8-005b-435f-bdde-acca134ed53e.png)

yinkh:
Another question: if you think this feature is needed and i want to develop this feature, the source code of  `material/assets/javascripts/bundle.efad133d.min.js` is need. Any way to fetch this source code?

squidfunk:
Thanks for suggesting. I really don't want to maintain two navigation tabs implementations 😉 I guess you can imagine that a project of this size with so many features has a lot of issues and feature requests popping up all the time, and adding a slightly different take on something that already exists is something I'm not very interested in. After all, it's me who has to maintain it after the original implementer decided to move on. For this reason, I'm not considering this. However:

> Another question: if you think this feature is needed and i want to develop this feature, the source code of material/assets/javascripts/bundle.efad133d.min.js is need. Any way to fetch this source code?

All templates and assets are built from the `src` folder. If you fork the theme and set up the [theme development pipeline](https://squidfunk.github.io/mkdocs-material/customization/#theme-development), you can make the changes yourself to adjust Material for MkDocs to fit your taste!

squidfunk:
On a side note, recommended read:

- https://github.com/squidfunk/mkdocs-material/discussions/2399

yinkh:
thanks for your reply,I will customize this feature myself. Thanks for this powerful document tools again!

yinkh:
I have achieve this feature,look like this:
https://user-images.githubusercontent.com/8170059/169009329-ac5acfeb-6c41-42d7-a8c6-3fd6c7c2a7e0.mp4

anyone has same requirement can ask me to post source code

[...]

PR #3936, opened 2022-05-19
Feature tabs in header

yinkh:
This is a PR for issue [Tabs inside header support](https://github.com/squidfunk/mkdocs-material/issues/3931)
It's ok if PR is rejtect,i just want someone who need  this feature can get source code quickly.

```
feature:
    - navigation.tabs
    - navigation.tabs.header
```
works like this:

https://user-images.githubusercontent.com/8170059/169200508-44be7cf4-9afe-4b10-b17d-3589febc7a18.mp4

```
feature:
    - navigation.tabs
    - navigation.tabs.header
    - navigation.tabs.sticky
```
works like this:

https://user-images.githubusercontent.com/8170059/169200591-772601d4-dc5e-4135-9a68-ef812d477288.mp4


squidfunk:
Thanks for the PR! However, as I already stated in https://github.com/squidfunk/mkdocs-material/issues/3931#issuecomment-1129576014, it's nothing I can commit to maintaining, so I can only reject the PR. If somebody wants to add support, this PR is a very good start though, so thanks for providing it!

PR #3939, opened 2022-05-19
Fix incorrect info about mathjax blocks

Andre601:
The information about mathjax blocks is (at least from personal testing using the provided javascript) incorrect.

The format actually has to be
```mathjax
$$...$$
```

and not
```mathjax
$$
...
$$
```

### DIFF
--- a/docs/reference/mathjax.md
+++ b/docs/reference/mathjax.md
@@ -85,20 +85,16 @@ See additional configuration options:
 
 ### Using block syntax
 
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. However, my testing shows that the examples work correctly on separate lines. If you inspect the source of the [MathJax reference document](https://raw.githubusercontent.com/squidfunk/mkdocs-material/master/docs/reference/mathjax.md), you'll see that we use this exact syntax. Closing, as this doesn't seem to be necessary.

PR #3941, opened 2022-05-20
Mention alternative to ignore files

Andre601:
MkDocs ignores files and folders starting with a dot, which is useful for snippets and is why I think it should be mentioned for it.


squidfunk:
Thanks for the PR! While this is supported by MkDocs, I don't think that it's a good idea to recommend this approach to users. All files that MkDocs should not consider as part of the hosted site (like includes) are better located outside of the `docs` folder. Why? Because MkDocs might change the behavior to ignore dotted folders at any time, so to be future-proof, putting it outside the `docs` folder is definitely a better idea.

PR #4010, opened 2022-06-10
fix(versioning): set correct link to supported release

kilchenmann:
There's a wrong link in the setting-up-versioning page. It should be fixed with this patch.


squidfunk:
Thanks! LGTM.

PR #4047, opened 2022-06-22
It is an insider feature

huangyinhaow:


squidfunk:
Thanks for the PR! However, diagrams became generally available as of [8.2.0](https://github.com/squidfunk/mkdocs-material/releases/tag/8.2.0) thanks to all of our awesome sponsors!

PR #4048, opened 2022-06-22
It is not an insider feature

huangyinhaow:


squidfunk:
I'm not sure what you're trying to achieve (see #4047), but Chinese search support is an Insider feature.

huangyinhaow:
But why can I still use Chinese to search? I am not an insder.

image

PR #4126, opened 2022-07-14
Move schema.json of git-revision-date-localized.json

timvink:
See https://github.com/timvink/mkdocs-git-revision-date-localized-plugin/issues/93 for context


squidfunk:
Awesome, thanks!

PR #4127, opened 2022-07-14
rename tags plugin

lbenedix:
to avoid collisions with pre-existing plugins like https://github.com/jldiaz/mkdocs-plugin-tags

### DIFF
--- a/docs/insiders/getting-started.md
+++ b/docs/insiders/getting-started.md
@@ -169,7 +169,7 @@ be mitigated by using [configuration inheritance]:
     plugins:
       - search
       - social
-      - tags
+      - material-tags
     ```
 
 === ":octicons-file-code-16: mkdocs.yml"
--- a/docs/setup/setting-up-tags.md
+++ b/docs/setup/setting-up-tags.md
@@ -25,7 +25,7 @@ the following lines to `mkdocs.yml`:
 
 ``` yaml
 plugins:
-  - tags
+  - material-tags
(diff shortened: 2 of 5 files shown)


squidfunk:
Thanks for your suggestion. We won't do that. I understand that there is a collision, however, there's no central registry for MkDocs plugins which means there will always be collisions. Since our tags implementation exceeds the functionality of the third-party plugin, it shouldn't be necessary to use the third-party plugin with Material for MkDocs.

There might be a way to fully qualify a plugin name, something like `material.plugins.tags`, as this works for `markdown.extensions.abbr` and `abbr`, but I couldn't find out how it works. @facelessuser, do you have any idea?

facelessuser:
@squidfunk, I'd have to look into this. It honestly depends on how the library processes the entry point request and how the plugin specifies the entry point. I know Python Markdown, while it supports entry points, does not explicitly require them and still accepts them as a string, will parse it, and then import it from the specified location.

Off the top of my head, I'd have to explore locally what could be done. I want to say it is possible? Though I'd have to actually try to attempt to implement it as I don't recall the requirements, nor I have personally implemented dealt with entry pints frequently. It's usually something I occasionally do. I may look into it sometime soon...if I remember 🙃 .

squidfunk:
No worries, I just tend to loop you in whenever I feel that my Python-fu is subpar, which is basically always, so no hurry 😉

PR #4130, opened 2022-07-16
Add mkdocs-glightbox as lightbox recommended plugin

blueswen:
Add [mkdocs-glightbox](https://github.com/blueswen/mkdocs-glightbox) as lightbox recommended plugin mentioned in https://github.com/squidfunk/mkdocs-material/issues/4096.

The header and sidebar vanishing bug has been fixed in [mkdocs-glightbox v0.1.5](https://github.com/blueswen/mkdocs-glightbox/releases/tag/v0.1.5). Also fix a content shrinking bug and use [document$.subscribe](https://github.com/blueswen/mkdocs-glightbox/blob/v0.1.5/mkdocs_glightbox/plugin.py#L58) to make sure the compatibility with Instant loading feature. All plugin options have been tested the compatibility with Material for MkDocs, and all options looks working fine without breaking Material for MkDocs.

Feel free to edit these contents. Not sure these information for user is enough or not.

PR #4133, opened 2022-07-17
Add mkdocs-glightbox plugin schema.json

blueswen:
Add mkdocs-glightbox plugin schema.json mentioned in https://github.com/squidfunk/mkdocs-material/pull/4130 .


squidfunk:
Awesome!

PR #4172, opened 2022-07-27
Align "edit this page" icon with the one from mkdocs-material

ojacques:
Make the "edit this page" icon the same as the one from the mkdocs-material documentation site.

As one can see from [the commit](https://github.com/squidfunk/mkdocs-material/commit/d926bb4c2791300548634bbd5f7d2ca23d997367) which adds "edit this page" and "view source of this page", the icon to edit the page has changed for the mkdocs-material documentation site. But not for any site made with mkdocs-material.

This fixes it.

Before: 

image

After:

image

### DIFF
--- a/material/partials/content.html
+++ b/material/partials/content.html
@@ -3,7 +3,7 @@
 -#}
 {% if page.edit_url %}
   <a href="{{ page.edit_url }}" title="{{ lang.t('edit.link.title') }}" class="md-content__button md-icon">
-    {% include ".icons/material/pencil.svg" %}
+    {% include ".icons/material/file-edit-outline.svg" %}
   </a>
 {% endif %}
 {% if "tags" in config.plugins %}


squidfunk:
Thanks for the PR. However, a better idea is to make the edit button icon configurable. You can create a PR or an issue for that. I changed the appearance because of the Markdown preview source button. Icons are always to-taste.

It's quite simple to implement, e.g. see how we did it for the logo:

https://github.com/squidfunk/mkdocs-material/blob/47801173ed1474d168e3e6aba6d6bd5af8bb4df1/src/partials/logo.html#L27-L28

We could add a new option called `config.theme.icon.edit` for this with a default of `material/pencil`. A PR would also include the necessary documentation [on how to change the icon (example)](https://squidfunk.github.io/mkdocs-material/setup/adding-a-git-repository/#repository-icon), probably as a sub-section of the [edit button](https://squidfunk.github.io/mkdocs-material/setup/adding-a-git-repository/#edit-button).

ojacques:
Thanks for moving forward on this in the last insiders build!

PR #3034, opened 2021-09-17
Add multi arch docker build

afritzler:
This PR adds support for building and publishing linux/arm64 and arm/v7 Docker images.

The output works fine on my Macbook M1 and can be found here: https://github.com/afritzler/mkdocs-material/pkgs/container/mkdocs-material

Additionally the arm/v7 images can now be run on a Raspberry Pi (tested on a Raspberry Pi 2 Model B Rev 1.1).

Fixes #2945

PR #4240, opened 2022-08-11
feat: multi platform docker image

caarlos0:
The original image is really slow on m1 macs, for example.

This will build for arm64 as well, which should improve things.

You can test it: https://hub.docker.com/r/caarlos0/material-mkdocs/tags

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -83,7 +83,7 @@ jobs:
         uses: docker/build-push-action@v3
         with:
           context: .
-          # platforms: linux/amd64,linux/arm64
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! Please see the reasons in #3034 why I'm currently not considering merging it. A third-party image is already provided via https://github.com/afritzler/mkdocs-material. I'll reconsider as soon as I'm on M1 myself, but for the time being, I have no way to test this image, so I need to rely on a third party.

caarlos0:
fwiw You should be able to test the image on an intel mac/linux with something like:

```
docker run --platform linux/arm64 image
```

squidfunk:
Thanks. Okay, let me check again.

PR #4243, opened 2022-08-13
Add and correct Hebrew translations

amire80:


src/partials/languages/he.html
```diff
@@ -24,22 +24,42 @@
...
+  "header.title": "כותרת עליונה",
   "meta.comments": "הערות",
-  "meta.source": "מָקוֹר",
+  "meta.source": "מקור",
+  "nav.title": "ניווט",
+  "search.config.lang": "he",
   "search.config.pipeline": " ",
-  "search.placeholder": "לחפש",
-  "search.result.placeholder": "הקלד כדי להתחיל לחפש",
+  "search.config.separator": "[\\s\\-]+",
```
squidfunk, line 43: The separator is only necessary to provide if it deviates from the standard one defined for `en` which it doesn't. Please remove this line. I'm going to improve the documentation for providing and updating translations in the near future.

squidfunk (changes requested):

src/partials/languages/he.html
amire80: Thanks for the comment. Done. I actually had a suspicion about it.

src/partials/languages/he.html
```diff
@@ -24,22 +24,41 @@
...
+  "edit.link.title": "עריכת הדף הזה",
+  "footer.previous": "הקודם",
+  "footer.next": "הבא",
+  "footer.title": "כותרת תחתונה",
+  "header.title": "כותרת עליונה",
   "meta.comments": "הערות",
-  "meta.source": "מָקוֹר",
+  "meta.source": "מקור",
+  "nav.title": "ניווט",
+  "search.config.lang": "he",
```
squidfunk, line 41: Oh, also, `he` is not available as a search language. Please remove. Search should work without this separator.

squidfunk (changes requested):

src/partials/languages/he.html
amire80: Done.

squidfunk:
Thanks! LGTM.

PR #4281, opened 2022-08-26
Cast all tags as strings to avoid TypeError during highlight

NeilHanlon:
If a tag is specified which is an integer and is unquoted in the source
document, it is parsed as an integer which cannot have .replace() called
on it during the highlight function [1], otherwise it will yield a
TypeError.

For example, in the tags below, the first tag will evaluate as a number,
not a string, and causes a TypeError to occur.

```
...
tags:
  - 9.0
  - some other tags
  - 9.0 with more text
...
```

There is likely a more appropriate way to fix this, but it does seem to
do the job. The error was a bit opaque (just some "Invalid query:
<search string>"), but I'm not sure that's worth addressing, either.

[1] https://github.com/squidfunk/mkdocs-material/blob/9655c3a92471f261533d48b8611a8d24dbfebb13/src/assets/javascripts/integrations/search/highlighter/index.ts#L88

### DIFF
--- a/src/assets/javascripts/integrations/search/_/index.ts
+++ b/src/assets/javascripts/integrations/search/_/index.ts
@@ -266,7 +266,7 @@ export class Search {
                 location,
                 title: highlight(title),
                 text:  highlight(text),
-                ...tags && { tags: tags.map(highlight) },
+                ...tags && { tags: tags.map(tag => highlight(String(tag))) },
                 score: score * (1 + boost),
                 terms
               })


squidfunk:
Thanks for the PR! I think it's a better idea that we don't let numeric tags land in the search index in the first place, thus the fix you're proposing fixes a symptom but not the cause of the problem. I went ahead and improved the tags plugin to ensure all tags are strings in 4c4e3c0ae, which also fixes the issue at hand.

NeilHanlon:
like I said, I'm sure there was a better place to fix it!

Thank you for the quick fix ☺

PR #4285, opened 2022-08-26
Update

Justman10000:


squidfunk:
Thanks for the PR. However, I'm not sure why the default copyright should be set? it would mean that my name would appear on every Material for MkDocs built page if the authors doesn't set a default. This makes no sense whatsoever.

Closing for the reasons stated.

Justman10000:
> Thanks for the PR. However, I'm not sure why the default copyright should be set? It would mean that my name would appear on every Material for MkDocs built page if the author doesn't set a default. This makes no sense whatsoever.
> 
> Closing for the reasons stated.

Instead of closing, can't mn just wait for an answer? Boy it would have done you good because YOU would have been MENTIONED! Then do your shit self, I put time and effort into it

squidfunk:
@Justman10000 regardless of whether you agree with my decision, there's no reason to be rude. In the future, please use inclusive and welcoming language, as stated in our [code of conduct](https://github.com/squidfunk/mkdocs-material/blob/master/CODE_OF_CONDUCT.md).

Justman10000:
> @Justman10000 regardless of whether you agree with my decision, there's no reason to be rude. In the future, please use inclusive and welcoming language, as stated in our [code of conduct](https://github.com/squidfunk/mkdocs-material/blob/master/CODE_OF_CONDUCT.md).

Ah, good to know you don't care about anyone else's time and effort... I'll keep that in mind...

PR #4351, opened 2022-09-13
Upgrade Mermaid to v9.1.6

andrewimeson:
I needed a bug fix in a newer version of Mermaid. I worked around it in my own use case by vendoring the dependency, but thought I'd take a swing at upgrading it upstream.

The testing I did (aside from that it builds) was to go to the [Mermaid page](http://127.0.0.1:8000/mkdocs-material/reference/diagrams/) and confirm the graphs looked like they rendered properly.

Release notes:
  https://github.com/mermaid-js/mermaid/releases


squidfunk:
Thanks for the PR! I also checked the changes and indeed nothing seems to break. However, I updated Mermaid to 9.1.7 which is even newer. Superseded by 30d812888

PR #4357, opened 2022-09-14
Add document for correct rendering social card previews

jasonren0403:
First-time Asian language users may be stucked with bad characters 😄


squidfunk:
Thanks!

PR #4383, opened 2022-09-19
Add modeline yaml schema method documentation

t-mart:
In addition to adding a new yaml schema in VS Code's settings, you can also add a [modeline comment](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml#associating-a-schema-in-the-yaml-file) directly to `mkdocs.yml` to enable validation.

mkdocs-yaml

I think this method is ideal, especially in collaborative environments because it's one less step to perform if you are working with an already-annotated file.

Therefore, I have added this method to the documentation. The original method is still present, but both now are in a sub list of the original instruction step.

### DIFF
--- a/docs/creating-your-site.md
+++ b/docs/creating-your-site.md
@@ -4,7 +4,7 @@ template: overrides/main.html
 
 # Creating your site
 
-After you've [installed] Material for MkDocs, you can bootstrap your project 
+After you've [installed] Material for MkDocs, you can bootstrap your project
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. However, this is [exactly what we recommend](https://squidfunk.github.io/mkdocs-material/creating-your-site/#__tabbed_3_2).

t-mart:
My bad 🤦‍♂️

Got tunnel-visioned on just looking at the VS Code section.

PR #4386, opened 2022-09-20
Update MathJaX Config file in mkdocs, compatible with Promises, in case of instant loading

rod2ik:
FYI, I am using the latest MathJaX@3.2.2, so my mkdocs.yml file contains the following, but that has nothing to do.

```yaml
markdown_extensions:
  - pymdownx.arithmatex:
      generic: true

extra_javascript:
  - javascripts/mathjax.js
  - https://polyfill.io/v3/polyfill.js?features=es5,es6,es7,es8&flags=gated
  - https://cdn.jsdelivr.net/npm/mathjax@3.2.2/es5/tex-mml-chtml.js
```

Just update the `MathJax.startup.promise` part of the docs, that's all there is to it.
Thanks !

### DIFF
--- a/docs/reference/mathjax.md
+++ b/docs/reference/mathjax.md
@@ -38,7 +38,9 @@ lines to `mkdocs.yml`:
     };
 
     document$.subscribe(() => { // (1)!
-      MathJax.typesetPromise()
+    MathJax.startup.promise
+      .then(() => MathJax.typesetPromise())
+      .then(() => { /* other actions once MathJax is done */ })
     })
     ```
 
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! Could you please explain why you believe this is necessary? We're running latest MathJax on our documentation and it works without any problems with `navigation.instant`:

Bildschirmfoto 2022-10-01 um 17 21 58

squidfunk:
Closing for the reasons stated. If you feel that this is still necessary, please comment and reopen.

PR #4392, opened 2022-09-22
Ensure latest version source

ofek:
Address https://github.com/squidfunk/mkdocs-material/pull/4341#issuecomment-1253398625

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@
 # IN THE SOFTWARE.
 
 [build-system]
-requires = ["hatchling", "hatch-requirements-txt", "hatch-nodejs-version"]
+requires = ["hatchling", "hatch-requirements-txt", "hatch-nodejs-version>=0.3.0"]
 build-backend = "hatchling.build"
 
 [project]


ofek:
Should be good to merge 👍

squidfunk:
Thanks for the PR. I'm discovering more and more that the ecosystem doesn't seem to be ready for the switch, e.g. PyPI stats can't resolve author, license, etc. anymore: https://pypistats.org/packages/mkdocs-material

I'm not sure this is ready for primetime.

ofek:
Hmm, cc @hugovk

hugovk:
https://pypi.org/pypi/mkdocs-material/json contains:

```json
{
"info": {
"author": "",
...
"license": "",
...
```

I see https://github.com/squidfunk/mkdocs-material/blob/master/pyproject.toml also contains no `license` or `authors` under `[project]`.


Compare:
```toml
license = {text = "MIT"}
authors = [{name = "Hugo van Kemenade"}]
```

* https://github.com/hugovk/pepotron/blob/885bd0b111202976b04dbac4950ea057bf706aa3/pyproject.toml#L17-L18

And they can be see at:

* https://pypi.org/project/pepotron/
* https://pypistats.org/packages/pepotron
* https://pypi.org/pypi/pepotron/json

```json
{
"info": {
"author": "Hugo van Kemenade",
...
"license": "MIT",
...
```

Should they be added here?

ofek:
1. So for the license see https://github.com/pypa/packaging.python.org/pull/1143, basically until PyPI supports PEP 639 one must use classifiers
2. The author in wheel is `Author-email: Martin Donath <martin.donath@squidfunk.com>` which is [also how setuptools works](https://github.com/pypa/packaging.python.org/issues/1134#issuecomment-1231564237) 

Both show correctly here under Meta tab https://pypi.org/project/mkdocs-material/

squidfunk:
I'm considering this ready for merge then. PyPI stats is not critical, but it would be great if this could be solved and reported. I'm not very familiar with the Python packaging ecosystem, so I would be thankful if one of you could take this on.

squidfunk:
So, as a temporary measure, couldn't we just add back all the information that was present in the setup.py before? This would at least fix the issues..

https://github.com/squidfunk/mkdocs-material/blob/cccc21431e8e986a6d51aaba6a212db231a5eb09/setup.py#L40-L78

ofek:
That wouldn't fix it, here's the known issue https://github.com/pypi/warehouse/issues/9400

PR #4401, opened 2022-09-25
Meta (Facebook)

UlviShukurzade:
Company rebranded


squidfunk:
Thanks for the PR! Yes, Facebook is now Meta. However, it's still hosted under Facebook on GitHub and even [uses the name Facebook in their README](https://github.com/facebook/docusaurus#contributing). If that changes, we should also change it.

PR #4424, opened 2022-09-29
Added MagicLink Extension Overview

ruudvh:
Added short overview of the MagicLink extension and the available supported options.

I was unsure when this feature was added so I used `3.3.0` which was the first release which included options for the MagicLink extension (note: MagicLink was included in earlier version). Based on a history search of [mkdocs.yml](https://github.com/squidfunk/mkdocs-material/blob/master/mkdocs.yml).

Happy to add a reference page with examples if I could get some color under which page it should fall (i.e. `Formatting`) or should create a new page (i.e. `Linking`).

Thank you,

### DIFF
--- a/docs/setup/extensions/python-markdown-extensions.md
+++ b/docs/setup/extensions/python-markdown-extensions.md
@@ -729,3 +729,58 @@ See reference for usage:
   [GitHub Flavored Markdown]: https://github.github.com/gfm/
   [Tasklist specification]: https://github.github.com/gfm/#task-list-items-extension-
   [Using task lists]: ../../reference/lists.md#using-task-lists
+  
+### MagicLink
+
+[:octicons-tag-24: 3.3.0][MagicLink support] ·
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! You're correct that MagicLink is supported, but it works without any changes to Material for MkDocs. This means that it should basically be supported since version 0.1.0. However, since we don't need a specific configuration (like for [Tabbed](https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown-extensions/#tabbed)), and don't provide any special integration, I'm not sure whether we should add it to the documentation, as we would need to keep track of changes to MagicLink without any particular upside.

Thinking about it, I think it would probably be a better idea to add a short paragraph directly under "Supported Extensions" saying that all other Python Markdown Extensions should work without problems, regardless of their configuration. If a user finds an extension not to be working with a specific configuration, we should add it to the documentation (as mentioned above).

squidfunk:
Added in 1e508d629.

PR #4434, opened 2022-10-01
Add external plugins, schema not provided

bangseongbeom:


squidfunk:
Thanks for the PR. Please provide more information why we should merge this.

bangseongbeom:
@squidfunk Because I've got a warning when using the extensions (e.g. [mkdocstrings](https://mkdocstrings.github.io/)).

squidfunk:
It's a better idea to take this upstream to the owner of the repository and ask him to add support, or add a schema to our external supported plugins and extensions [here](https://github.com/squidfunk/mkdocs-material/tree/master/docs/schema/plugins/external). If we start adding wildcards, you lose the safety of validation because now typos go undetected. For instance, if we would merge your PR, this would now be considered valid:

``` yaml
plugins:
  - social # should be social
```

For this reason I don't think this is a good fit.

PR #4491, opened 2022-10-11
Update docs for new GitHub Pages deployment method

br3ndonland:
## Description

GitHub Pages now supports two different deployment methods:

1. Committing files to a branch like `gh-pages` (the old method using [`mkdocs gh-deploy`](https://www.mkdocs.org/user-guide/deploying-your-docs/) and [`ghp-import`](https://github.com/c-w/ghp-import))
2. Publishing a build artifact (the [new default method](https://github.blog/2022-08-10-github-pages-now-uses-actions-by-default/)). As the [GitHub docs](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site) explain, the advantage of publishing a build artifact is that the compiled site files do not need to be committed to the repo.

Material for MkDocs provides a helpful page on [publishing your site](https://squidfunk.github.io/mkdocs-material/publishing-your-site/). It would be great to update this page for the new deployment method, especially because the new method is now the default for GitHub Pages.

## Changes

This PR will update the page on [publishing your site](https://squidfunk.github.io/mkdocs-material/publishing-your-site/) to:

- Explain the difference between the two deployment methods
- Provide resources for the new method (publishing a build artifact)

## Related

- [MkDocs: Deploying your docs](https://www.mkdocs.org/user-guide/deploying-your-docs/)
- [Material for MkDocs: Publishing your site](https://squidfunk.github.io/mkdocs-material/publishing-your-site/)
- [GitHub Blog 2022-08-10: GitHub Pages now uses Actions by default](https://github.blog/2022-08-10-github-pages-now-uses-actions-by-default/)
- [GitHub Changelog 2022-08-10: GitHub Pages: Builds with GitHub Actions GA](https://github.blog/changelog/2022-08-10-github-pages-builds-with-github-actions-ga/)
- [GitHub docs: GitHub Pages - Configuring a publishing source for your GitHub Pages site](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site)
- [`actions/deploy-pages`](https://github.com/actions/deploy-pages)
- [Workflow example from `actions/starter-workflows`](https://github.com/actions/starter-workflows/blob/main/pages/static.yml)
- mkdocs/mkdocs#2144
- mhausenblas/mkdocs-deploy-gh-pages#188

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -16,12 +16,24 @@ charge and pretty easy to set up.
 
   [GitHub Pages]: https://pages.github.com/
 
-### with GitHub Actions
+GitHub offers two options for configuring a GitHub Pages [publishing source]:
 
-Using [GitHub Actions] you can automate the deployment of your project
-documentation. At the root of your repository, create a new GitHub Actions
-workflow, e.g. `.github/workflows/ci.yml`, and copy and paste the following
-contents:
+1. Publishing from a build artifact with GitHub Actions
+2. Publishing from a branch
+
+### Publishing from a build artifact with GitHub Actions
+
+Using [GitHub Actions], you can automate the deployment of your project
+documentation. Publishing from a build artifact with GitHub Actions is the
+[default GitHub Pages deployment method].
+
+As the GitHub Pages docs on configuring a [publishing source] explain, the
+advantage of publishing a build artifact is that the compiled site files do not
+need to be committed to the repo.
+
+To deploy your documentation with GitHub Actions, create a YAML file in the
+`.github/workflows` directory, e.g. `.github/workflows/ci.yml`, and copy and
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. Since this is pretty new, I'm not sure we can recommend switching. Note that some tools like `mike` rely on the `gh-pages` branch being present. I suspect that versioning will not work with the new method. Thus, I'm not feeling confident to recommend it. Additionally, and this is the most important thing, the getting started guide got a whole lot more complicated. My aim is to make it as simple as it can possibly be, and this sometimes incurs the use of older (but proven) methods. For this reason, I currently can't consider merging this PR.

PR #4504, opened 2022-10-16
Add namespace to plugins and update to MkDocs 1.4.1

squidfunk:
Before merging this, we need to make sure that the Docker image build check still works. [Reported here](https://gitter.im/squidfunk/mkdocs-material?at=6352e6fc27f328266d4d053c).


squidfunk:
Superseded by #4628.

PR #4584, opened 2022-11-01
Revert to pre-8.0.0 simpler CSS selectors

squidfunk:
This PR reverts to the pre-8.0.0 behavior for simple selector lists using `:is(...)`, which overcomplicated overrides. When this PR is merged, CSS classes are again much easier to override and some specificity hacks are not necessary anymore. While the size of the CSS will slightly increase, it should be roughly the same after running through gzip.

Additionally, all alternative admonition qualifiers have been removed after they have been deprecated.


squidfunk:
Superseded by #4628.

PR #4615, opened 2022-11-09
Added missing step to create a new site

damascene:
Quick start instruction jumps to editing `mkdocs.yml` file before creating a new site. Install doesn't create the file so the instruction seems missing. 

I've added the step to create a new site to the quick start as in the [official documentation](https://squidfunk.github.io/mkdocs-material/creating-your-site/).

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,11 @@ Material for MkDocs can be installed with `pip`:
 pip install mkdocs-material
 ```
 
+Create new site:
+```
+mkdocs new .
+```
+
 Add the following lines to `mkdocs.yml`:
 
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. However, this is a quick start guide. It assumes that you have MkDocs already installed. I don't want to blow up the README. If the user reads the guide and doesn't know what it's about, the user should follow the link below:

> For detailed installation instructions, configuration options, and a demo, visit [squidfunk.github.io/mkdocs-material](https://squidfunk.github.io/mkdocs-material/)

PR #4628, opened 2022-11-12
Material for MkDocs 9

squidfunk:
This PR contains all changes for the next major version of Material for MkDocs, which will be released as soon as the [Carolina Reaper](https://squidfunk.github.io/mkdocs-material/insiders/#10000-carolina-reaper) funding goal is hit. The biggest addition will be the new built-in search, which is finally released into the community edition. Sponsors have been using the new search [for more than a year](https://squidfunk.github.io/mkdocs-material/insiders/changelog/#3.0.0).

The following major changes are included (__sponsors have been using most of them for over a year__):

- [x] Rewrite of search plugin with support for [rich previews](https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-search/#rich-search-previews)
- [x] Detached search plugin implementation from MkDocs search plugin
- [x] Added support for search [tokenizer lookahead](https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-search/#tokenizer-lookahead) 
- [x] Added support for [better search highlighting](https://squidfunk.github.io/mkdocs-material/blog/2021/09/13/search-better-faster-smaller/#accurate-highlighting)
- [x] Added support for [excluding content from search](https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-search/#search-exclusion)
- [x] Added support for [configurable search pipeline](https://github.com/mkdocs/mkdocs/issues/2380)
- [x] Added native support for offline search via [offline plugin](https://squidfunk.github.io/mkdocs-material/setup/building-for-offline-usage/)
- [x] Omit code block line numbers from search
- [x] Improved stability of search plugin when using `--dirtyreload`
- [x] Removed support for prebuilding search indexes 
- [x] Removed support for indexing only titles in search
- [x] Fixed detection of composition events in search interface
- [x] Fixed search plugin not using title set via front matter
- [x] Rewrite of worker communication layer

Additionally, since we're releasing a new major version, we take the opportunity to remove several deprecations and add some more feature flags to allow for more flexibility, the most important of which:

- [x] [Removed alternative admonition qualifiers](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#fn:1) in https://github.com/squidfunk/mkdocs-material/pull/4584/commits/daffd085f3bc35058ada32fad0a21b2d88ac0d3f
- [x] Removed `:is` selectors (in output) to allow for easier overriding of CSS in daffd085f3bc35058ada32fad0a21b2d88ac0d3f
[...]

PR #4644, opened 2022-11-16
Fix: Norwegian translations

tersor:
Updated Norwegian translations with latest strings from `en`.

Also, there are two official languages in Norway: `Norwegian nynorsk` and `Norwegian bokmål`. Therefore, the Norwegian bokmål file has been moved to use the correct locale, which is **`nb`** according to the [ISO 639-1 standard](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes).

Most libraries today offer either `nb` or `nn` locale for Norwegian. One example is [Moment.js](https://github.com/moment/moment/tree/develop/locale).


squidfunk:
Thanks for the PR. So what about `no`? Is it not necessary anymore? Can you provide it as part of the PR as well? I'm sorry if the question is dumb, but I'm not speaking Norwegian, so I'm dependent on collaborators like you 😉 We could change to `nb` in version 9 (see #4628, because it would be a breaking change for all users using `no`).

PR #4647, opened 2022-11-16
Fix: Norwegian translations

tersor:
@squidfunk Agree! Removing `no` could be a breaking change, better to do it in next major release.


squidfunk:
Thanks for providing the updated PR. Unfortunately, not all of my qquestions were answered – does it make sense to have support for `no`? Does this PR contain actual translations for `no` or is it a copy of `nb`? Don't worry about the breaking change, it's my job to take care of it, but I need more information to decide what to do exactly.

tersor:
> does it make sense to have support for no?

No, it does not make any sense. The `no` locale should be avoided since it is a [macrolanguage](https://en.wikipedia.org/wiki/ISO_639_macrolanguage). It is much more preferable to use the code for Norwegian bokmål (variant of the Norwegian language) according to the [ISO 639-1 standard](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes), which is **`nb`**.

> Does this PR contain actual translations for no or is it a copy of nb?

It is a copy of `nb` to avoid any breaking change. Historically (and unfortunately) `no` has been used as the locale for `Norwegian bokmål`. But most modern libraries today use `nn` and `nb` exclusively.

tersor:
@squidfunk done!

src/partials/languages/nn.html
```diff
@@ -23,22 +23,43 @@
...
   "edit.link.title": "Rediger denne sida",
   "footer.previous": "Førre",
   "footer.next": "Neste",
+  "footer.title": "Footer",
+  "header.title": "Header",
   "meta.comments": "Kommentarar",
   "meta.source": "Kjelde",
-  "search.config.lang": "no",
+  "nav.title": "Navigasjon",
+  "search.config.lang": "nn",
```
squidfunk, line 41: Please remove the changes for `search.config.*` (here and in the other translation). I'm pretty confident that with those lines present, search will not work. [Lunr languages](https://github.com/MihaiValentin/lunr-languages#lunr-languages-----) only supports `no`.

src/partials/languages/nn.html
```diff
@@ -23,22 +23,43 @@
 <!-- Translations: Norwegian Nynorsk -->
 {% macro t(key) %}{{ {
   "language": "nn",
+  "direction": "ltr",
```
squidfunk, line 26: Please remove `direction`. It's inherited from `en`, so not necessary.

squidfunk (changes requested):
See my comments and our [contributing guide](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md#missing-translations).

src/partials/languages/nn.html
tersor: Much appriciated feedback @squidfunk! Sorry for not reading the **entire** contribution guide, my bad 🙄 

I notice in [lunr.no.js](https://github.com/MihaiValentin/lunr-languages/blob/master/lunr.no.js) that it mixes _both_ languages (`nn and `nb`), so I think the best to define `"search.config.lang": "no"`. Then the search will work too.

squidfunk (approved):

squidfunk:
All good! I've merged your changes into the v9 branch, tracked in #4628.

gallaher:
Error message using Norwegian Bokmål:
@squidfunk Using the follwing language code for norwgian nynorsk, nn, works perfectly:
theme: 
    name: material
    language: nn

However, using the code for Norwegian Bokmål, nb, does not work:
theme: 
    name: material
    language: nb

This combination gives an error message and tells me the web page is unavailable. Is there a way to choos Norwegian Bokmål as the language for the page?

Regards,
Martin

gallaher:
Thanks, I thought I had installed the latest, but it was v.8.3.3 :-( After updating to the latest revision everything works.

PR #4672, opened 2022-11-23
Gitlab example using rules

kgoossens:
Adapted the 2 examples to use the `rules` keyword and to run the pages:deploy job on the default branch (master/main/... whatever is defined)

```
only and except are not being actively developed. [rules](https://docs.gitlab.com/ee/ci/yaml/#rules) is the preferred keyword to control when to add jobs to pipelines.
```
[Source: Only/Except](https://docs.gitlab.com/ee/ci/yaml/#only--except) 

Hope this is useful.


squidfunk:
Thanks for the PR! I'm not a GitLab user, so I'm very much dependent on collaborators like you. Maybe some other GitLab user can comment? Otherwise we can merge your suggestions and see if somebody runs into problems.

squidfunk:
Okay, since `only` and `except`. are deprecated, I'm merging this right now.

kgoossens:
Glad I can help @squidfunk 

I have it running [here](https://gitlab.com/kgoossens/kgoossens.gitlab.io/-/blob/41fe3c94353b4e557c48267f5902f58e9c948573/.gitlab-ci.yml) and err... if you like to become a GitLab user, let me know, I might know some people 😜

PR #4712, opened 2022-12-06
Updated Dockerfile to Python 3.11 on Alpine 3.17

### LINKED ISSUE #4711: Dockerfile base python image hasn't been updated in 2 years and has known vulnerabilities
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the problem doesn't occur with the `mkdocs` or `readthedocs` themes
- [X] ... the problem persists when all overrides are removed, i.e. `custom_dir`, `extra_javascript` and `extra_css`

### Description

The currently configured base image in the docker file, python:3.9.2-alpine3.13 has known critical vulnerabilities and needs to be updated to a newer image.

### Expected behaviour

Not failing vulnerabilities checks.

### Actual behaviour

fails vulnerability check.

CVE-2022-42915

### Steps to reproduce

run vulnerability scan.

### Package versions

python:3.9.2-alpine3.13

### Configuration

```yaml
standard
```

### System information

python:3.9.2-alpine3.13

vedranmiletic:
Fixes #4711.

Tested using [gaseri/website](https://github.com/gaseri/website).

Running emits:

```
INFO     -  DeprecationWarning: Use setlocale(), getencoding() and getlocale() instead
File "/usr/local/lib/python3.11/site-packages/cairosvg/features.py", line 9, in <module>
LOCALE = locale.getdefaultlocale()[0] or ''
File "/usr/local/lib/python3.11/locale.py", line 559, in getdefaultlocale
warnings.warn(
```

This should likely be fixed by upgrading CairoSVG later.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,13 +18,13 @@
 # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
 # IN THE SOFTWARE.
 
-FROM python:3.9.2-alpine3.13
+FROM python:3.11.0-alpine3.17
 
 # Build-time flags
 ARG WITH_PLUGINS=true
 
 # Environment variables
-ENV PACKAGES=/usr/local/lib/python3.9/site-packages
+ENV PACKAGES=/usr/local/lib/python3.11/site-packages
 ENV PYTHONDONTWRITEBYTECODE=1
 
 # Set build directory


squidfunk:
Thanks! Could we also upgrade CairoSVG in the process? Otherwise users will complain about the deprecation warnings.

vedranmiletic:
I checked and unfortunately there is no newer upstream release of CairoSVG and the last release was made in mid-2021.

There is an option to use Python 3.10 to avoid the deprecation warning which will avoid the user complaints. However, I would lean towards using Python 3.11 here as it is 10-15% faster than 3.10 when building documentation, which is a nice improvement.

What would you prefer?

squidfunk:
Thanks for investigating! Faster is a better argument than less noise, so deprecation warnings it is for now. Thus, this PR can be merged as it is right now, correct?

vedranmiletic:
Sure. Maybe it would be a good idea to add a general note in the docs for Python 3.11 (that specifically applies to the Docker image) stating that CairoSVG will emit a depecation warning which can be safely ignored?

squidfunk:
> Maybe it would be a good idea to add a general note in the docs for Python 3.11 (that specifically applies to the Docker image) stating that CairoSVG will emit a depecation warning which can be safely ignored?

Let's keep it short and concise. If somebody trips over the warning and creates an issue, we can link here.

squidfunk:
I've just set up a brand new M2 and don't see the warning in `cairosvg` – this was fixed in [2.6.0](https://github.com/Kozea/CairoSVG/releases/tag/2.6.0)!

PR #4723, opened 2022-12-09
Add 'yaml.customTags' to 'settings.json' example

TetraTheta:
Without `yaml.customTags` in `.vscode/settings.json`, these lines will be marked as error.
``` yaml
markdown_extensions:
  - pymdownx.emoji:
      emoji_index: !!python/name:materialx.emoji.twemoji
      emoji_generator: !!python/name:materialx.emoji.to_svg
```

So I added `yaml.customTags` part in example `settings.json` in the documentation.

### DIFF
--- a/docs/creating-your-site.md
+++ b/docs/creating-your-site.md
@@ -58,12 +58,12 @@ theme:
         2.  Add the schema under the `yaml.schemas` key in your user or
             workspace [`settings.json`][settings.json]:
 
-            ``` json
+            ``` { .json .annotate }
             {
               "yaml.schemas": {
                 "https://squidfunk.github.io/mkdocs-material/schema.json": "mkdocs.yml"
               },
-              "yaml.customTags": [
+              "yaml.customTags": [ // (1)!
                 "tag:yaml.org,2002:python/name:materialx.emoji.to_svg",
                 "tag:yaml.org,2002:python/name:materialx.emoji.twemoji",
                 "tag:yaml.org,2002:python/name:pymdownx.superfences.fence_code_format"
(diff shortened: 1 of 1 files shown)


docs/creating-your-site.md
```diff
@@ -62,10 +62,17 @@ theme:
...
+              },
+              "yaml.customTags": [
+                "tag:yaml.org,2002:python/name:materialx.emoji.to_svg",
+                "tag:yaml.org,2002:python/name:materialx.emoji.twemoji",
+                "tag:yaml.org,2002:python/name:pymdownx.superfences.fence_code_format"
+              ]
             }
             ```
 
+            `yaml.customTags` part will be needed if you want to use and configure [Emojis](reference/icons-emojis.md).
```
squidfunk, line 74: Could you please move this into a [code annotation](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#code-annotations) on line 66 in the source file?

squidfunk (changes requested):
Thanks! Only a minor change necessary 😊

docs/creating-your-site.md
TetraTheta: I think I moved it into code annotation

squidfunk:
Okay. My fault. JSON doesn't support comments, so no point in using annotations. Sorry. Could you just revert the latest commit? I'll merge it then. Sorry for the inconvenience 😅

TetraTheta:
> Okay. My fault. JSON doesn't support comments, so no point in using annotations. Sorry. Could you just revert the latest commit? I'll merge it then. Sorry for the inconvenience 😅

I think current way is better looking. There is [JSON5](https://json5.org/), which is superset of JSON and also supports comments. Even VSCode makes part starts with `//` green which indicates 'comment' in `.json` file by default.

So, I think using code annotation looks better.

If you still think previous way is better, I will revert code annotation.

squidfunk:
Wow, since when does this work? It wasn't possible for a long time to add comments to JSON, so code annotations would not work. I just tested it and it works fine. Related:

- #3667

That said, it looks great!

squidfunk (approved):

PR #4751, opened 2022-12-12
feat: grouping tags option for tags plugin

holamgadol:
Sometimes _tags_ page may look like a big solid list of links under the tags labels. I think it would be great to group page links by their urls. It will increase the readability of the whole _tags_ page.
Proposed syntax:

```yaml
plugins:
  - tags:
      tags_file: tags.md
      group_by: path
```
### Usage example

Suppose you have some articles in different directories with the same tag `africa`

- ungrouped tags:

  image

- grouped tags:

  image


squidfunk:
Thanks for the PR. I need some time to investigate this proposed change. Also, it gives us the chance to see if it would be useful to other users.

bigpick:
I think a tag grouping would be nice - but if adding in grouping, I'd vote for also adding in dropdowns/the ability to collapse sub-categories alongside it, similarly to how the nav bar functions currently.

holamgadol:
> I think a tag grouping would be nice - but if adding in grouping, I'd vote for also adding in dropdowns/the ability to collapse sub-categories alongside it, similarly to how the nav bar functions currently.

Do you mean collapse only groups or groups and tags both?

squidfunk:
So, thanks for suggesting! I'm currently not considering merging this PR, as one can see from the discussion in this issue, the way tags should be grouped may be dependent on the use case. Some users might want to group by path, some might want to have a hierarchy of tags. I'm currrently working on a slight refactoring of the tags plugin that will allow for more flexibility, i.e. to provide your own template or grouping mechanisms.

However, you are free to fork the built-in tags plugin with your changes applied and provide it as a third-party plugin – Material for MkDocs is MIT licensed so this is possible and even encouraged!

holamgadol:
Thank you for the comprehensive response. Yes, a third-party tags plugin for mkdocs-material is a great idea.

PR #4770, opened 2022-12-16
Add plugins schema

TetraTheta:
This PR adds schema of these MkDocs plugins:
  * [literate-nav](https://github.com/oprypin/mkdocs-literate-nav)
  * [macros](https://github.com/fralau/mkdocs_macros_plugin)
 
Also this adds missing properties of this MkDocs plugin:
  * [minify](https://github.com/byrnereese/mkdocs-minify-plugin)
      * `cache_safe`

NOTE: Although `macros` plugin supports `!env` at `on_error_fail` property, I couldn't find a way to support both `boolean` and `string` starts with `!env`. 

I tested this with my MkDocs project. No 'error' was printed and VSCode thinks the `mkdocs.yml` is well formed  which is normal.

### DIFF
--- a/docs/schema/plugins.json
+++ b/docs/schema/plugins.json
@@ -58,6 +58,12 @@
           "$ref": "plugins/external/git-revision-date.json"
         },
         {
+          "$ref": "plugins/external/literate-nav.json"
+        },
+        {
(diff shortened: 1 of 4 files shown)


squidfunk:
Thanks!

PR #4773, opened 2022-12-18
fix: some new words were missing in Italian translations

StefanoCecere:
some new words were missing

### DIFF
--- a/src/partials/languages/it.html
+++ b/src/partials/languages/it.html
@@ -23,7 +23,6 @@
 <!-- Translations: Italian -->
 {% macro t(key) %}{{ {
   "language": "it",
-  "direction": "ltr",
   "announce.dismiss": "Non mostrare più",
   "clipboard.copy": "Copia",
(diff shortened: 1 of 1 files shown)


src/partials/languages/it.html
```diff
@@ -23,8 +23,13 @@
 <!-- Translations: Italian -->
 {% macro t(key) %}{{ {
   "language": "it",
+  "direction": "ltr",
```
squidfunk, line 26: Please remove this, together with the search translations mentioned in https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md#missing-translations.

squidfunk (changes requested):

src/partials/languages/it.html
StefanoCecere: ops. done. thank you

squidfunk:
Thanks!

Issue #4783, opened 2022-12-20
Mermaid sequence diagrams: alt fragment titles rendering in odd ways


imkacarlson:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the problem doesn't occur with the `mkdocs` or `readthedocs` themes
- [X] ... the problem persists when all overrides are removed, i.e. `custom_dir`, `extra_javascript` and `extra_css`
- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem - there is a related problem but not exactly the same: #4713

### Description

Mermaid's sequence diagrams support alt fragements: https://mermaid.js.org/syntax/sequenceDiagram.html#alt

They work great! However in light mode some of the titles are bolded and then when switching to dark mode some of the titles are unreadable.

Light mode issue:
![image](https://user-images.githubusercontent.com/31493326/208743352-90d2a837-b632-48b9-af69-0b908a4b75e3.png)

Dark mode issue:
![image](https://user-images.githubusercontent.com/31493326/208743471-9a8951c8-158f-44f6-a45c-519bb48e98fe.png)

### Expected behaviour

The titles should be readable and consistent fonts

### Actual behaviour

Some titles are bold, and black on black text is not readable

### Steps to reproduce

Create a sequence diagram like so:

```
sequenceDiagram
    Alice->>Bob: Hello Bob, how are you?
    alt is sick
        Bob->>Alice: Not so good :(
    else is well
        Bob->>Alice: Feeling fresh like a daisy
    end
    opt Extra response
        Bob->>Alice: Thanks for asking
    end
```
Toggle between light and dark mode. 

### Package versions

- Python: Python 3.9.7
- MkDocs: mkdocs, version 1.4.2
- Material: Version: 8.5.11

### Configuration

```yaml
site_name: Test
nav:
  - Home: index.md
theme:
  name: material
  palette:
  - media: "(prefers-color-scheme: light)"
    scheme: default
    toggle:
      icon: material/brightness-7
      name: Switch to dark mode
  - media: "(prefers-color-scheme: dark)"
    scheme: slate
    toggle:
      icon: material/brightness-4
      name: Switch to light mode 
    primary: indigo
    accemt: yellow
markdown_extensions:
  - pymdownx.superfences:
[...]

PR #4784, opened 2022-12-20
Sequence diagram alt fragment title fix

imkacarlson:
Mermaid diagram alt fragment titles had some styling issues (see #4783 ) but here is a fix!


imkacarlson:
PR re-created in #4785

PR #4814, opened 2023-01-03
Update Indonesian language

Hans5958:
Seems the translations are a little bit outdated, so here's an update for it.


src/partials/languages/id.html
```diff
@@ -25,19 +25,37 @@
...
+  "search.result.more.one": "1 lagi di halaman ini",
+  "search.result.more.other": "# lagi di halaman ini",
+  "search.result.term.missing": "Tidak ada",
+  "select.language": "Pilih bahasa",
+  "select.version": "Pilih versi",
+  "source": "Ke repositori",
   "source.file.date.created": "Dibuat",
+  "source.file.date.updated": "Pembaruan terakhir",
+  "tabs": "Tab",
   "toc": "Daftar isi"
```
squidfunk, line 59: A comma is missing here – the build will not work.

squidfunk (changes requested):
Thanks for the PR! Please check that the language can actually be used – there are syntactical errors inside your translations.

src/partials/languages/id.html
```diff
@@ -25,19 +25,37 @@
...
   "footer.previous": "Sebelumnya",
   "meta.comments": "Komentar",
   "meta.source": "Sumber",
+  "nav": "Navigasi",
+  "search": "Cari",
   "search.config.pipeline": " ",
   "search.placeholder": "Cari",
+  "search.share": "Bagikan",
+  "search.reset": "Kosongkan",
+  "search.result.initializer": "Initializing search",
```
squidfunk, line 45: Also, could you please translate this placeholder?

src/partials/languages/id.html
Hans5958: This has been fixed.

src/partials/languages/id.html
Hans5958: I actually left this because I still a little bit hesitant. I translated it as "preparing search", does this fine to you?

Hans5958:
A little bit of request of advice: Is the `search.config.pipeline` fine as is (`" "`)?

src/partials/languages/id.html
squidfunk: Yes. This translation is shown when the search index is opened, but the search index is still building, so the user has to wait until it is ready. When it is ready, the `search.resut.placeholder` is shown, so "Ketik untuk mulai pencarian".

squidfunk:
Yes, please don't change `search.config.pipeline`, as described in our [contributing guide](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md#missing-translations).

squidfunk (approved):

src/partials/languages/id.html
Hans5958: Alright. I will leave it translated as that.

PR #4822, opened 2023-01-03
Explained why edit_uri requires edit string in URI

vedranmiletic:
Official MkDocs configuration mentions blob/main/docs/ as the path for the edit_uri when using GitHub. This will not work as intended with Material for MkDocs since blob will not get replaced by edit and raw in the resulting Code action button URLs.

### DIFF
--- a/docs/setup/adding-a-git-repository.md
+++ b/docs/setup/adding-a-git-repository.md
@@ -104,8 +104,12 @@ Some popular choices:
 
 If the [repository URL] points to a [GitHub], [GitLab] or [Bitbucket] repository,
 buttons for code actions can be added at the top of each document. Currently,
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. How about replacing `edit` and `blob` with `raw`?

vedranmiletic:
That would be neat, but we can take it even further, how about replacing `blob` with `edit` and `raw`, depending on the button?

squidfunk:
We shouldn't replace something in the `edit_uri`, that would be counter intuitive. However, as of c4d61cdc4, `blob` and `edit` with `raw`  if `content.action.view` is enabled. This also fixes the issue this PR was adressing.

vedranmiletic:
Your approach is indeed the right one. Thanks for addressing the issue.

PR #4836, opened 2023-01-06
Added partials/javascripts/nav-item.html, updated partials/nav-item.h…

mnazzaro:
…tml and base.html so that navigation labels would have accessibility tags and tabindex=0. Previously, they weren't tabbable

### DIFF
--- a/src/base.html
+++ b/src/base.html
@@ -314,6 +314,9 @@
                 </div>
               </div>
             {% endif %}
+
+            <!-- Accessibility for nav-items -->
+            {% include "partials/javascripts/nav-item.html" %}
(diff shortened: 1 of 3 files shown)


squidfunk:
Thanks! As a first feedback, This has to wander into the actual application core (not into inline scripts!), since it doesn't need to be run _immediately_ when the page is loaded. If it would be contained in inline scripts, each built site will be bigger.

squidfunk:
Superseded by https://github.com/squidfunk/mkdocs-material/commit/f8f44f58376206903e925d0b3db1e6e5f89e94ca.

PR #4847, opened 2023-01-10
Add support for Plausible Analytics

notpushkin:
I've implemented [Plausible](https://plausible.io/) support. Usage:

1. Add configuration to mkdocs.yml:

    ```yml
    extra:
      analytics:
        provider: plausible
        domain: example.com

        # If using custom domain proxy or self-hosting Plausible, specify script path:
        # src: "https://plausible.example.com/js/plausible.js"

        feedback:
          title: Was this page helpful?
          ratings:
            - icon: tabler/thumb-up
              name: This page was helpful
              data: good
              note: "Thanks for your feedback!"
            - icon: tabler/thumb-down
              name: This page could be improved
              data: bad
              note: "Thanks for your feedback! Help us improve this page by..."
    ```

2. In your Plausible account, go to your website's settings and visit the **"Goals"** section. For each rating defined, click on the **"+ Add goal"** button, select **Custom event** as the goal trigger and enter `Feedback: {rating data value}`.

    For example, if you have two ratings – `good` and `bad`, add `Feedback: good` and `Feedback: bad` goals.

Ratings will be shown in the **Goal Conversions** section at the very bottom of the page, as soon as any are available:

image

You can click on a specific “goal” to filter your dashboard by it. For example, if you filter by the `Feedback: bad` goal, you can see which pages need the most attention in the **Top Pages** section.

Let me know if you have any questions!

### DIFF
--- a/src/partials/integrations/analytics/plausible.html
+++ b/src/partials/integrations/analytics/plausible.html
@@ -1,0 +1,38 @@
+<script
+  defer
+  data-domain="{{ config.extra.analytics.domain }}"
+  src="{{ config.extra.analytics.src | default("https://plausible.io/js/plausible.js", true) }}"
+></script>
+
+<script>
+  window.plausible = window.plausible || function () { (window.plausible.q = window.plausible.q || []).push(arguments) };
+
+  /* Register event handlers after documented loaded */
+  document.addEventListener("DOMContentLoaded", function () {
+    /* Set up feedback, i.e. "Was this page helpful?" */
+    document$.subscribe(function () {
+      var feedback = document.forms.feedback;
+      if (typeof feedback === "undefined") return;
+
+      /* Send feedback to Plausible */
+      for (var button of feedback.querySelectorAll("[type=submit]")) {
+        button.addEventListener("click", function (ev) {
+          ev.preventDefault();
+
+          var page = document.location.pathname;
+          var value = this.getAttribute("data-md-value");
+          console.log("[feedback] User clicked", { value });
+          plausible(`Feedback: ${value}`);
+
+          /* Disable form and show note, if given */
+          feedback.firstElementChild.disabled = true;
+          var note = feedback.querySelector(".md-feedback__note [data-md-value='" + data + "']");
+          if (note) note.hidden = false;
+        })
+
+        /* Show feedback */
+        feedback.hidden = false;
+      }
+    });
+  });
+</script>


notpushkin:
(I'll look into the build failure in a bit; also gonna write up some docs if this looks good to you otherwise!)

squidfunk:
Thanks for providing. [Since this service is not free](https://plausible.io/#pricing), I can't commit to maintaining this change. You can provide this as a theme extension or third-party repository, but we can't merge this into `master`.

notpushkin:
The service is paid indeed, but it is open source and can be self-hosted. I can also provide an account on my instance – would that help?

Alternatively, any suggestions on how to best package it as an addon?

squidfunk:
> Alternatively, any suggestions on how to best package it as an addon?

It's probably a good idea to package this into a [custom theme](https://www.mkdocs.org/dev-guide/themes/#creating-a-custom-theme). In your theme, you can set `theme.name` to `material`, and then set `custom_dir` to your overrides:

> Generally, when building your own custom theme, the theme.[name](https://www.mkdocs.org/user-guide/configuration/#name) configuration setting would be set to null. However, if the theme.[custom_dir](https://www.mkdocs.org/user-guide/configuration/#custom_dir) configuration value is used in combination with an existing theme, the theme.[custom_dir](https://www.mkdocs.org/user-guide/configuration/#custom_dir) can be used to replace only specific parts of a built-in theme. For example, with the above layout and if you set name: "mkdocs" then the main.html file in the theme.[custom_dir](https://www.mkdocs.org/user-guide/configuration/#custom_dir) would replace the file of the same name in the mkdocs theme but otherwise the mkdocs theme would remain unchanged. This is useful if you want to make small adjustments to an existing theme.

As I read it, this can be used to package stuff up into a new installable theme. Other than that, we might come up with a better way to provide reusable overrides in the future. It might be a good idea to easier allow third party extensions, but this might need to be implemented in MkDocs itself.

notpushkin:
Yeah, I think to implement it properly we need a way for Mkdocs plugins to provide templates / partials. I've looked through [Template events](https://www.mkdocs.org/dev-guide/plugins/#template-events) but it seems you can't hook in before templates are parsed from the theme (and `custom_dir`).

Feel free to close this off then! Hopefully it still helps anybody who searches issues for it.

squidfunk:
> Yeah, I think to implement it properly we need a way for Mkdocs plugins to provide templates / partials. I've looked through [Template events](https://www.mkdocs.org/dev-guide/plugins/#template-events) but it seems you can't hook in before templates are parsed from the theme (and custom_dir).

I just verified – it's very simple to bundle extra templates or partials using hooks or plugins. You can just add your directory containing the overridden partials using this insanely simple hook:

``` yaml
def on_config(config):
    config.theme.dirs.insert(0, "extensions")
```

This will add the `extensions` folder before all other partials, so you can also overriden Material's own partials. By wrapping this into a plugin, you can make that installable as a plugin. Note that hooks are as I understand promoted to plugins.

Here's your template in a self-contained example you can run:

[mkdocs-bundle-templates-example.zip](https://github.com/squidfunk/mkdocs-material/files/10390749/mkdocs-bundle-templates-example.zip)

This is actually an excellent use case for theme extension and something we should add to our [examples repository](https://github.com/mkdocs-material/examples), which we're currently working on. In case you bundle this into a plugin, feel free to PR a documentation change so that we can mention it in our documentation. We're always happy to advertise third-party extensions providing extra functionality that we cannot commit to maintain. 🚀

notpushkin:
Thank you for the example! I've ported it to a standalone MkDocs plugin and it does indeed seem to work: https://material-plausible-plugin.ale.sh/

Issue #4854, opened 2023-01-11
Emojis do not work in site lists in navigation


pktiuk:
### Bug description

![obraz](https://user-images.githubusercontent.com/45544416/211923399-a4af09bd-afa5-4401-8c81-c17ad5e6d368.png)

Emoji is shown properly in the title of that page but not in the navigation list.

### Related links

- [Bug reporting](https://squidfunk.github.io/mkdocs-material/bug-report/)
- [Emojis usage](https://squidfunk.github.io/mkdocs-material/reference/icons-emojis/?h=icon)

### Reproduction

Config file:

```yml
site_name: My Docs

theme:
  name: material

markdown_extensions:
  - attr_list
  - pymdownx.emoji:
      emoji_index: !!python/name:materialx.emoji.twemoji
      emoji_generator: !!python/name:materialx.emoji.to_svg
```

docs/index.md

```md
# Welcome to MkDocs :worm:

For full documentation visit [mkdocs.org](https://www.mkdocs.org).

```

### Steps to reproduce

1. Add support to emojis according to docs
2. Use emoji in the header of file `# text :worm:`

### Browser

Firefox

### Before submitting

- [X] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/bug-report/).
- [X] I have attached links to [the documentation](https://squidfunk.github.io/mkdocs-material), and possibly related [issues](https://github.com/squidfunk/mkdocs-material/issues) or [discussions](https://github.com/squidfunk/mkdocs-material/discussions).
- [X] I assure that I have [removed all customizations](https://squidfunk.github.io/mkdocs-material/bug-report/#remove-customizations) before submitting this bug report.
- [X] I have attached a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/bug-report/reproduction/).

facelessuser:
Upgrade pymdown-extensions to 9.9.1

facelessuser:
Never mind, this is a navigation question, this may not be related to the recent Twemoji issue.

squidfunk:
This does not work with MkDocs, but the [new built-in typeset plugin](https://squidfunk.github.io/mkdocs-material/reference/#built-in-typeset-plugin) is exactly what you're looking for, as it allows to use formatted titles in both sidebars, i.e. the navigation and the table of contents.

PR #4857, opened 2023-01-12
Mention typeset plugin in Emoji docs

pktiuk:
Followup to issue: https://github.com/squidfunk/mkdocs-material/issues/4854

I think it would be good to mention this plugin in emoji docs


squidfunk:
Thanks! LGTM!

PR #4864, opened 2023-01-15
Update CHANGELOG

andrewchung:
Modified year designation from 2022 to 2023 on changelog items mkdocs-material-9.0.1 thru to mkdocs-material-9.0.5


squidfunk:
Thanks!

PR #4876, opened 2023-01-16
Fix issue with 'detected dubious ownership' in docker image

timvink:
Seems I need to add `site/` as well for some users. 

See https://github.com/timvink/mkdocs-git-revision-date-localized-plugin/issues/107 for details.


squidfunk:
Thanks!

PR #4890, opened 2023-01-20
Escape parentheses separator in docs

phoracek:
Without escaping, these parentheses are interpreted as regex operators and the whole separator string then does not work as expected.

### DIFF
--- a/docs/setup/setting-up-site-search.md
+++ b/docs/setup/setting-up-site-search.md
@@ -112,7 +112,7 @@ The following configuration options are supported:
     ``` yaml
     plugins:
       - search:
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. However, the assumptions of the PR are incorrect. Parenthesis, i.e., `(` and `)`, albeit being control characters, do not need escaping __in character ranges__. Try the following regular expressions in your browser console:

``` js
/^[()]$/.test("(") // => true
/^[()]$/.test(")") // => true
/^[()]+$/.test(")(") // => true
```

Thus, there's no need for this change.

PR #4912, opened 2023-01-25
Rename browser-support.md to mkdocs-browser-support.md

q1blue:


squidfunk:
This PR makes no sense. This will break all links on the website and has absolutely no upside. Closing.

PR #4917, opened 2023-01-26
Updated Polish translations

kamilkrzyskow:
➕ Added missing entries.
🛠️ Modified `clipboard.copied` to follow same wording as `clipboard.copy`.
🛠️ Modified `source` to follow same wording as `action.skip`.
🔀 Moved `search.config.pipeline` to alphabetical ordering.

~~My only doubt is for `search.share`, as "Podziel się" is in my opinion the best translation here, but maybe the generic "Udostępnij" might be better for a "dry" documentation page 🤔. Most likely doesn't matter 🤷.~~

I've noticed a lack of universal fallback text for the palette toggle name. The default admonition names are not supported for localization. The "Made with Material for MkDocs" notice too, but at least that makes sense, since the documentation is in English. 😉 

I also tried setting `search.config.lang` to `pl`, and this leads to nonfunctional search, as in it's initializing search forever, because the search index contains `"lang":["pl"]` which isn't supported by `lunr`. So maybe a **hidden issue**, because this localization override bypasses the supported languages check and doesn't fall back to `en`? 

Seems like `search.config.pipeline` is an empty space for languages that aren't supported in `lunr`. Supposedly, because it improves search for them. However, the `pipeline` setting isn't documented. In our page, we are currently implementing localization with the `i18n` plugin, with `en` as default. The search index is global for all languages, therefore it's using the default `stopWordFilter`. It would be nice to know which setting would be best for multilingual search results.


kamilkrzyskow:
Ok, after ruminating over it and looking over different Polish websites, I've noticed that "Udostępnij" is more widely used with that sort of action / icon. So even though, I prefer the previous translation, I shall go with a more generic, typical word.

Please squash those commits on merge, no need for them to be separate.

squidfunk:
Thanks for the PR. However, please mark unfinished pull requests as drafts. It's otherwise unclear if this is ready to be merged or not. I have to trust you on translations, since I don't speak Polish.

kamilkrzyskow:
Thanks for pointing that out @squidfunk, I'll keep that in mind next time. Both translations are correct and in online use, so I was confident it's finished. I just started noticing the other translation more and more, so seeing that it's not merged I decided to make the change. 👌 

I haven't created an issue for the behavior of the `search.config.lang` setting, as I'm not sure if it's truly an issue or the intended behavior. So I hope that such remarks, like the one in the PR description, are enough for you to take note of such things 😄

kamilkrzyskow:
Research Done ✅ I won't add more changes to this PR 💯

squidfunk:
Thanks again. Regarding `search.config.*`: these settings are highly relevant to make search work for languages that Lunr.js doesn't have out-of-the-box support for. Thus, they must not be changed. This is also [mentioned in our contributing guide](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md#missing-translations):

> The following translations are for technical purposes and should not be updated, so if they're missing from your language, it's for good reason:
> 
> - search.config.lang
> - search.config.pipeline
> - search.config.separator

On a side note: we're currently reworking our entire translation contribution process, making it much easier to contribute and update translations in the future 😊

kamilkrzyskow:
Thanks for addressing my concern. I've read that, I knew I can omit untranslated entries from that contribution guide, but I completely didn't pay attention to those explicitly listed settings. 😅

PR #4925, opened 2023-01-27
Update pt-BR.html

gavlinski:
Review of pt-BR translation


kamilkrzyskow:
Hi,
you've edited the `material/partials/languages/pt-BR.html` file, instead of the file in the `src` path tree.
More explained here https://squidfunk.github.io/mkdocs-material/customization/#development-mode

squidfunk:
Thanks for the PR! @kamilkrzyskow is right, but I went ahead and handled everything in e9274f61e 😊

kamilkrzyskow:
Wait a second @squidfunk , I've looked over the commits, and I've noticed that you didn't attribute him as the change author in the `src` file commit 🤔. I know it's only a few lines and not a big deal, I just believe that every single line should have a correct git blame history 😸

Issue #4934, opened 2023-01-28
Update Polish translations


balansczerni:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### Language availability

- [X] I've checked the [list of available languages](https://bit.ly/33vFDD0)

### Translations

```Jinja
{% macro t(key) %}{{ {
  "language": "pl",
  "action.edit": "Edytuj tę stronę",
  "action.skip": "Przejdź do treści",
  "action.view": "Zobacz kod źródłowy tej strony",
  "announce.dismiss": "Nie pokazuj tego ponownie",
  "blog.archive": "Archiwum",
  "blog.categories": "Kategorie",
  "blog.categories.in": "",
  "blog.continue": "Czytaj dalej",
  "blog.draft": "Wersja robocza",
  "blog.index": "Powrót do indeksu",
  "blog.meta": "Metadane",
  "blog.references": "Powiązane łącza",
  "clipboard.copy": "Kopiuj do schowka",
  "clipboard.copied": "Skopiowano do schowka",
  "consent.accept": "Akceptuj",
  "consent.manage": "Zarządzaj ustawieniami",
  "consent.reject": "Odrzuć",
  "footer": "Stopka",
  "footer.next": "Następna strona",
  "footer.previous": "Poprzednia strona",
  "header": "Nagłówek",
  "meta.comments": "Komentarze",
  "meta.source": "Kod źródłowy",
  "nav": "Nawigacja",
  "readtime.one": "Czas czytania: 1 min",
  "readtime.other": "Czas czytania: # min",
  "rss.created": "Kanał RSS",
  "rss.updated": "Kanał RSS zaktualizowanych treści",
  "search": "Szukaj",
  "search.placeholder": "Szukaj",
  "search.share": "Udostępnij",
  "search.reset": "Wyczyść",
  "search.result.initializer": "Inicjowanie wyszukiwania",
  "search.result.placeholder": "Zacznij pisać, aby szukać",
  "search.result.none": "Brak wyników wyszukiwania",
  "search.result.one": "Wyniki wyszukiwania: 1",
  "search.result.other": "Wyniki wyszukiwania: #",
  "search.result.more.one": "1 więcej na tej stronie",
  "search.result.more.other": "# więcej na tej stronie",
  "search.result.term.missing": "Brak",
  "select.language": "Wybierz język",
[...]

PR #4957, opened 2023-01-30
Updated Polish translations

kamilkrzyskow:
~~Ready for merge~~ Turned out it wasn't...


src/partials/languages/pl.html
```diff
@@ -39,8 +47,11 @@
   "meta.comments": "Komentarze",
   "meta.source": "Kod źródłowy",
   "nav": "Nawigacja",
+  "readtime.one": "Czas czytania: 1 min",
+  "readtime.other": "Czas czytania: # min",
+  "rss.created": "Kanał RSS",
+  "rss.updated": "Kanał RSS zaktualizowanych treści",
   "search": "Szukaj",
-  "search.config.pipeline": " ",
```
squidfunk, line 43: Please don't remove this setting. This is exactly what I referred to in https://github.com/squidfunk/mkdocs-material/issues/4934#issuecomment-1408146570

squidfunk (changes requested):

src/partials/languages/pl.html
kamilkrzyskow: Oh, this is the template generated from the first day iirc, The template generated the Issue without the pipeline and I just copy pasted it over. I'm sorry for not understanding previously.

squidfunk: Yes, the key is not included in the template because users always changed it to their language which makes no sense and will break search.

Issue #4919, opened 2023-01-26
Add support for Semantic Versioning to the Version Warning 


SaltyAimbOtter:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I want to suggest an idea and checked that ...

- [X] ... to my best knowledge, my idea wouldn't break something for other users
- [X] ... the documentation does not mention anything about my idea
- [X] ... there are no open or closed issues that are related to my idea

### Description

The version warning should be able to correctly inform the user about outdated versions if the [semantic versioning](https://semver.org/) version format is used.

The current behaviour is like this:
Green = Correctly shows no warning (latest)
Red = Shows no warning although it should
Light Green = Correctly shows a warning 
![grafik](https://user-images.githubusercontent.com/40303883/214889540-d38a69c1-75e6-485d-905d-7cc7b1158674.png)

Mike does claim that it is made for two digit version only. However, it also claims that it is "flexible". In our project, we never had any problems using semantic versioning with mike. I'm sure it's a common use case considering that semantic versioning is quite popular.

In addition to that a dedicated banner for development versions would be useful:
We have one RELEASE (1.2.3) doc version and one DEV doc version (e.g. 2.0.0-DEV). Additionally there are many old docs versions (e.g. 1.2.2, 1.2.1 etc.). 
We would like to show one variant of the warning banner in the docs of legacy versions and another in the development docs.
This could be attached to aliases:
| Mike Alias | Banner|
|--|--|
|  latest | No Banner  |
| development | Banner Variant 1 |
| none | Banner Variant 2 |

Where Banner Variant 1 would say something like: `You are viewing the docs for the development version! Be careful or use the latest release!`
And Banner Variant 2 would say something like: `You are viewing outdated docs and will recieve no support! Update to the latest release!`

The amount of version qualifiers might differ between projects though. So this should be customizable. I'm not sure if and how this could be implemented considering the current implementation works with theme overrides.

### Use Cases

Many users probably release one documentation version together with a software release. Since semantic versioning is very popular this should affect a large part of the user base.

squidfunk:
[...]

PR #4976, opened 2023-02-01
add support for multiple default aliases in version warning

SaltyAimbOtter:
This PR adds the ability to define not only one but multiple default version aliases that hide the version warning.

## Docs Changes Preview
grafik

## Context
This is related to #4919: After opening the issue, our project members had an internal discussion. I will post our conclusions in the issue shortly after opening this PR.

Please excuse any mistakes regarding the tooling, I'm new to web development.


squidfunk:
Thanks for the PR! We should allow both syntaxes, a string value and a list to allow, as it's a much more natural syntax to define it as a list in YAML. Judging from the [current implementation](https://github.com/squidfunk/mkdocs-material/blob/e496f77b6ce5febc448345cb8a0d7af594627300/src/base.html#L386), this should work out-of-the-box from a template perspective, and only needs adjustment in the JavaScript. We could then do this:

``` yaml
extra:
  version:
    default: stable
```

... as well as:

``` yaml
extra:
  version:
    default:
      - stable
      - development
```

What needs to be adjusted is the logic here – we now need to handle strings as well as arrays of strings:
https://github.com/squidfunk/mkdocs-material/pull/4976/files#diff-dfd2137f020c891f8ca6ca734202c930b2008ace56b84bc0cd6f61ed7c33c0dcL162

Also note that your PR contains too many files. You somehow committed the files that are generated during development, as well as the files that are generated by the build, e.g. `search.js` and `search.********.js`. Pleae make sure to exclude the former from the PR.

---

If you feel that this is too much to handle, I can take care of it when I find some time 😊

SaltyAimbOtter:
I'm totally lost 🙈 
I managed to develop the change but I have no idea how to build the theme the right way. Therefore I have stripped this branch of all the dev stuff and just included the changes. Please checkout the branch and do whatever magic is required 🧙

SaltyAimbOtter:
Hey Martin,
I think something with your build tools is broken. I have asked a few friends with more experience with the tools at hand but none of them could get it to work. 

When running `npm run build`, the css files `main`, `palette`, `custom` all contain absolute paths instead of relative ones like before the build:
`src/.overrides/assets/stylesheets/custom/_typeset.scss` becomes:
`file:///C:/Users/CyberOtter/IdeaProjects/mkdocs-material/src/.overrides/assets/stylesheets/custom/_typeset.scss`.

Other than that it seems do everything right, e.g. replacing / renaming `custom.acd3711.min.js` with `custom.317484c8.min.js` based on the new hash of the file.

I have tried building with IntelliJ's and VSCode's terminals and from both Windows and WSL.
node: `v18.12.0`
npm: `9.4.2`

squidfunk:
Thanks for noting. I don't have a Windows machine at my disposal, so any help on this is appreciated. We need to make sure that it runs on Linux, macOS and Windows, but I can only test for the two former.

SaltyAimbOtter:
@seyfahni build this on Linux for me. It should be good to merge now.

 It seems like `build:all` is required for a PR? If we run just build too many files get deleted. Perhaps you should add some "participating" documentation for contributors like you did for issues.

squidfunk:
Please give me some more time to look into this.

squidfunk:
So, thanks again for your work on this! I've taken your work and refactored it a little. Additionally, I've added the ability to provide regular expressions as default versions. This should allow for even greater flexibility, as you could now tag version aliases with suffixes or prefixes to identify stable versions:

``` yaml
extra:
  version:
    provider: mike
    default:
      - '-dev$' 
      - ...
```

Regular expression support is highly experimental, which is why it's not yet documented. Let's first give it a try and see if we can consider it useful enough. Superseded by b25723571. I've added you as a co-author.

PR #4984, opened 2023-02-03
Fix mermaid Entity-Relationship-Diagram attribute colors

### LINKED ISSUE #4975: Minor adjustment in mermaid entity relationship diagram colors to improve readability in dark mode
### Context

_No response_

### Description

Currently, mermaid entity relationship diagrams look good in dark mode, if the entities do not contain any attribute definitions, as in the example [here](https://squidfunk.github.io/mkdocs-material/reference/diagrams/?h=diagrams#using-entity-relationship-diagrams).

When defining attributes, it still looks readable in light mode, but in dark mode it is not readable at all. See the visuals for illustration.

As it turns out, to change the CSS behavior of mermaid diagrams like this, a custom theme needs to be created, which is not as convenient as most users would like to have it.

I found that the issue at hand can be found here: https://github.com/squidfunk/mkdocs-material/blob/e496f77b6ce5febc448345cb8a0d7af594627300/src/assets/stylesheets/main/integrations/_mermaid.scss#L33  
Here, a foreground color is applied to a background item, which is not surprising to cause some issues.  
Changing the line to `--md-mermaid-node-bg-color: var(--md-default-bg-color);` changed the coloring to be in line with how other mermaid elements are rendered, and is much more readable (again, see visuals).

Arguably, a solution with two shades of color alternating would correspond better with the official mermaid rendering, but I think even the simple, proposed solution would be an improvement.

If this change is approved, I'd be happy to open a PR from my fork, or see this incorporated in any other way.

### Related links

- [ER Diagram example in docs](https://squidfunk.github.io/mkdocs-material/reference/diagrams/?h=diagrams#using-entity-relationship-diagrams)
- [modifying CSS in mermaid is limited](https://github.com/squidfunk/mkdocs-material/discussions/4582)

### Use Cases

This wouldn't introduce drastic change, some ER diagrams would just suddenly be more readable in dark mode.

## status quo
currently, in light mode, all looks ok:
status quo light mode - no problem

but in dark mode it's not really readable
problem

## proposed change
proposed solution

### Before submitting

- [X] I have ensured that, to the best knowledge, [my idea will benefit the entire community](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/#benefit-for-the-community).
- [X] I have included relevant links to [the documentation](https://squidfunk.github.io/mkdocs-material), related [issues](https://github.com/squidfunk/mkdocs-material/issues) and [discussions](https://github.com/squidfunk/mkdocs-material/discussions) to underline the need for my idea.

BalduinLandolt:
Resolves #4975

I went through all diagram types covered with examples - there was nothing else significant missing, so nothing more to fix than the issue I reported initially.

### DIFF
--- a/src/assets/stylesheets/main/integrations/_mermaid.scss
+++ b/src/assets/stylesheets/main/integrations/_mermaid.scss
@@ -30,7 +30,7 @@
 
   // Colors
   --md-mermaid-edge-color:             var(--md-code-fg-color);
-  --md-mermaid-node-bg-color:          var(--md-accent-fg-color--transparent);
+  --md-mermaid-node-bg-color:          var(--md-default-bg-color);
   --md-mermaid-node-fg-color:          var(--md-accent-fg-color);
   --md-mermaid-label-bg-color:         var(--md-default-bg-color);
   --md-mermaid-label-fg-color:         var(--md-code-fg-color);
--- a/docs/reference/diagrams.md
+++ b/docs/reference/diagrams.md
@@ -246,6 +246,10 @@ a specific domain of knowledge:
 erDiagram
   CUSTOMER ||--o{ ORDER : places
   ORDER ||--|{ LINE-ITEM : contains
+  LINE-ITEM {
+    string name
+    int pricePerUnit
+  }
   CUSTOMER }|..|{ DELIVERY-ADDRESS : uses
 ```
 ````
@@ -256,6 +260,10 @@ erDiagram
 erDiagram
   CUSTOMER ||--o{ ORDER : places
   ORDER ||--|{ LINE-ITEM : contains
+  LINE-ITEM {
+    string name
+    int pricePerUnit
+  }
   CUSTOMER }|..|{ DELIVERY-ADDRESS : uses
 ```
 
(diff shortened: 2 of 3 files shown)


src/assets/stylesheets/main/integrations/_mermaid.scss
```diff
@@ -30,7 +30,7 @@
 
   // Colors
   --md-mermaid-edge-color:             var(--md-code-fg-color);
-  --md-mermaid-node-bg-color:          var(--md-accent-fg-color--transparent);
```
squidfunk, line 33: Okay, so this is actually a pretty fundamental change:

__Before__:

Bildschirmfoto 2023-02-07 um 16 44 23

__Now__:

Bildschirmfoto 2023-02-07 um 16 44 54

I'm afraid I don't like the changes, and expect many users to complain since they're quite fundamental. Could you please check if there's another solution? Unfortunately, this solution can't be merged.

squidfunk:
Superseded by f7598410c. The commit fixes the issue without changing all styles:

Bildschirmfoto 2023-02-07 um 18 42 58

Issue #5029, opened 2023-02-11
Inline Admonition docs don't indicate where the title text should go


tedyoung:
### Description

The examples don't show how to use `inline` (or `inline end`) when there's a title. I thought it would go right after the admonition type, but it actually goes at the end, after any `inline [end]` modifiers.

### Related links

- [Admonitions: Inline blocks](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#inline-blocks)

### Proposed change

Either add additional examples showing the correct syntax, or modify existing examples to use a title.

For example, replace the existing example (and similarly the `inline`-only version) with ones that have titles.

```
!!! info inline end "Title Text"

    Lorem ipsum dolor sit amet, consectetur
    adipiscing elit. Nulla et euismod nulla.
    Curabitur feugiat, tortor non consequat
    finibus, justo purus auctor massa, nec
    semper lorem quam in massa.
```

### Before submitting

- [X] I have read and followed the [documentation issue reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-docs-issue/).
- [X] I have attached the links to the described sections of [the documentation](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-docs-issue/#related-links)

squidfunk:
Fixed in https://github.com/squidfunk/mkdocs-material/commit/04ece016c1113a0489979f10ba59532e20e2cfb1, see https://github.com/squidfunk/mkdocs-material/pull/5030#issuecomment-1426984087.

PR #5030, opened 2023-02-11
Add title text for inline examples

tedyoung:
As per issue #5029.

Also minor grammar fix to replace i.e. with e.g. (for example) and a comma after.


squidfunk:
Thanks for the PR! However, I think it's better to use "Lorem ipsum" here, so that it's clear that the text between quotes is not Markdown syntax but can be freely configured, which I just did in 04ece016c. IMHO this is now pretty apparent.

I've also included your grammar changes in e5c751883.

PR #5055, opened 2023-02-16
Fix cache: pip Example

RammusXu:
It shows this error

example: https://github.com/RammusXu/toolkit/blob/69612a431d5f0aa7ae85a7ed7d66850e67494a7e/.github/workflows/publish.yaml
```
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/cache@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.Show more
```

image

After this change, it will fix.

example: https://github.com/RammusXu/toolkit/actions/runs/4193199747

ref: https://github.com/actions/setup-python#caching-packages-dependencies

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -37,11 +37,8 @@ contents:
           - uses: actions/checkout@v3
           - uses: actions/setup-python@v4
             with:
-              python-version: 3.x
-          - uses: actions/cache@v2
-            with:
-              key: ${{ github.ref }}
-              path: .cache
+              python-version: 3.9
+              cache: 'pip' # caching pip dependencies
           - run: pip install mkdocs-material # (3)!
           - run: mkdocs gh-deploy --force
     ```


docs/publishing-your-site.md
```diff
@@ -37,11 +37,8 @@ contents:
           - uses: actions/checkout@v3
           - uses: actions/setup-python@v4
             with:
-              python-version: 3.x
```
squidfunk, line 40: Thank for the PR. However, you're confusing `pip` caching with asset caching. See [our documentation](https://squidfunk.github.io/mkdocs-material/setup/setting-up-social-cards/#caching). Please make sure that the `.cache` folder is still cached – it's not about `pip` caching.

squidfunk (changes requested):

PR #5065, opened 2023-02-18
Remove polyfills for ancient browsers from bundle

squidfunk:
Most polyfills are not necessary in todays browsers – they are only necessary for ancient browsers on systems that cannot be updated. This PR removes all unnecessary polyfills from Material for MkDocs and introduces a new configuration setting, which can be used to add the polyfills back that are needed by the clients visiting the respective site:

``` yaml
config:
  extra:
    polyfills:
      - https://unpkg.com/resize-observer-polyfill
      - ...
```

As a result, the shipped JavaScript is now significantly lighter:

- `.min`:  114kb → 94kb (~20% less)
- `.min.gz`: 36kb →29kb (~20% less)

This results in faster load times and better lighthouse scores. Of course, this setting also integrates with the [built-in privacy plugin](https://squidfunk.github.io/mkdocs-material/setup/ensuring-data-privacy/#built-in-privacy-plugin), which makes self-hosting polyfills trivial. As an alternative, polyfills can also be downloaded and relatively linked.


squidfunk:
Superseded by #5684.

Issue #5080, opened 2023-02-21
Extend the Data Tables page to include info on how to load data from static file


chrimaho:
### Description

Can we please extend the [Data Tables][data-tables] page to include information on how to load data from a static file? Such as a CSV or Excel file?

This will not require extending the `mkdocs-material` package to include this funtionality, as the [`mkdocs-table-reader-plugin`][table-reader-plugin] currently fulfils this need.

This request is just to extend the docs page to show how this may be achieved. In the same way that the [Images][images] page shows how to use the [glightbox][glightbox] plugin for extended functionality.

[table-reader-plugin]: https://timvink.github.io/mkdocs-table-reader-plugin/
[data-tables]: https://squidfunk.github.io/mkdocs-material/reference/data-tables/
[images]: https://squidfunk.github.io/mkdocs-material/reference/images/
[glightbox]: https://github.com/blueswen/mkdocs-glightbox

### Related links

- Docs page to be extended: [Data Tables](https://squidfunk.github.io/mkdocs-material/reference/data-tables/)
- Discussion page for background context: https://github.com/squidfunk/mkdocs-material/discussions/3315#discussioncomment-5066920
- Docs page for the suggested plugin: [mkdocs-table-reader-plugin](https://timvink.github.io/mkdocs-table-reader-plugin/)

### Proposed change

 I've drafted the change. I will submit a Pull Request for you to review.

### Before submitting

- [X] I have read and followed the [documentation issue reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-docs-issue/).
- [X] I have attached the links to the described sections of [the documentation](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-docs-issue/#related-links)

squidfunk:
Resolved by #5081.

PR #5081, opened 2023-02-21
Extend the Data Tables page to include info on how to load data from static file

chrimaho:
Closes:
- #5080

As discussed here:
- #3315

We can use the plugin [`mkdocs-table-reader-plugin`][table-reader-plugin] to facilitate loading data from a static file.

You'll notice there that in the 'Results' sections, I have manually added the tables as Markdown format... This is because in order to actually render these files using the plugin, it will mean adding new packages to the `requirements.txt` file, which I think is not desirable for extending docs pages. So can you please adjust this on your side to install the new plugin and render it correctly on the live web page?

Thank you.

[table-reader-plugin]: https://timvink.github.io/mkdocs-table-reader-plugin/


docs/reference/data-tables.md
```diff
@@ -182,3 +182,145 @@ numbers, filesizes, dates and month names. See the [tablesort documentation]
...
+
+        And it will then return a result like this:
+
+        col1|col2|col3
+        ----|----|----
+        r1c1|r1c2|r1c3
+        r2c1|r2c2|r2c3
+        r3c1|r3c2|r3c3
+
+=== "Excel"
```
squidfunk, line 242: Please don't use double nested tabs. That's an anti-pattern. Yes, we also use it in 1-2 sections, but I think this section could be easily split in two sections with headlines + one tabbed container.

docs/reference/data-tables.md
```diff
@@ -182,3 +182,145 @@ numbers, filesizes, dates and month names. See the [tablesort documentation]
...
+
+    For this example, let's use an Excel file in the local directory.
+
+    === "Data"
```
squidfunk, line 246: Please use file icons and filenames here, so it's clear that this refers to a file. See how we use this pattern here: https://squidfunk.github.io/mkdocs-material/reference/tooltips/#adding-a-glossary

squidfunk (changes requested):
Thanks for the PR! I think this is a great addition to our documentation. I only have some minor change requests regarding the structure 😊

chrimaho:
Thanks @squidfunk! I've now updated it. Please check 👍

docs/reference/data-tables.md
```diff
@@ -182,3 +182,154 @@ numbers, filesizes, dates and month names. See the [tablesort documentation]
...
+
+    ```md title="./markdown.md"
+    ...
+    {{ read_csv('./data.csv') }}
+    ...
+    ```
+
+And it will then return a result like this:
+
+{{ read_csv('./data.csv') }}
```
squidfunk, line 234: You didn't add the plugin to `mkdocs.yml`, so this will be printed as-is (= no table). However, I would prefer if we don't add it to `mkdocs.yml`, but it means we need to include the plugin output manually, so please do so. We do the same at different positions of our documentation, because sometimes options are even conflicting, so we have to "fake" it. Additionally, we need to try and keep the dependencies minimal.

Additionally, please remove both data files.

squidfunk (changes requested):
Thanks again! Some additional minor changes – we need to "fake" the output. See comment.

docs/reference/data-tables.md
squidfunk: When including the plugin output, please use Markdown table syntax.

chrimaho:
Thanks @squidfunk! Great feedback. I've now updated it 👍 Please check.

squidfunk:
Thanks!

squidfunk (approved):

PR #5085, opened 2023-02-23
Use edit in edit_uri example

mark-heptinstall:
To match the behaviour when the default branch is not `master` (e.g. main), the example `edit_uri` should start with `edit` instead of `blob`


squidfunk:
Thanks! Indeed, a good idea.

PR #5088, opened 2023-02-24
ci: merge publish workflow with build one

crazy-max:
follow-up https://github.com/squidfunk/mkdocs-material/pull/4994

Merge publish workflow with build one so we can build python package and docker image on pull request to make sure there is no regression.

Also use [`docker/metadata-action`](https://github.com/docker/metadata-action) to generate tags and labels and don't set latest tag on prerelease.

In a follow-up I will work on making the Docker image multi-platform without build time penalty.


squidfunk:
Thank you! I need a little time to review this PR as we currently need to handle the situation with GitHub Sponsors.

.github/workflows/build.yml
```diff
@@ -61,3 +66,89 @@ jobs:
...
+          labels: ${{ steps.meta.outputs.labels }}
+
+      - name: Check Docker image
+        working-directory: /tmp
+        run: |
+          docker run --rm -i -v ${PWD}:/docs ${{ github.event.repository.full_name }}:${{ steps.meta.outputs.version }} new .
+          docker run --rm -i -v ${PWD}:/docs ${{ github.event.repository.full_name }}:${{ steps.meta.outputs.version }} build
+
+      - name: Publish Docker image
+        uses: docker/build-push-action@v3
```
squidfunk, line 149: Can we upgrade this to v4? That was my initial question in #4994 😅 Otherwise dependabot will complain.

crazy-max: Oh I forgot about this, let me make the change.

crazy-max: Done, tested here: https://github.com/crazy-max/mkdocs-material/actions/runs/4274609950/jobs/7441347841

squidfunk: Perfect! I will merge this now and then we can test-drive it pushing out 9.0.15. Really looking forward to arm64 support!

squidfunk (approved):
LGTM!

.github/workflows/build.yml
crazy-max: Looks good: https://github.com/squidfunk/mkdocs-material/actions/runs/4274651507/jobs/7441421218#step:9:218

squidfunk:
All good:
https://github.com/squidfunk/mkdocs-material/actions/runs/4274651507

Now the only thing that is needed would be to build images for other architectures on a release event. I guess we could make the architectures to build for configurable via environment variable or something similar. Building for `arm64` so support M1 and M2 Macs should be sufficient, I don't think we need `arm/v7`, or do we?

crazy-max:
> I don't think we need `arm/v7`, or do we?

I don't know many people developing on this platform but doesn't hurt to have it too.

> Prior attempt:
> 
>     *

It needs more changes in the Dockerfile than adding extra platforms. I will carry that. Keep you posted.

squidfunk:
Aight, thanks for your work and time invested!

PR #5102, opened 2023-02-26
ci: build multi-platform image

crazy-max:
follow-up https://github.com/squidfunk/mkdocs-material/pull/5088#issuecomment-1445331154
closes #4240

Builds additional platforms for the Docker image (`linux/arm64` and `linux/armv7`) on release only using emulation to start with.

Build takes approximately 15 min for a fresh build which is not that bad if you're ok with it: https://github.com/crazy-max/mkdocs-material/actions/runs/4274790257/jobs/7441657533#step:10:143

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -99,6 +99,9 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
 
@@ -145,10 +148,16 @@ jobs:
           docker run --rm -i -v ${PWD}:/docs ${{ github.event.repository.full_name }}:${{ steps.meta.outputs.version }} new .
           docker run --rm -i -v ${PWD}:/docs ${{ github.event.repository.full_name }}:${{ steps.meta.outputs.version }} build
 
+      - name: Set platforms
+        if: github.event_name == 'release'
+        run: |
+          echo "PLATFORMS=linux/amd64,linux/arm64,linux/arm/v7" >> $GITHUB_ENV
+
       - name: Publish Docker image
         uses: docker/build-push-action@v4
         with:
           context: .
+          platforms: ${{ env.PLATFORMS }}
           push: ${{ github.event_name == 'release' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -161,3 +161,14 @@ jobs:
           push: ${{ github.event_name == 'release' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+
+      - name: Check manifest
+        if: github.event_name == 'release'
+        run: |
+          docker buildx imagetools inspect ${{ github.event.repository.full_name }}:${{ steps.meta.outputs.version }}
+
+      - name: Inspect image
+        if: github.event_name == 'release'
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -149,6 +154,20 @@ jobs:
         uses: docker/build-push-action@v4
         with:
           context: .
+          platforms: linux/amd64,linux/arm64,linux/arm/v7
           push: ${{ github.event_name == 'release' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha,scope=mkdocs-docker-image
+          cache-to: type=gha,scope=mkdocs-docker-image,mode=max
```
crazy-max, line 162: Also adds cache support for Docker build which should reduce build time for incremental builds but we need to keep `__pycache__` and `/root/.cache` and rework a bit the Dockerfile.

crazy-max:
@squidfunk This is just to start with. Best would be of course to support cross compilation for python but it needs more work to handle this. Maybe through https://pypi.org/project/crossenv/.

.github/workflows/build.yml
crazy-max: Added extra commit to use [multi-stage builds](https://docs.docker.com/build/building/multi-stage/) in the Dockerfile so cache is more efficient and we don't need to remove temp folders anymore. It also reduces final image size and layers.

squidfunk:
Just ping me when you think this is ready! Looks great so far from what I can tell. 15min is okay if we can limit that to release builds, having a fast path that only checks that the Docker image build for `x86_64` on non-release builds. I saw that you also started to rework the Docker image to use a multi-stage build. That's awesome, wanted to do this but haven't found the time to wrap my head around it.

crazy-max:
> Just ping me when you think this is ready! Looks great so far from what I can tell. 15min is okay if we can limit that to release builds, having a fast path that only checks that the Docker image build for `x86_64` on non-release builds. I saw that you also started to rework the Docker image to use a multi-stage build. That's awesome, wanted to do this but haven't found the time to wrap my head around it.

Should be good for review. I removed the multi-stage commit for now. Will do that as follow-up along other changes in the Dockerfile for cross compilation support so we can build a multi-platform image on pull request too without build time penalty. Will also look at a better cache support as follow-up as well.

squidfunk:
Okay, thanks! We'll test it on the next release. Thanks again for your work!

squidfunk:
Just triggered the release of 9.1.0 – let's see how that goes!
https://github.com/squidfunk/mkdocs-material/actions/runs/4312089371

squidfunk:
Success! It took 25m, but all good. Could we leverage caching? The dependencies installed through `apk` rarely change – most changes are scoped to themes files, so I expect that this could speed up the build big time 😀

Thanks again for this!

squidfunk:
cc @afritzler

afritzler:
Those are great news! I guess I can archive my multi-arch project and use the upstream images from now on. 👌🏻

crazy-max:
> Success! It took 25m, but all good. Could we leverage caching? The dependencies installed through `apk` rarely change – most changes are scoped to themes files, so I expect that this could speed up the build big time 😀
> 
> Thanks again for this!

Yes we can improve this with multi-stage. Will open a follow-up with this and add some cache too.

squidfunk:
@crazy-max sorry to necro-post, but the Docker build seems broken and I'm a bit lost what the problem is. Could you maybe take a look? It's really meh, because it's the 9.2 release that so many people awaited and I'm not sure what I can do now to publish it: https://github.com/squidfunk/mkdocs-material/actions/runs/5929515795/job/16078210069

squidfunk:
Nevermind. Stupid me, I forgot to update the `Dockerfile`, `libxml` and `libxslt` were missing. Sorry for the noise!

PR #5122, opened 2023-02-28
chore(stylelint): replace unmaintained stylelint-config-rational-order with stylelint-config-recess-order

pereorga:
[stylelint-config-rational-order](https://github.com/constverum/stylelint-config-rational-order) has been unmaintained since 2019. It is not being advertised in https://github.com/hudochenkov/stylelint-order anymore.

This PR is to replace it with [stylelint-config-recess-order](https://github.com/stormwarning/stylelint-config-recess-order), which is maintained and used by [more projects](https://npmtrends.com/stylelint-config-rational-order-vs-stylelint-config-recess-order) (Bootstrap is one of them).

This PR is a bit opinionated. With this new package the order of the properties changes a bit, and in my opinion the result is generally better (transition/transform/animation properties are put together, font-family goes before font-size, font-size before font-weight, background is after color, margin after padding, etc.).

Also, as a result of this, `package-lock.json` file is reduced considerably, fewer dev dependencies are now required.

This PR also fixes all files.

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,6 +1,6 @@
 {
   "extends": [
-    "stylelint-config-rational-order",
+    "stylelint-config-recess-order",
     "stylelint-config-recommended",
     "stylelint-config-standard-scss"
   ],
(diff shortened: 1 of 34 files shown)


squidfunk:
Thanks for the PR! I was searching for a replacement for the plugin, so your work on this is greatly appreciated! Phew, quite a few changes. I've eye-balled all of them and think this is safe to merge. Let's hope we didn't forget something 😀

Issue #5191, opened 2023-03-13
Update Portuguese translations


DavidFreire-FEUP:
### Translations

```Jinja
{% macro t(key) %}{{ {
  "language": "pt",
  "action.edit": "Editar esta página",
  "action.skip": "Ir para o conteúdo",
  "action.view": "Ver fonte desta página",
  "announce.dismiss": "Não mostrar novamente",
  "blog.archive": "Arquivo",
  "blog.categories": "Categorias",
  "blog.categories.in": "em",
  "blog.continue": "Continuar leitura",
  "blog.draft": "Rascunho",
  "blog.index": "Voltar ao índice",
  "blog.meta": "Metadados",
  "blog.references": "Ligações relacionadas",
  "clipboard.copy": "Copiar para área de transferência",
  "clipboard.copied": "Copiado para área de transferência",
  "consent.accept": "Aceitar",
  "consent.manage": "Gerir configurações",
  "consent.reject": "Rejeitar",
  "footer": "Rodapé",
  "footer.next": "Próximo",
  "footer.previous": "Anterior",
  "header": "Cabeçalho",
  "meta.comments": "Comentários",
  "meta.source": "Fonte",
  "nav": "Navegação",
  "readtime.one": "1 min de leitura",
  "readtime.other": "# min de leitura",
  "rss.created": "canal RSS",
  "rss.updated": "canal RSS com conteúdo atualizado",
  "search": "Pesquisar",
  "search.placeholder": "Buscar",
  "search.share": "Compartilhar",
  "search.reset": "Limpar",
  "search.result.initializer": "Inicializando a pesquisa",
  "search.result.placeholder": "Digite para iniciar a busca",
  "search.result.none": "Nenhum resultado encontrado",
  "search.result.one": "1 resultado encontrado",
  "search.result.other": "# resultados encontrados",
  "search.result.more.one": "Mais 1 nesta página",
  "search.result.more.other": "Mais # nesta página",
  "search.result.term.missing": "Ausente",
  "select.language": "Selecione o idioma",
  "select.version": "Selecione a versão",
  "source": "Ir ao repositório",
  "source.file.contributors": "Colaboradores",
  "source.file.date.created": "Criada",
  "source.file.date.updated": "Última atualização",
  "tabs": "Abas",
  "toc": "Índice",
  "top": "Voltar ao topo"
}[key] }}{% endmacro %}
```

### Before submitting

[...]

PR #5192, opened 2023-03-13
Portuguese translations

DavidFreire-FEUP:
Added all missing Portuguese (European) translations.

I accidentally didn't build overrides in the first commit. Although I did for the second, there is still one missing file (iconsearch_index.json), is it supposed to?

PR resolves https://github.com/squidfunk/mkdocs-material/issues/5191


squidfunk:
Thanks for the PR! You likely didn't install the Python dependencies with `pip install -e .`, which is why this file is missing, but I think we also didn't manage it in our translation guide. We're currently revamping the documentation and add this info. Could you just exclude the deleted file from the build, so that only the files that have additions are included?

DavidFreire-FEUP:
Well, this is weird, because I ran the commands specified in the docs before the build (including the `pip install -e .`)
I ran all these in this specific order:

```
pip install -e .
pip install mkdocs-minify-plugin
pip install mkdocs-redirects
npm install
```
Anyway, just commited with the requested change, sorry for the fail! And thanks for the fast feedback + maintaining mkdocs-material :) .

squidfunk:
Thanks! LGTM!

PR #5206, opened 2023-03-16
Set link color in dark mode to indigo by default

ile-24556:
When using dark mode without setting the primary color, the link color becomes too dark (screenshot below).

image

This is because the default link color for light mode is used.

To avoid this, set `--md-typeset-a-color` to lighter (for-dark-mode) indigo when `data-md-color-scheme` is `slate` and `data-md-color-primary` is empty.


squidfunk:
Thanks for the PR. However, 22 files changed for such a simple change doesn't look right. Please check and remove all unnecessary changes that are not necessary for the proposed change.

ile-24556:
Thank you for your quick review. I had added all the npm-built files, sorry for bothering you.

I reverted the first commit and added the necessary changes. If you prefer a clean history, I would squash these commits and force push.

squidfunk:
Thanks for the PR! However, I think it's a better idea if we automatically set `indigo` as a color when none is provided, effectively removing the "empty" value which you fixed in this PR. Fixed in 6fe1d5560.

squidfunk:
As an addendum: the only downside of this approach is that you now need to be explicit when using multiple color palettes:

``` yaml
theme:
  palette: 

    # Palette toggle for light mode
    - scheme: default
      primary: red
      accent: red

    # Palette toggle for dark mode
    - scheme: slate
      # The following settings are now the default, but were priorly inherited from the default scheme, i.e., red
      # primary: indigo
      # accent: indigo
```

IMHO being more explicit is better than too much magic.

PR #5216, opened 2023-03-19
Add forkawesome/brands/matrix.svg

AnthonyAkentiev:
Matrix is a federated messenger. 
It would be great to have its icon. I need to use it in the footer near other social account icons.


AnthonyAkentiev:
The icon should be inverted in color, sorry

PR #5218, opened 2023-03-19
Add forkawesome/brands/matrix.svg

AnthonyAkentiev:
Matrix is a federated messenger.
It would be great to have its icon. I need to use it in the footer near other social account icons.

p.s. I've inverted the icon as required


squidfunk:
Thanks for the PR. There's no need to add this icon, as we also ship Simple Icons which include thousands of brand logos, including matrix, Just use:

```
:simple-matrix:
```

PR #5244, opened 2023-03-22
Fix examples to use correct HTML <meta> syntax

bannmann:
The docs mention `<meta>` tags for robots, but the examples mistakenly use the attribute name `property` instead of the correct `name`.


squidfunk:
Woopsie, you're absolutely right!

PR #5268, opened 2023-03-27
Clarify where to add `extra_css`

Shrinks99:
I wrongly assumed after reading the documentation that the `extra_css` value would be a part of the material theme settings.  This is of course incorrect, and having it clarified in the docs would have saved me a solid 10 minutes of being annoyed! :)

Arguably this is on me, after figuring this out I do understand that if it were the way I initially thought it would probably be described as:
```yaml
theme:
  extra_css:
    - stylesheets/extra.css
```

...Which it's not! 🙃 

So hey, if that's the case and this isn't a change that should be accepted go ahead and close the PR without merging! :)

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -27,7 +27,7 @@ new style sheet file in the `docs` directory:
 └─ mkdocs.yml
 ```
 
-Then, add the following lines to `mkdocs.yml`:
+Then, add the following lines to `mkdocs.yml` before your `theme` section:
 
 ``` yaml
 extra_css:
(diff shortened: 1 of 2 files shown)


squidfunk:
Thanks for the PR. All code examples that contain YAML always specify the full path from the root. Albeit I understand that you ran into troubles, we shouldn't make the documentation more verbose. I strongly recommend setting up [configuration validation](https://squidfunk.github.io/mkdocs-material/creating-your-site/#visual-studio-code), because then your IDE would've told you that the setting is not in the right place 😉

PR #5281, opened 2023-03-27
Allow overriding all HTML-inlined icons via yaml only

nejch:
I noticed that although most icons can be overridden in the yaml, for some of them we potentially need to override entire partials which can be a bit more work to align with upstream changes.

This takes the existing approach for configurable icons and applies it to all of the ones inlined via HTML. I realize it gets a bit repetitive so happy to extract this somewhere if needed, e.g. along with `icons.html`, but this seemed like a smaller change for now.

Also, this assumes that some of the icon config is reused between different partials. But perhaps they should be split in the config for more granular control (e.g. in the left/right in the footer vs nav).

Edit: not entirely sure if I should also commit the changes generated in `material/.overrides/` for this, seems like CI will want that.

### DIFF
--- a/docs/setup/changing-the-logo-and-icons.md
+++ b/docs/setup/changing-the-logo-and-icons.md
@@ -72,6 +72,37 @@ theme:
   [Favicon support]: https://github.com/squidfunk/mkdocs-material/releases/tag/0.1.0
   [Favicon default]: https://github.com/squidfunk/mkdocs-material/blob/master/material/assets/images/favicon.png
 
+### Site icons
+
+Most icons you see on your site, such as navigation icons, can also be changed. For example,
+to change the navigation arrows in the footer, add the following lines to `mkdocs.yml`:
+
+```yaml
+theme:
+  icon:
+    previous: fontawesome/solid/angle-left
+    next: fontawesome/solid/angle-right
+```
+
+The following is a complete list of customizable icons used by the theme:
+
+| Icon name    | Purpose                                                                       |
+|:-------------|:------------------------------------------------------------------------------|
+| `logo`       | See [Logo](#logo)                                                             |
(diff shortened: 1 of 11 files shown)


squidfunk:
Thanks for the PR! Your changes look good so far. However, I don't see any documentation updates? Do you have a good idea where we could mention this in the documentation so it's actually discoverable?

nejch:
Thanks for taking a look @squidfunk! I've pushed a docs commit and rebased.

It seems to me like [Changing the logo and icons](https://squidfunk.github.io/mkdocs-material/setup/changing-the-logo-and-icons/) is the best place to document this, as it already has similar content and examples (for the logo etc). Some of the documentation for existing options is a bit scattered around other pages (for good reason, as it's part of other features as well), so I've created a table with references where relevant. Let me know what you think!

I based the table of all available icons off of the existing table for available blocks (https://squidfunk.github.io/mkdocs-material/customization/#overriding-blocks).

I also took the docs example for a spin on this branch to make sure, seems ok:

image

nejch:
But I'm also happy to change this docs approach if it's not entirely aligned. If you had something else in mind, just let me know and I'll try to rework/rebase ;)

squidfunk:
Sorry for not getting back to you yet, I need to get some time to review the impact of this. Please give me some more time 😅

squidfunk:
@nejch finally found some time, LGTM, and thanks for the great work and also updating the documentation!

nejch:
Nice, thanks @squidfunk!

squidfunk:
By the way, this will be part of 9.2.0, so I changed the base branch to `merge/piri-piri`, which contains the latest features that will be released as the 12k funding goal was hit 😊 This will be a beta release at first.

PR #5293, opened 2023-03-29
Corrected Japanese translation of "Created"

tealgreen0503:
I have revised the word "Created" to be more natural for Japanese speakers.

The term "作成した" does not convey the meaning of the date something was created for Japanese speakers. Instead, the more direct expression "作成日" is more natural as it explicitly indicates the "date of creation."


squidfunk:
Sounds good to me, thanks for your contribution!

PR #5295, opened 2023-03-30
adding Trivy to open source user list

AnaisUrlichs:
Hi there,

trivy has been using the material theme for mkdocs for a long time now, and would be cool to be lister :) 
https://aquasecurity.github.io/trivy/


squidfunk:
Thanks for supporting and sponsoring the project! With 16k stars, I guess this project is pretty popular and well known, so we can surely add it to the list 😊

PR #5389, opened 2023-04-18
docs(pymdownx): add `url_download` option

robingenz:
Add the [`url_download`](https://facelessuser.github.io/pymdown-extensions/extensions/snippets/#options) option:

> Allows URLs to be specified as file snippets. URLs will be downloaded and inserted accordingly.


squidfunk:
Thanks!

PR #5444, opened 2023-04-29
added katex documentation

adhadse:
The KaTeX documentation as described [here in Arithmatex](https://facelessuser.github.io/pymdown-extensions/extensions/arithmatex/#loading-katex) docs is old, since:
1. KaTeX now does support its own [renderer](https://katex.org/docs/autorender.html)
2. The documentation will not work with instant loading with mkdocs-Material.

New users who wants fast loading may find it absurdly weird why KaTeX doesn't seem to work with XHR (instant loading) and so this little documentation without making mkdocs-material to support KaTeX out of the box will help those users to utilize KaTeX in their docs.


adhadse:
Thinking about this, this would promote using CDNs. Are they [good/bad](https://httptoolkit.com/blog/public-cdn-risks/)?

adhadse:
I'll promptly update the PR with you suggestions by next week and ping you!

Thank you, for the suggestions.

adhadseKavida:
@katharinalisalin I've updated the PR, please have a look. Does it require more changes? Will it be helpful enough for users?

katharinalisalin:
Thanks again for your updated version, superseded by #5553 

We decided to restructure the entire math page to make it more transparent for other users how to integrate both of these libraries and explained the pros and cons.

PR #5454, opened 2023-05-01
add vscode remote dev container support

prnam:
Adds VS Code remote dev container support for those developers who use VS Code and containers for development.

- It is configures to create container with Nodejs 18 LTS & Python 3.11
- Post creation of the container it executes the essential commands in [environment setup](https://squidfunk.github.io/mkdocs-material/customization/#environment-setup) and build it, thus enabling quick and easy way to get everything setup for new contributors without context switching.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,11 @@
+{
+	"name": "Material for MkDocs - VS Code dev container",
+	"image": "mcr.microsoft.com/devcontainers/typescript-node:0-18",
+	"features": {
+		"ghcr.io/devcontainers/features/python:1": {
+			"installTools": true,
+			"version": "3.11"
+	}
+	},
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! Honestly, I'm not feeling comfortable maintaining this, because none of us maintainers uses dev containers. Are you up for maintaining this configuration for the time to come? If so, we can merge it. Additionally, could you please clean up the source files? There are a lot of commented out lines that seem very generic.

prnam:
I can maintain it.

Comments are kept on purpose those are helpful for folks to learn what each does than explicitly search what it is. Ideally we keep them but if you insist I can remove, @squidfunk. Let me know! 🙂

squidfunk:
Thanks! Yes, let's remove them. Folks who are using dev containers should consult the dev containers documentation, as commented-out is not executed and thus just rots ☺️ LGTM!

PR #5460, opened 2023-05-02
Revert cache to use `github.ref` instead of `github.sha`

kamilkrzyskow:
Fixes my mistake from #5448.
I added a code annotation to the Publishing Guide, explaining when should the users expect a cache update. 
I considered to add another sentence "If your project **requires** more frequent cache updates, then use `github.sha`", but I guess this is a too much of an edge-case, that people might just overuse "just in case". Also using a YYYY-MM-DD date instead would be better than `github.sha`, because it would limit caches to 1 day.

Sorry for the PR chaos 🙏😅

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -50,7 +50,7 @@ jobs:
         uses: actions/cache@v3
         id: cache
         with:
-          key: mkdocs-material-${{ github.sha }}
+          key: mkdocs-material-${{ github.ref }}
           path: .cache
           restore-keys: |
             mkdocs-material-
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -40,11 +40,11 @@ contents:
               python-version: 3.x
           - uses: actions/cache@v3
             with:
-              key: mkdocs-material-${{ github.sha }}
+              key: mkdocs-material-${{ github.ref }} # (3)!
               path: .cache
               restore-keys: |
                 mkdocs-material-
-          - run: pip install mkdocs-material # (3)!
+          - run: pip install mkdocs-material # (4)!
           - run: mkdocs gh-deploy --force
     ```
 
@@ -53,7 +53,10 @@ contents:
     2.  At some point, GitHub renamed `master` to `main`. If your default branch
         is named `master`, you can safely remove `main`, vice versa.
 
-    3.  This is the place to install further [MkDocs plugins] or Markdown
+    3.  The `github.ref` property assures that the cache will
+        update on each pull request merge.
+
+    4.  This is the place to install further [MkDocs plugins] or Markdown
         extensions with `pip` to be used during the build:
 
         ``` sh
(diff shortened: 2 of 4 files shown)


squidfunk:
Thanks for the quick fix and no bad feelings! ☺️

kamilkrzyskow:
@squidfunk 
This will probably come of as some bad joke, and I'm embarrassed to say it, but the action saved the `refs/heads/master` `github.ref`
https://github.com/squidfunk/mkdocs-material/actions/runs/4861591365/jobs/8666814586#step:14:5

This means that even for pull request merges the `github.ref` uses the `branch` rules 😨
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

> The fully-formed ref of the branch or tag that triggered the workflow run. For workflows triggered by push, this is the branch or tag ref that was pushed. For workflows triggered by pull_request, this is the pull request merge branch. For workflows triggered by release, this is the release tag created. For other triggers, this is the branch or tag ref that triggered the workflow run. This is only set if a branch or tag is available for the event type. The ref given is fully-formed, meaning that for branches the format is refs/heads/<branch_name>, 

> **for pull requests** it is refs/pull/<pr_number>/merge, and for tags it is refs/tags/<tag_name>. For example, refs/heads/feature-branch-1.

After rereading that I figured that the pull request `github.ref` would work out, but I was mistaken, yet again...

---

So I'm currently out of ideas, because I do think that `github.sha` might be used to abuse the cache creation system with 5 million users and 10GB storage cap, and adding a step to setup a date variable seems to be too much, but maybe the lesser evil 🤔
```yml
    - name: Set Date String
      run: echo "date=$(date +'%Y-%m-%d')" >> $GITHUB_ENV
    # ...
    key: mkdocs-material-${{ env.date }}
    # ...
```

Sorry for the confusion once again 😞

squidfunk:
Haha, no worries. Many thanks for taking the time and investigating, much appreciated. Should we revert the last commit? I'm not sure what to do here and would follow your recommendation.

squidfunk:
Thanks again! I'm having a little hard time wrapping my head around this (currently fighting other battles), so I'll trust your recommendation on this. If you wish to follow up on this, feel free to take the necessary time to come up with a solution that is simple to follow for the user (= less maintenance for us maintainers because of fewer questions) and works reasonably well. Maybe someone else with more experience can share some ideas as well ☺️

> > It might lead to flaky builds that resolve on the next day

> I'm not sure if I understand this sentence.
> Are you saying that adding a date might create a scenario where a build doesn't work one day and works on the next day?

As we both know, caching is hard. We had some problems a while ago with stale Python dependencies that were tricky to resolve. For this reason, I deactivated caching of pip-installed dependencies. Scoping caches to days introduces a new axis of fragility. When we run into a caching issue on day 1, and try to debug it on day 2, it might be hard to reproduce. When it's scoped to a ref, it's likely easier to debug, as we don't have time working against us. That's basically what I meant. I'm all for solutions that are simple, yet solve the widest range of problems while being easy to debug and extend. Maybe we can find such a solution here as well.

Issue #5455, opened 2023-05-01
Support map for plugins in JSON schema


pawamoy:
### Context

The documentation recommends setting plugins as a map to be able to deep-merge plugins from a base `mkdocs.yml` into a `mkdocs.insiders.yml` file. However it seems that the JSON schema expects an array.

I'm using this in VSCode to validate the MkDocs configuration:

```json
    "yaml.schemas": {
        "https://squidfunk.github.io/mkdocs-material/schema.json": "mkdocs.yml"
    },
```

### Description

I'd like the schema to support defining plugins as a map as well as an array. The following should not be underlined as having issues:

```yaml
plugins:
  search: {}
```

### Related links

- https://squidfunk.github.io/mkdocs-material/insiders/getting-started/?h=inher#built-in-plugins ("Sharing plugin and extension configuration" tip)

### Use Cases

Inheriting from a base config while avoiding duplicate configuration for plugins between both files.

### Visuals

![pluginsmaps](https://user-images.githubusercontent.com/3999221/235463421-7d291f6d-b356-4cff-be0c-2bc075eb07ff.png)

### Before submitting

- [X] I have read and followed the [change request guidelines](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/).
- [X] I have verified that [my idea is a change request and not a bug report](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/#its-not-a-bug-its-a-feature).
- [X] I have ensured that, to the best of my knowledge, [my idea will benefit the entire community](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/#benefit-for-the-community).
- [X] I have included relevant links to [the documentation](https://squidfunk.github.io/mkdocs-material), related [issues](https://github.com/squidfunk/mkdocs-material/issues), and [discussions](https://github.com/squidfunk/mkdocs-material/discussions) to underline the need for my idea.

squidfunk:
Thanks for suggesting! This is actually an excellent issue for outside contributors, as it is not related to the codebase of Material for MkDocs. If somebody is up, a PR is very much welcome! If not, I'll look into it when I find some time.

squidfunk:
[...]

PR #5461, opened 2023-05-02
Support plugins map in JSON schema

pawamoy:
This is what I have right now. I'm not a JSON Schema expert so there might be a more elegant solution. https://www.jsonschemavalidator.net/ has trouble resolving the references:

> Could not resolve schema reference 'https://raw.githubusercontent.com/mondeja/mkdocs-include-markdown-plugin/master/schema.json#/oneOf/1/properties'

However from what I could read on the web, this is a perfectly valid reference (JSON pointer).

Other tools like [`check-jsonschema`](https://pypi.org/project/check-jsonschema/) seem to have no issues with the schema, however they also validate instances that are invalid, so I think they don't understand the schema as I expect them to.

I wonder if using `additionalProperties` with `anyOf` is the right thing to do. I tried instead to use `anyOf` with `properties`, but this is invalid against JSON Schema Draft v7:

```json
      "anyOf": [
        {
          "properties": {
            "$ref": "#/definitions/built-in-properties"
          }
        },
        {
          "properties": {
            "$ref": "#/definitions/external-properties"
          }
        },
        {
          "properties": {
            "$ref": "#/definitions/external-community-properties"
          }
        }
      ]
```

Here's a string you can use to test the schema: `{"mkdocstrings": {"handlers": {}}}`. It *must* validate. Replace "handlers" by garbage and it *must not* validate anymore.

---

Implements #5455.

### DIFF
--- a/docs/schema/plugins.json
+++ b/docs/schema/plugins.json
@@ -2,22 +2,42 @@
   "$schema": "https://json-schema.org/draft-07/schema",
   "title": "Plugins",
   "markdownDescription": "https://www.mkdocs.org/dev-guide/plugins/",
-  "type": "array",
-  "items": {
-    "anyOf": [
-      {
-        "$ref": "#/definitions/built-in"
+  "oneOf": [
+    {
+      "type": "array",
+      "items": {
+        "anyOf": [
+          {
+            "$ref": "#/definitions/built-in"
+          },
+          {
+            "$ref": "#/definitions/external"
+          },
+          {
+            "$ref": "#/definitions/external-community"
+          }
+        ]
       },
-      {
-        "$ref": "#/definitions/external"
-      },
-      {
-        "$ref": "#/definitions/external-community"
+      "uniqueItems": true,
+      "minItems": 1
+    },
+    {
+      "type": "object",
+      "additionalProperties": {
+        "anyOf": [
+          {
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! I'm not sure whether this is a good idea. The problem is that we're now relying on the internal structure of the plugins. While this might be okay for the built-in plugins, as we manage them ourselves, it might completely break JSON validation when the structure of one of the external sources is changed, which we don't control.

I'm aware that we advertise the key-value syntax for `mkdocs.yml` on our documentation, but I think using that syntax is really not a good idea. A better idea might be to use schema merging and [anchors](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/) to deduplicate configurations between files. We need to investigate whether we can switch to a better approach to maintain/merge multiple configurations.

Additionally, many of the built-in plugins support multiple instances, e.g. the `optimize`, `privacy`, `social` and `tags` plugins. With the key/value syntax it's impossible to make use of multiple instances.

I'll leave this PR open until this is resolved.

pawamoy:
Yeah that makes sense, I'd be okay with using anchors instead of a map. I'll do that. Thanks for your help!

squidfunk:
Great! I decided to close the PR, as we have an open issue that track the same thing

Issue #5503, opened 2023-05-10
Update GitHub Action step "Check docker image" to accommodate for usernames with upper case characters


joshooaj:
### Context

We intend to keep our fork in sync, and automatically publish updated container images to our GHCR.

### Description

Our organization username has upper case characters in it, and this causes the "Check docker image" build step to fail with the following error:

```
Run docker run --rm -i --user $(id -u):$(id -g) -e GH_TOKEN=${GH_TOKEN} -v ${PWD}:/docs ghcr.io/MilestoneSystemsInc/mkdocs-material-insiders:9.1.11-insiders-4.33.1 new .
docker: invalid reference format: repository name must be lowercase.
See 'docker run --help'.
Error: Process completed with exit code 125.
```

To resolve this for any user or organization with upper case characters in their GitHub name, I'm proposing to convert the value of `github.event.repository.full_name` to lower case before calling `docker run` when validating the docker image.

_Note: This is being shared as a change request instead of a bug because the bug template requires an attached reproduction. I don't think it makes sense to follow the [creating-a-reproduction guide](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/) for an issue related to GitHub Action workflow(s)._

## Proposed change

In the example below, the "Check Docker image" step has the following modifications:

1. The value of `github.event.repository.full_name` is stored in an environment variable named `REPO_FULL_NAME`.
2. The executions of `docker run` are updated to use `${REPO_FULL_NAME,,}` in the image name instead of the raw value of `github.event.repository.full_name`.

_Note: The syntax `${VAR,,}` is an example of [parameter expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) in bash. In this case it converts all upper case characters to lowercase. The reverse could be done with `${VAR^^}`._

```yaml
      - name: Check Docker image
        working-directory: /tmp
        env:
          REPO_FULL_NAME: '${{ github.event.repository.full_name }}'
        run: |
          docker run --rm -i --user $(id -u):$(id -g) -e GH_TOKEN=${GH_TOKEN} -v ${PWD}:/docs ghcr.io/${REPO_FULL_NAME,,}:${{ steps.meta.outputs.version }} new .
[...]

PR #5504, opened 2023-05-10
Update "Check Docker image" build step to handle upper case chars in repo names

joshooaj:
Issue #5503 describes an issue where the "Check Docker image" build step throws an error when the build action runs on a fork owned by a user with one or more upper case characters in their name.

This PR handles this by assigning the value of `github.event.repository.full_name` to an environment variable, and then referencing that variable when calling `docker run` using bash [parameter expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) to convert characters to lower case.

The reason the build doesn't fail earlier in the job is because [docker/metadata-action](https://github.com/docker/metadata-action) automatically converts image names to lower case.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -144,9 +144,11 @@ jobs:
 
       - name: Check Docker image
         working-directory: /tmp
+        env:
+          REPO_FULL_NAME: '${{ github.event.repository.full_name }}'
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks!

Issue #5517, opened 2023-05-15
Issue with Open Sans font in social plugin


croesus:
### Bug description

Google's Open Sans font package contains several font variations in subdirectories. The code picks the last one that ends in "-Regular.ttf" which returns an arbitrary variation based on the order the OS returns files.

```
Open_Sans
├── OFL.txt
├── OpenSans-Italic-VariableFont_wdth,wght.ttf
├── OpenSans-VariableFont_wdth,wght.ttf
├── README.txt
└── static
    ├── OpenSans
    │   ├── OpenSans-Bold.ttf
    │   ├── OpenSans-BoldItalic.ttf
    │   ├── OpenSans-ExtraBold.ttf
    │   ├── OpenSans-ExtraBoldItalic.ttf
    │   ├── OpenSans-Italic.ttf
    │   ├── OpenSans-Light.ttf
    │   ├── OpenSans-LightItalic.ttf
    │   ├── OpenSans-Medium.ttf
    │   ├── OpenSans-MediumItalic.ttf
    │   ├── OpenSans-Regular.ttf
    │   ├── OpenSans-SemiBold.ttf
    │   └── OpenSans-SemiBoldItalic.ttf
    ├── OpenSans_Condensed
    │   ├── OpenSans_Condensed-Bold.ttf
    │   ├── OpenSans_Condensed-BoldItalic.ttf
    │   ├── OpenSans_Condensed-ExtraBold.ttf
    │   ├── OpenSans_Condensed-ExtraBoldItalic.ttf
    │   ├── OpenSans_Condensed-Italic.ttf
    │   ├── OpenSans_Condensed-Light.ttf
    │   ├── OpenSans_Condensed-LightItalic.ttf
    │   ├── OpenSans_Condensed-Medium.ttf
    │   ├── OpenSans_Condensed-MediumItalic.ttf
    │   ├── OpenSans_Condensed-Regular.ttf
    │   ├── OpenSans_Condensed-SemiBold.ttf
    │   └── OpenSans_Condensed-SemiBoldItalic.ttf
    └── OpenSans_SemiCondensed
        ├── OpenSans_SemiCondensed-Bold.ttf
        ├── OpenSans_SemiCondensed-BoldItalic.ttf
        ├── OpenSans_SemiCondensed-ExtraBold.ttf
        ├── OpenSans_SemiCondensed-ExtraBoldItalic.ttf
        ├── OpenSans_SemiCondensed-Italic.ttf
[...]

PR #5518, opened 2023-05-15
Select the correct font variant where there are options in social plugin

croesus:
Ref [Issue 5517](https://github.com/squidfunk/mkdocs-material/issues/5517)


squidfunk:
Thanks! Did you test this with other fonts than Open Sans?

croesus:
I haven't tested it exhaustively but it works with the following fonts that have different layouts in the zip:
Lobster - contains a single .ttf at the top level of the zip
Roboto - contains several ttfs at the top level of the zip
Inter - contains several ttfs in a `static/` subdirectory in the zip
Open Sans - contains several versions of the font in subdirectories of the `static/` subdirectory of the zip

squidfunk:
Thanks for testing! Let's give this a try.

PR #5536, opened 2023-05-21
Strip tags of ToC item's aria label

pawamoy:
Some plugins/extensions might use the data-toc-label attribute of headings to a particular string for the ToC item title.

While not explicitly documented as possible or allowed, they might set some HTML code instead of text. In this case, the aria label of Toc nav elements break the HTML as they will contain unescaped double quotes.

To fix this, we use the `striptags` filter to remove any tags from the aria label, keeping text only.


src/partials/toc-item.html
```diff
@@ -28,7 +28,7 @@
 
   <!-- Table of contents list -->
   {% if toc_item.children %}
-    <nav class="md-nav" aria-label="{{ toc_item.title }}">
+    <nav class="md-nav" aria-label="{{ toc_item.title|striptags }}">
```
squidfunk, line 31: Please stick to our code style and insert spaces left and right to the `|` pipe ☺️ Also, please build the theme and make the changes in the `material` folder, or they won't be picked up.

squidfunk (changes requested):
Thanks! Sounds legit, as previously discussed via Gitter. Some minor adjustments.

pawamoy:
I ran `npm install` then `npm run build`.

squidfunk:
Thanks! You need to run `npm run build:all`, or the custom overrides are not built.

pawamoy:
Oh OK, this wasn't clear to me from the docs:

> While this command will build all theme files, it will skip the overrides used in Material for MkDocs' own documentation which are not distributed with the theme. If you forked the theme and want to build the overrides as well, use:

I understood this as "use build:all if you have a custom fork of the theme" (I don't, I'm just sending a PR).
Maybe something simpler like "use build:all if you modified partials/overrides" would do the trick :slightly_smiling_face:

squidfunk:
Okay, sorry, last thing: the iconsearch index got deleted - please remove that from the commit as well. This is likely because the emoji database wasn't found in the virtual environment (we kind of assume that `venv` is the folder name):

https://github.com/squidfunk/mkdocs-material/blob/2cff28709745af3263f90623e50c9cacc2b40f9c/tools/build/index.ts#L276

This is a shortcoming of our current build process. We're going to revamp the docs and the process in the near future.

pawamoy:
Sure, no worries! Indeed I usually use `.venv` as folder name :slightly_smiling_face:

squidfunk:
Perfect, good to go now!

PR #5565, opened 2023-05-27
Update Dockerfile to latest alpine

yubiuser:
Updates the Dockerfile to Alpine 3.18. Additionally, update `pip` itself before installing any packages.


Dockerfile
```diff
@@ -18,7 +18,7 @@
 # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
 # IN THE SOFTWARE.
 
-FROM python:3.11.0-alpine3.17
+FROM python:3.11.3-alpine3.18
```
crazy-max, line 21: We can just use `major.minor` imo:

```suggestion
FROM python:3.11-alpine3.18
```

crazy-max (commented):
LGTM

squidfunk:
Perfect, thanks!

PR #5617, opened 2023-06-17
Rename getting-started.md to getting-started

Kordokareem:
Sarmadi


Kordokareem:
Thanks again so for your

squidfunk:
Thanks for the PR: However, it makes no sense and will actually break the build, because MkDocs expects the files to be suffixed with `.md`. Thus, this PR cannot be merged. Closing.

PR #5626, opened 2023-06-20
Add support for user-provided python packages

hellt:
Hi @squidfunk 
With this PR I wanted to solve a challenge of onboarding user-provided packages to the mkdocs-build artifact.

Even though build workflow includes some plugins by default, the list is not extensive, nor it assumes that users should edit it.

With this little change the workflow for users who want to install additional plugins/packages is as follows:

1. users check in a user-requirements.txt file with packages they desire to be installed after all default packages are done installing
2. done

the build workflow will copy to the build context all files ending with *requirements.txt and then a new bash command will start `pip install -r user-requirements.txt` if the user-requirements.txt files was found in the repo.

That way we provide flexibility in what users can install on top of the default plugins.

PS. This is a backwards compatible change, nothing is required from users who don't want to have custom packages installed.

If you agree with this change, I can raise a similar PR against insiders.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ WORKDIR /tmp
 COPY material material
 COPY package.json package.json
 COPY README.md README.md
-COPY requirements.txt requirements.txt
+COPY *requirements.txt ./
 COPY pyproject.toml pyproject.toml
 
 # Perform build and cleanup artifacts and caches
@@ -66,6 +66,8 @@ RUN \
       "pillow>=9.0" \
       "cairosvg>=2.5"; \
   fi \
+&& \
+  [ -e user-requirements.txt ] && pip install -r user-requirements.txt \
 && \
   apk del .build \
 && \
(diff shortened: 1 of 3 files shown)


squidfunk:
That's a great idea. Could you also add a note to the documentation, please?

hellt:
@squidfunk, what would be a proper place to add it?

squidfunk:
Best fit would probably be the box on [how to add plugins to the Docker image](https://squidfunk.github.io/mkdocs-material/getting-started/#with-docker), but I'm open to suggestions.

hellt:
Yes, this looks like a perfect place to add this info. Added in [aec8973](https://github.com/squidfunk/mkdocs-material/pull/5626/commits/aec897302ff6dc9cfa9857e906181da9616d1485)

squidfunk:
LGTM!

hellt:
@squidfunk would you be able to bring these changes to insiders or shall I do it? I am primarily interested in insiders supporting user-provided packages installation.

squidfunk:
I'll do the merge later, you don't need to do it 😊

PR #5632, opened 2023-06-23
Add missing comma to light-mode selector example

charliermarsh:
## Summary

The dark-mode example (a few lines below) has a comma here, but the light-mode example does not. I _think_ it's just missing, since IIUC this should be an "OR" query.

Thanks for the great project! (We use `mkdocs-material` to power the [Ruff documentation](https://beta.ruff.rs/docs/).)

### DIFF
--- a/docs/reference/images.md
+++ b/docs/reference/images.md
@@ -182,7 +182,7 @@ hash fragment to the image URL:
     === "Custom light scheme"
 
         ``` css
-        [data-md-color-scheme="custom-light"] img[src$="#only-dark"]
+        [data-md-color-scheme="custom-light"] img[src$="#only-dark"],
         [data-md-color-scheme="custom-light"] img[src$="#gh-dark-mode-only"] {
           display: none; /* Hide dark images in light mode */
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! Jup, there needs to be a comma. Good catch! Oh, and I'm psyched to see Ruff use Material for MkDocs!

charliermarsh:
Of course!

I will admit that I didn't look into it deeply, but I also had to add this (I believe only the latter was necessary -- dark-mode-only images still had `display: none` in dark mode):

```css
[data-md-color-scheme="astral-light"] img[src$="#only-light"],
[data-md-color-scheme="astral-light"] img[src$="#gh-light-mode-only"] {
  display: inline; /* Show light images in light mode */
}

[data-md-color-scheme="astral-dark"] img[src$="#only-dark"],
[data-md-color-scheme="astral-dark"] img[src$="#gh-dark-mode-only"] {
  display: inline; /* Show dark images in dark mode */
}
```

PR #5656, opened 2023-06-28
Custom search.seperator with lookbehind, in combination with highlighter, produces highlighting all over the place

### LINKED ISSUE #5655: Custom search.seperator with lookbehind, in combination with highlighter, produces highlighting all over the place
### Context

A custom search.seperator with lookbehind, in combination with highlighter, produces highlighting all over the place on the page.

This occured when adjusting the camelCasing/PascalCasing seperator to not seperate words if the words get too small e.g. `AStar` should not be seperated to highlight `A`+`Star`. In such cases the relevance of pages containing `AStar` compared to chapters containing just `Star` gets left behind and not found by users.

### Description

There is in the sourcecode a hack in place `src\assets\javascripts\integrations\search\highlighter\index.ts` which reuses the search.seperator for the highlighting by removing lookaheads, it does however not yet remove lookbehinds. Thus when using lookbehinds the highlighting goes bonkers.

There would be two ways to solve this:
* add a new search.seperator_highlight parameter which only is used for highlight seperation (new feature)
* for now also remove lookbehinds (bug of in place hack)

So its kind of a bug, but also a change request.

To solve the bug adjust in  `src\assets\javascripts\integrations\search\highlighter\index.ts`:
```
-    const temp = term.replace(/(\(\?[!=][^)]+\))/g, "")
+    const temp = term.replace(/(\(\?[!=<][^)]+\))/g, "")
```

### Related links

- [Highlighter camelCase](https://squidfunk.github.io/mkdocs-material/blog/2021/09/13/search-better-faster-smaller/?h=camel+case#case-changes)

### Use Cases

In general seperating things like `IsFloat` etc also is not useful as when searching code Is+Float will result in many results but not in the function to find.

So the idea was to start the seperation of camelCasing/PascalCasing only with `3>=` characters. Which would start with the classic `Get / Set / Can` but not with Is, or Single char prefixes.

The separator casing fragment was modified in such way:

`(?!\b)(?=[A-Z][a-z]{2,})(?<!\b\w{0,2})`

Sample:
* Starts seperating on GetThis, but not AStar or IsStar

### Visuals

_No response_

### Before submitting

- [X] I have included relevant links to [the documentation](https://squidfunk.github.io/mkdocs-material), related [issues](https://github.com/squidfunk/mkdocs-material/issues), and [discussions](https://github.com/squidfunk/mkdocs-material/discussions) to underline the need for my idea.

hendrikp:
Closes #5655


squidfunk:
Thanks for the PR! Also see https://github.com/squidfunk/mkdocs-material/issues/5655#issuecomment-1612756351

PR #5657, opened 2023-06-29
Revert "Add support for user-provided python packages"

squidfunk:
Reverts squidfunk/mkdocs-material#5626

PR #5658, opened 2023-06-29
Added user-requirements.txt install workflow

hellt:
Hi @squidfunk 
sorry to hear that I made a build break, apparently, I didn't use a proper bashism for a conditional pip install.

Now I used the same syntax you used in the Dockerfile, and now build starts fine

image

re #5626 #5657


squidfunk:
Thanks for investigating!

PR #5710, opened 2023-07-14
wrote section on how to highlight multiple lines of code

nh916:
# Description
I was using this software a few months ago to write documentation for my code and it took me a very long time before I could figure out how to highlight many lines of code instead of just a single line. I tried multiple different ways for a very long time until I eventually found the right combination. I would like to write this documentation and make the contribution so no one else in the future has to be confused on this.

### DIFF
--- a/docs/reference/code-blocks.md
+++ b/docs/reference/code-blocks.md
@@ -380,6 +380,31 @@ def bubble_sort(items):
 
 </div>
 
+
+??? tip "Highlight Multiple Lines"
+    To highlight multiple lines, you can specify a range of line numbers using the format `hl_lines="start-end"`:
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! Please use [content tabs](https://squidfunk.github.io/mkdocs-material/reference/content-tabs/) instead of an admonition/details, similarly to [how we're doing it here](https://squidfunk.github.io/mkdocs-material/reference/content-tabs/#unordered-list_1)? The tabs could be named something like "Lines" and "Line ranges", I think this should be clear enough.

squidfunk:
Superseded by 1ff2aec06.

PR #5711, opened 2023-07-14
add example for light/dark mode with primary and accent colors

nh916:
updated documentation to include an example for light and dark mode with primary and accent color

when I was originally working with the software to have both light/dark mode and primary/accent colors it took me a long time before figuring out how to have both, so I wrote documentation example, so no one else has to struggle to find it

### DIFF
--- a/docs/setup/changing-the-colors.md
+++ b/docs/setup/changing-the-colors.md
@@ -308,6 +308,29 @@ theme:
         name: Switch to system preference
 ```
 
+??? Example "light / dark mode with primary and secondary colors"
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. However, I'm not sure we should add more combined examples to our documentation, because by that logic, we need combined examples for all features that allow for multiple settings, which means we will have much more stuff to maintain and keep fresh. I understand that you had trouble figuring it out, but it's the first time we hear about it, which is why we don't consider this a common pitfall and won't consider merging this right now.

Thanks for your understanding.

PR #5716, opened 2023-07-16
Prepare for MkDocs 1.5 support of `script` tag attributes

oprypin:
Note that the current state of the code will *crash* with MkDocs 1.5, soon to be released. As follows:

        {% if path.endswith(".mjs") %}
      File "python3.10/site-packages/jinja2/utils.py", line 83, in from_obj
        if hasattr(obj, "jinja_pass_arg"):
    jinja2.exceptions.UndefinedError: 'mkdocs.config.config_options.ExtraScriptValue object' has no attribute 'endswith'

New official instructions: https://github.com/mkdocs/mkdocs/blob/8353493606344f3b25922e2a877c4b8485827a69/docs/dev-guide/themes.md#picking-up-css-and-javascript-from-the-config

### DIFF
--- a/material/base.html
+++ b/material/base.html
@@ -241,13 +241,13 @@
     {% endblock %}
     {% block scripts %}
       <script src="{{ 'assets/javascripts/bundle.220ee61c.min.js' | url }}"></script>
-      {% for path in config.extra_javascript %}
-        {% if path.endswith(".mjs") %}
-          <script type="module" src="{{ path | url }}"></script>
-        {% else %}
-          <script src="{{ path | url }}"></script>
-        {% endif %}
-      {% endfor %}
+      {%- for script in config.extra_javascript %}
+        {%- if script.path %}
+          {{ script | script_tag }}
+        {%- else %}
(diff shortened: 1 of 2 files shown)


tuanggo:
open

squidfunk:
Thanks for the PR! We can issue a new patch release of 9.1.x to make sure that we're ready for MkDocs 1.5, and after MkDocs 1.5 is out, drop support for MkDocs 1.4, once 9.2.0 is stable.

However, there seems to be an error – could you please check why the build now fails?

squidfunk:
Additionally, are there any changes that are necessary to plugins? As I understand, `extra_javascript` now has another structure. Some of the built-in plugins alter and/or access `extra_javascript`, so it would be good to know.

oprypin:
> However, there seems to be an error – could you please check why the build now fails?

This build system seems to try to parse Jinja syntax as HTML, and although it somehow worked before, it doesn't work now, so I'd just need to randomly try until it happens to like the syntax again

squidfunk:
Superseded by e897b8912.

PR #5739, opened 2023-07-21
Remove unnecessary aria-label from nav element

pereorga:
This removes the "aria-label" attribute from the nav element.

The nav element inherently carries accessibility features recognized by screen readers, making a label unnecessary in this context. The nav element itself is a landmark role, and screen readers can recognize it as such. Including an additional label may be redundant and in this case potentially misleading, especially as it doesn't have a "role" that corresponds to the label.

The issue has been reported to me by an Accessibility expert. If there are questions, or if it doesn't make sense, I'm happy to pass the feedback and report the answer back.

Thanks!


squidfunk:
Thanks for the PR. I'm not an expert. We added them because @jaimeiniesta raised a PR in https://github.com/squidfunk/mkdocs-material/commit/e7f763227864158b5d3529fab7dcb6333a13a046. Maybe he can comment what the rational is, and you both can discuss what makes sense 😊

pereorga:
Thanks, I see this was added in `src/partials/header.html` too.

@jaimeiniesta my understanding is that this is a menu, not a tab list. That is why the HTML element `<nav>` is used.

Perhaps you got confused because the element has a class and data attribute that contains the name "tab"? (`class="md-tabs" data-md-component="tabs"`). I think this has nothing to do with the role of the element.

As mentioned, I'm not an expert. If what I'm saying is wrong, I'm happy to forward your answer to an expert and come back.

squidfunk:
Converting to draft, because this PR is not ready for merging yet. We need to sort out what we should do to provide optimal support for screen readers. Please check with the original authors whether removing is a good idea.

jaimeiniesta:
Hey there, I just wanted to show up here as I got mentioned but I'm traveling these days during my summer holidays and I don't have time to properly review this.

I suggest using the Axe dev tools to check this and add more light here:

https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd

squidfunk:
I'm closing this PR as stale. If somebody can come up with a proper explanation why this is or is not necessary, we can reconsider, but for now we'll keep the label.

PR #5743, opened 2023-07-22
Improve support for instant loading with keyboard navigation

squidfunk:
This PR improves instant loading for keyboard navigation. We need to refactor instant loading a bit to support adding and removal of elements in the `head` when they're not present. This needs to be addressed before merging this PR.


vedranmiletic:
> We need to refactor instant loading a bit to support adding and removal of elements in the `head` when they're not present. This needs to be addressed before merging this PR.

Sorry to go on a trajectory a bit, but this is very interesting. For our use case, the custom `onion-location` meta tag should be updated when navigating from a page to another page. Is there a description of the planned implementation somewhere?

squidfunk:
@vedranmiletic the instant loading code got a refactor and I took the chance to very thoroughly document how it's done and what the limitations are. Take a look at the [source code](https://github.com/squidfunk/mkdocs-material/blob/master/src/assets/javascripts/integrations/instant/index.ts). If questions remain, please ask them here.

filips123:
[This](https://github.com/squidfunk/mkdocs-material/blob/master/src/assets/javascripts/integrations/instant/index.ts#L228-L243) is the list of tags that get updated on instant loading, right?

I think it would be useful to let users extend this list easily (probably in the config file), so if we use overrides/templates to add custom tags, those would be handles as well. Support for adding and removing tags would also be useful when custom tags are added only to specific pages.

squidfunk:
Thanks for the suggestion. It's a better idea to deduplicate and update all tags, so there's no configuration necessary, which is exactly what we will do. This will also mean that custom tags are correctly added/removed/replaced.

squidfunk:
@vedranmiletic this PR implements auto-replacement of tags – this should also solve the problem you described ☺️

vedranmiletic:
@squidfunk that's great, eagerly waiting for the next release!

squidfunk:
Looks like it's going to be released it tomorrow.

squidfunk:
Released as part of [9.4.2](https://squidfunk.github.io/mkdocs-material/changelog/#9.4.2).

PR #5806, opened 2023-08-06
fix: version selector not 'hoverable' on some Android devices

### LINKED ISSUE #2429: Version selector not touch-friendly for Android devices
- [x] I've read the [contribution guidelines][1] and agree with them

__I've found a bug and checked that ...__

## Description

Chrome Android, unable to easily change version via the drop-down selector via touch interaction.

### Expected behavior

Be able to select version from drop-down on touch device.

### Actual behavior

Tapping version selector activates drop-down then navigates to 1st listed version.

1. Visit https://squidfunk.github.io/mkdocs-material-example-versioning/latest/
2. Tap the version to activate the drop-down selector.
3. Page will act as if you selected the top listed (1st item) version.

### Package versions / Project configuration / System information

Active release used for project demo site: https://squidfunk.github.io/mkdocs-material-example-versioning/latest/

axieum:
It appeared that the `pointer` media query for my Pixel 7 Pro device was reporting as `pointer: fine`, meaning the hover fix wasn't applying to the Version selector.

I can see this is already being used here:

https://github.com/squidfunk/mkdocs-material/blob/f458f04a5b6ff87f1ed0a5ffc6792b4ba74c5adb/src/assets/stylesheets/main/_typeset.scss#L295-L296

I confirmed working on Desktop, Android, and an iOS mobile device, but it may be best to triple **check iOS compatibility please**.

**Developer Note:** When you have Dev Tools connected to the Android device, it changes to `pointer: coarse` and works as expected, making debugging quite difficult. So you mustn't observe with Dev Tools open while trying to reproduce the error on an Android device.

Fixes #2429

### DIFF
--- a/src/assets/stylesheets/main/components/_version.scss
+++ b/src/assets/stylesheets/main/components/_version.scss
@@ -106,8 +106,7 @@
     }
 
     // Fix hover on touch devices
-    @media (pointer: coarse) {
-
+    @media (pointer: coarse), (hover: none) {
       // Switch off on hover
       .md-version:hover & {
         animation: hoverfix 250ms forwards;


squidfunk:
Hmm, I'm not feeling confident to merge this through right now, because some touch devices will not match `hover: none`, at least as far as I know. IIRC, `pointer: coarse` is the recommendation and we probably should switch other parts of our code base to it, but I don't have and Android or Windows phone at my disposal, so we definitely need to test this thoroughly before merging with the help of other users.

axieum:
Agreed. I did originally have `@media (pointer: coarse), (hover: none) {`, shall I bring that back?

squidfunk:
If you're confident that it solves the problem, we can give it a try.

PR #5823, opened 2023-08-09
Added info about markdown_captions extension to images.md

pana-exfo:
Added another option how markdown_captions extension can be used to show image alt text as caption.


squidfunk:
Thanks for the PR. Interesting! Does it allow to use HTML or Markdown inside the image caption?

pana-exfo:
> Thanks for the PR. Interesting! Does it allow to use HTML or Markdown inside the image caption?

Icons, links (Markdown and HTML) and footnotes work.

PR #5830, opened 2023-08-12
use 3.12-rc.alpine3.18 image

leeyuentuen:


squidfunk:
Thanks for the PR. However, we shouldn't use release candidates. We should wait until 3.12 is stable.

PR #5831, opened 2023-08-13
Fix: TypeError: file.replace is not a function

purplecandy:
Fixes issue #3890  which happens while developing on Linux.

Chowkidar returns an instance of `FSWatcher` which may or may not produce a [Stat](https://nodejs.org/api/fs.html#fs_class_fs_stats)  object with change events instead of just a string. The issue is easily reproducible on Linux systems.

### DIFF
--- a/tools/build/_/index.ts
+++ b/tools/build/_/index.ts
@@ -135,9 +135,11 @@ export function resolve(
 export function watch(
   pattern: string, options: WatchOptions
 ): Observable<string> {
-  return fromEvent(
+  return (fromEvent(
     chokidar.watch(pattern, options),
     "change"
+  ) as Observable<string|[string,{}]>).pipe(
+    map((event) => (typeof event === "string" ? event : event[0]))
   ) as Observable<string>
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. This sounds like something that should be fixed upstream, or is it documented behavior? Have you checked the issue tracker of chokidar whether this is a known issue?

purplecandy:
There are no open issues related to it as it's not a bug. It is just that the underlying watch event implementation uses different utilities depending on the OS.  The only thing I found related to it on chokidar was [this](https://github.com/paulmillr/chokidar/blob/master/README.md?plain=1#L208)

squidfunk:
Thanks again! Okay, so after investigating, I can say with confidence that the problem is triggered by RxJS, and not Chokidar. If there's more than one argument that is passed to the handler, RxJS emits an array, otherwise the first argument. However, we can just use the `resultSelector` to always return the first argument. Fixed in and superseded by 26de75679.

PR #5848, opened 2023-08-16
Avoid whitespaces between social link icons

sisp:
I've adapted the Jinja syntax related to generating social links in the footer to strip whitespaces around the `<a>` tags to avoid whitespaces between these tags as they are interpreted as actual whitespaces by the browser and, thus, add some implicit yet irrelevant spaces between the social link icons.

See this screenshot:

image


squidfunk:
Thanks for the PR. Actually, the whitespace is quite intentional, because otherwise there would be no visual indicator which link your are currently hovering or focusing, as when hovering the whitespace between the links, the normal cursor is shown. I'm not for removing that. If you want to remove it, just set `.md-social` to `flex`:

``` css
.md-social {
  display: flex;
}
```

sisp:
I get the cursor argument, but the icons color changes on hovering, and accessibility indicators work as well. I'd rather add an explicit margin between the links than rely on a somewhat arbitrary whitespace. Wouldn't you agree?

squidfunk:
Hmm, well, yeah. Honestly, I've closed this because what you asked is perfectly achievable with a few lines of customization, and we always need to make sure not to break stuff, so changing things like this isn't an easy decision with a project of this size.  However, I get the argument that it's easier and cleaner to customize when we set explicit spacing via CSS. We can reopen and merge this, but please since this is a fundamental template change, we should release it as part of 9.2.

squidfunk:
I've changed the base branch to `merge/pirir-piri` – Could you please resolve the conflicts and go through [the build process](https://squidfunk.github.io/mkdocs-material/customization/#theme-development), so we can merge it? You only made changes in `src`, but those need to be in `material` as well.

squidfunk:
Okay, some more thoughts on this: we shouldn't use Jinja's whitespace syntax for this – it will break too easily when we or authors change the template and we'll be facing this problem again. We should, in fact, set `display: inline-flex` and set `gap` to `4px` or `8px` to replicate the current behavior, as it's much more stable than managing whitespace in the template. As a project of this size, we try to make things as resilient as possible, while providing maximum flexibility.

If you're up to it, you can submit a PR ☺️

sisp:
Good idea! :+1: Sure thing, I'll send a PR.

PR #5852, opened 2023-08-17
Use flex gaps instead of whitespaces between social icons

sisp:
I've made the `.md-social` container an inline-flex container with a gap to replace the whitespace gaps between the social icons by proper gap margins.

Supersedes #5848.


squidfunk:
Thanks! LGTM!

PR #5853, opened 2023-08-17
Update required Node.js version in docs

sisp:
I've updated the required Node.js version in the docs to v18 as declared in `package.json`:

https://github.com/squidfunk/mkdocs-material/blob/ccc3c444edc66479074aa480b3afca56f17d38ed/package.json#L105-L107


squidfunk:
Thanks!

PR #5876, opened 2023-08-21
Improve development environment setup instructions

### LINKED ISSUE #5854: Improve development environment setup instructions
### Description

I've just gone through the [development environment setup](https://squidfunk.github.io/mkdocs-material/customization/#environment-setup) and stumbled over two aspects that might be improved:

1. **Usage of a virtual env:** Python dependencies should be installed in a virtual env, and this is probably clear to experienced Python developers but certainly not to everybody. I think the docs should advise to use a virtual env and show how to set it up.
2. **Setup of Node.js:** This project is _mainly_ a Python project (with regard to its build artifact) but requires Node.js for development. The docs just mention that Node.js is needed but don't explain how to set it up. Well, the same is true for Python, but since this is a Python project, it's probably okay to assume contributors have a working Python environment. But for Node.js, we could leverage [`nodeenv`](https://github.com/ekalinin/nodeenv) to install Node.js within the virtual env and, thus, ease the setup.

See below how I suggest to extend the setup instructions.

### Related links

- [Environment setup](https://squidfunk.github.io/mkdocs-material/customization/#environment-setup)
- [venv — Creation of virtual environments](https://docs.python.org/3/library/venv.html)
- [nodeenv](https://github.com/ekalinin/nodeenv)

### Proposed change

I think we could extend `docs/customization.md` as follows:

````diff
 Next, all dependencies need to be installed, which is done with:

 ```
 cd mkdocs-material
+python -m venv .venv
+. .venv/bin/activate
 pip install -e .
 pip install mkdocs-minify-plugin
 pip install mkdocs-redirects
+pip install nodeenv
+python -m nodeenv -p -n lts
 npm install
 ```
````

This way, a virtual env is created and activated, and `nodeenv` installs the latest LTS version (currently v18) into the virtual env along with NPM, so a contributor doesn't need to worry about installing (the right version of) Node.js. The only minor problem I see with those instructions is that the command for activating the virtual env isn't the same across platforms.

WDYT, @squidfunk?

### Before submitting

- [X] I have read and followed the [documentation issue reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-docs-issue/).
- [X] I have attached the links to the described sections of [the documentation](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-docs-issue/#related-links)

sisp:
I've improved the development environment setup instructions to

1. use a Python virtual environment, and to
2. use [`nodeenv`](https://pypi.org/project/nodeenv/) to install Node.js into the Python virtual environment, which eases the setup of a recent Node.js version.

`nodeenv` is also [used by the popular `pre-commit`](https://github.com/pre-commit/pre-commit/blob/a1f1d1915646865be2fe84d04633ba964feb0ba0/setup.cfg#L24) project to install Node.js for Node.js-based pre-commit hooks.

Resolves #5854.

WDYT, @squidfunk?

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -235,20 +235,38 @@ directly in the source of the theme and recompile it.
 
 ### Environment setup
 
-In order to start development on Material for MkDocs, a [Node.js] version of
-at least 18 is required. First, clone the repository:
+First, clone the repository:
 
 ```
 git clone https://github.com/squidfunk/mkdocs-material
-```
-
-Next, all dependencies need to be installed, which is done with:
-
-```
 cd mkdocs-material
+```
+
+Next, create a new [Python virtual environment][venv] and
+[activate][venv-activate] it:
+
+```
+python -m venv .venv
+source .venv/bin/activate
+```
+
+  [venv]: https://docs.python.org/3/library/venv.html
+  [venv-activate]: https://docs.python.org/3/library/venv.html#how-venvs-work
+
+Then, install all Python dependencies:
+
+```
 pip install -e .
 pip install mkdocs-minify-plugin
 pip install mkdocs-redirects
+pip install nodeenv
+```
+
+Finally, install the [Node.js] LTS version into the Python virtual environment
+and install all Node.js dependencies:
+
+```
+python -m nodeenv -p -n lts
 npm install
 ```
 


squidfunk:
Thanks for your work! In the future, we might break this guide up in native installing a Node version via a Node version manager and Python's `nodeenv`, but it's already much better than we currently have. We plan to revisit the customization guide at some point in the future.

sisp:
Sounds good, happy to help improving those docs further. :slightly_smiling_face:

PR #5877, opened 2023-08-21
Extract header options into partials

### LINKED ISSUE #5858: Extract header options into partials
### Context

I'm working on a customization that requires more than pure stylistic changes. In my case, the header options (color palette and language selector) and search form need to be moved to from the header nav to the tabs nav. I'm trying to achieve this with the least amount of code duplication as any copied code from upstream means technical debt.

### Description

I suggest to extract the header options ([color palette](https://github.com/squidfunk/mkdocs-material/blob/9.1.21/src/partials/header.html#L74-L110) and [language selector](https://github.com/squidfunk/mkdocs-material/blob/9.1.21/src/partials/header.html#L112-L140)) into partials, so I can include them (in my case) in my `tabs.html` override. To hide those components in the header nav, I could either apply `display: none` to their containers, or `mkdocs-material` could additionally wrap them in a Jinja [block which I can override](https://squidfunk.github.io/mkdocs-material/customization/#overriding-blocks) with empty content.

### Related links

- [Requesting a change](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/)
- [Extending the theme](https://squidfunk.github.io/mkdocs-material/customization/#extending-the-theme)

### Use Cases

This change would improve structural theme customization beyond style changes that require rendering the mentioned components in a different place.

### Visuals

_No response_

### Before submitting

- [X] I have included relevant links to [the documentation](https://squidfunk.github.io/mkdocs-material), related [issues](https://github.com/squidfunk/mkdocs-material/issues), and [discussions](https://github.com/squidfunk/mkdocs-material/discussions) to underline the need for my idea.

sisp:
I've extracted the header options markup into partials and include those in place of the inline markup.

@squidfunk Do you think the Jinja conditions should be moved into the partials as well? And what do you think about wrapping the `{% include %}` statement into a Jinja block, so a derived theme could override it, e.g., with empty content to disable rendering the component there?

Resolves #5858.

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -100,6 +100,7 @@ assets may also be put in the `overrides` directory:
 │  │  └─ analytics.html                # Analytics setup
 │  ├─ languages/                       # Translation languages
 │  ├─ actions.html                     # Actions
+│  ├─ alternate.html                   # Translation language selector
 │  ├─ comments.html                    # Comment system (empty by default)
 │  ├─ consent.html                     # Consent
 │  ├─ content.html                     # Page content
@@ -113,6 +114,7 @@ assets may also be put in the `overrides` directory:
 │  ├─ nav.html                         # Main navigation
 │  ├─ nav-item.html                    # Main navigation item
 │  ├─ pagination.html                  # Pagination (used for blog)
+│  ├─ palette.html                     # Color palette toggle
 │  ├─ post.html                        # Blog post excerpt
 │  ├─ search.html                      # Search interface
 │  ├─ social.html                      # Social links
--- a/docs/setup/setting-up-a-blog.md
+++ b/docs/setup/setting-up-a-blog.md
@@ -775,10 +775,10 @@ The following configuration options are available for author info:
     ``` yaml
     plugins:
       - blog:
-          authors_file: .authors.yml
+          authors_file: blog/.authors.yml
     ```
 
-    The path must be defined relative to [`blog_dir`][this is configurable].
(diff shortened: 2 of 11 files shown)


squidfunk:
Converting to draft, as you still seem to be working on this...

sisp:
Sorry for the confusion. I realized a bit after the first commit that I was missing the copyright headers in the partials and the listing of those files in the docs. Should be ready for a review now.

squidfunk:
Okay great! I changed the base to `merge/piri-piri`, as this is our current development branch which will shortly be released.

sisp:
Would you like me to rebase to `merge/piri-piri`?

squidfunk:
No, it's fine, I don't think thats necessary.

squidfunk:
LGTM! I think we can ship this with 9.2 and see how it goes.

> Do you think the Jinja conditions should be moved into the partials as well?

I checked all partials again, and no, we should keep conditions outside of the partials. I've adjusted the `tags` partial to behave like that, because partials should, in general, be dumb. If a partial is rendered should be decided on the site of inclusion, as that allows for more flexibility. One exception is the feedback partial, which checks if analytics is configured.

> And what do you think about wrapping the {% include %} statement into a Jinja block, so a derived theme could override it, e.g., with empty content to disable rendering the component there?

You cannot define blocks in partials. If it would work, we would've done it long ago.

squidfunk:
BTW, great work! Bonus points for updating the docs – that is often forgotten. Happy to collaborate on further PRs! 🚀

PR #5878, opened 2023-08-21
Add additional pymdownx.snippets url options

nofoxsteven:
|Option|Description|
|:-|:-|
|url_max_size|Sets an arbitrary max content size. If content length is reported to be larger, and exception will be thrown. Default is ~32 MiB.|
|url_timeout|Passes an arbitrary timeout in seconds to URL requestor. By default this is set to 10 seconds.|
|url_request_headers|Passes arbitrary headers to URL requestor. By default this is set to empty map.|

See: https://facelessuser.github.io/pymdown-extensions/extensions/snippets/#options


squidfunk:
Thanks! LGTM.

PR #5898, opened 2023-08-26
setting-up-versioning.md: explain "stay on the same page" feature

ilyagr:
While the feature is supposed to work out of the box, there are a few non-obvious conditions required for it to function.


ilyagr:
Feel free to use this as a documentation bug report instead of using my text. I wonder if I missed some conditions, e.g. depending on the JS API used, perhaps the site must be retrieved over HTTPS (encrypted) for the feature to work.

squidfunk:
Thanks!

ilyagr:
Thank you!

PR #5900, opened 2023-08-26
Update publishing-your-site.md GitHub Actions template

twangodev:
Added full fetch depth during checkout action.

Originally, this would result in a force push to the `gh-pages` branch of a repository. I've noticed that by adding `fetch-depth: 0` (the full repository history is retrieved), mkdocs will simply create a new commit, preserving the history of `gh-pages` branch.

This change does not conflict with the `--force` flag when `mkdocs gh-deply --force` is run.

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -35,6 +35,8 @@ contents:
         runs-on: ubuntu-latest
         steps:
           - uses: actions/checkout@v3
+            with:
+              fetch-depth: 0
           - uses: actions/setup-python@v4
             with:
               python-version: 3.x


squidfunk:
Thanks for the PR. However, I'm not sure I understand the entire context. You'll need this when using something like `git-revision-date` to properly retrieve the creation dates of files.

twangodev:
Yes, `fetch-depth: 0` is required for plugins that require information about the git repository.

However, information/history from the `gh-pages` branch isn't retrieved at all with the default [checkout action](https://github.com/actions/checkout)

> Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set fetch-depth: 0 to fetch all history for all branches and tags. Refer here to learn which commit $GITHUB_SHA points to for different event. 

Since mkdocs uses the [ghp-import](https://github.com/c-w/ghp-import) tool as mentioned [here](https://www.mkdocs.org/user-guide/deploying-your-docs/#project-pages), it by default should just create a new commit on the `gh-pages` branch. Therefore by fetching the history for `gh-pages`, a force push isn't strictly necessary.

This aims to avoid force pushing to the `gh-pages` branch every time the GitHub action is run, and instead create a new commit. This way, the history for the `gh-pages` branch is preserved, rather than being wiped by the force push.

squidfunk:
> This aims to avoid force pushing to the gh-pages branch every time the GitHub action is run, and instead create a new commit. This way, the history for the gh-pages branch is preserved, rather than being wiped by the force push.

I'm not sure this is desirable. To me, it isn't, because the `gh-pages` branch is essentially a storage for artifacts, so there's no need to go back to a previous version. Adding new commits all the time would mean that the repository would grow with every deployment with no apparent upside (at least to me), because AFAIK you can't deploy a previous version. 

Thus, I'm not sure we want to generally recommend to set `fetch-depth: 0`. It slows down builds, as it needs to pull all git history, which might be quite a lot (as it is for our repository). Plugins that need the history should mention this necessity to the user in the documentation. We should recommend the absolute minimum that is necessary for publishing.

Issue #5910, opened 2023-08-28
Questions regarding development steps


sisp:
### Description

I've stumbled over some issues when contributing to `mkdocs-material` and following the [theme development instructions](https://squidfunk.github.io/mkdocs-material/customization/#theme-development). When I [set up the environment](https://squidfunk.github.io/mkdocs-material/customization/#environment-setup) and then, without modification of the sources, [build the theme](https://squidfunk.github.io/mkdocs-material/customization/#building-the-theme), I see the following `git status` output:

```console
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    material/.overrides/assets/javascripts/custom.149a65e0.min.js
	deleted:    material/.overrides/assets/javascripts/custom.149a65e0.min.js.map
	deleted:    material/.overrides/assets/javascripts/iconsearch_index.json
	deleted:    material/.overrides/assets/stylesheets/custom.f7ec4df2.min.css
	deleted:    material/.overrides/assets/stylesheets/custom.f7ec4df2.min.css.map
	deleted:    material/.overrides/home.html
	deleted:    material/.overrides/hooks/translations.html
	deleted:    material/.overrides/hooks/translations.py
	deleted:    material/.overrides/main.html

```

Since I made no changes, I would have expected an empty change list.

When I run `npm run build:all` instead, I get this:

```console
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    material/.overrides/assets/javascripts/iconsearch_index.json
```

Why has his file been deleted? :thinking:

And when I make a small change in, e.g., `src/assets/stylesheets/main/integrations/_mermaid.scss` and `src/assets/javascripts/components/content/mermaid/index.css` (while working on #5824) and run `npm run build:all`, I get this:

```console
$ git add .
$ git status
On branch feat/mermaid-sequence-diagram-styles
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	deleted:    material/.overrides/assets/javascripts/iconsearch_index.json
	renamed:    material/assets/javascripts/bundle.78eede0e.min.js -> material/assets/javascripts/bundle.db3e830e.min.js
[...]

PR #5919, opened 2023-08-29
Update development environment setup documentation

sisp:
I've updated the documentation about setting up the development environment according to https://github.com/squidfunk/mkdocs-material/issues/5910#issuecomment-1696993538 and another minor simplification for running `nodeenv`.


squidfunk:
Thanks!

PR #5927, opened 2023-08-30
Fix minor typos

omahs:
Fix minor typos


squidfunk:
Tahnk yuo fro teh fxies!

PR #5937, opened 2023-08-31
Work around annotation icon rotation bug on Firefox

sisp:
I've added a workaround for a [Firefox bug that renders a subpixel outline when rotating a mask image element](https://bugzilla.mozilla.org/show_bug.cgi?id=1671784) &ndash; in this case the annotation icon:

image


src/assets/stylesheets/main/components/_tooltip.scss
```diff
@@ -162,6 +162,12 @@
       transition: z-index 250ms;
     }
 
+    // Hack: Work around Firefox bug that renders a subpixel outline when
+    // rotating a mask image element.
+    // https://bugzilla.mozilla.org/show_bug.cgi?id=1671784
+    overflow: hidden; // stylelint-disable-line
```
squidfunk, line 168: Why does stylelint need to be disabled here?

squidfunk (commented):
Thanks for the PR! Just a minor question ☺️

src/assets/stylesheets/main/components/_tooltip.scss
sisp: Because it complains otherwise:

```
Expected "overflow" to come before "outline"  order/properties-order
```

And I thought it's good to keep this hack separate from other styles, starting with a comment, to remember later that this is a hack and to ease its removal at some point when Firefox has fixed the bug and enough time has passed to justify dropping support for an old buggy Firefox version.

squidfunk: We should move it into the right location and keep the comment above the property, exactly as we do it with other hacks we have in place, e.g.:

https://github.com/squidfunk/mkdocs-material/blob/8f1433fa3916d5f7c75186e1ae262b9f1ec802a6/src/assets/stylesheets/main/components/_base.scss#L29-L37

sisp: I see. But then the two relevant styles are not collocated, so the comment needs to be duplicated. Is this how you would like it?

squidfunk: Valid point. Then please [only disable the specific rule with stylelint](https://stylelint.io/user-guide/ignore-code/), not the entirety of stylelint, as we'd be losing linting of other errors otherwise.

sisp: Good idea. I've pushed an update as you suggested.

squidfunk:
Thanks!

PR #5966, opened 2023-09-04
Document slug feature for blog pages

timvink:
I am looking to migrate my blog, and I want to keep the same URL for my blogposts. AFAIK that's not possible because there will still be the `<blog>` part in the url  `/blog/<slug>`, but while looking at this I noticed you can set a custom `slug`:

https://github.com/squidfunk/mkdocs-material/blob/86d52b93cddfe1cb2ce7c6bef1100a55acd7e804/src/plugins/blog/plugin.py#L769-L774

This is quite useful for people migrating & is undocumented. If you want to have this documented I can polish the text a bit further. You might have a good reason for _not_ documenting this yet, so feel free to close the PR also, no worries!

### DIFF
--- a/docs/setup/setting-up-a-blog.md
+++ b/docs/setup/setting-up-a-blog.md
@@ -1177,6 +1177,19 @@ linked with their titles.
   [built-in tags plugin]: setting-up-tags.md#built-in-tags-plugin
   [tags index]: setting-up-tags.md#adding-a-tags-index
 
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks! Definitely a good addition. We might change wording a little, as we're currently reworking all setup guides, but for now this is perfectly suitable.

PR #6015, opened 2023-09-14
Add CI Configuration part for `enable creation date` in `git-revision-date-localized` plugin

mendax1234:
I am not sure whether it is suitable to use `note` here, but if not, let me know and it can be changed at any time before merging.
Also, l think a simple example here would be better. However, l think doing another complete section about adding `CI workflow` to your documentation site would be much better. Since l am not an expert at the `CI workflow`, l think maybe l can just link to the `CI workflow configuration` part if it is done.

### DIFF
--- a/docs/conventions.md
+++ b/docs/conventions.md
@@ -41,6 +41,11 @@ like the site language, repository provider, or other settings.
 Some properties do not contain default values. This means that the functionality
 that is associated with them is not available unless explicitly enabled.
 
+### <!-- md:flag metadata --> – Metadata property { #metadata data-toc-label="Metadata property" }
+
+This symbol denotes that the thing described is a metadata property, which can
+be used in Markdown documents as part of the front matter definition.
+
 ### <!-- md:feature --> – Optional feature { #feature data-toc-label="Optional feature" }
 
 Most of the features are hidden behind feature flags, which means they must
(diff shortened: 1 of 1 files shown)


alexvoss:
Thanks for this. Not everything is using a CI system, though, so perhaps add a sentence that says that this is needed if people are deploying through a CI system? (M4MkDocs is used by a very wide range of users.)

docs/setup/adding-a-git-repository.md
```diff
@@ -229,6 +229,110 @@ The following configuration options are supported:
           enable_creation_date: true
     ```
 
+    !!! note
+        If you are deploying through a CI system, please do the CI configuration as described in the [git-revision-date-localized](https://github.com/timvink/mkdocs-git-revision-date-localized-plugin#note-when-using-build-environments) documentation.
+
+        [With Github Actions](../publishing-your-site.md#with-github-actions):
+
+        === "Material for MkDocs"
```
squidfunk, line 237: Please don't copy our entire CI setup guide, as we'd now have two places where we have to maintain it. Please just clarify with a sentence that when building in a CI environment, you might need to adjust the setting when fetching the code. That's exactly what @alexvoss mentioned. Thanks.

squidfunk (changes requested):

docs/setup/adding-a-git-repository.md
mendax1234: Ok, thanks! I will change it.

PR #6019, opened 2023-09-15
Add CI Configuration part for `enable creation date` in `git-revision-date-localized` plugin

mendax1234:
The Pull request is after this [PR](https://github.com/squidfunk/mkdocs-material/pull/6015). Since the former PR has conflicts, so l create a new one.


squidfunk:
Thanks!

mendax1234:
> Thanks!

Hi, I am just a bit curious about that in mkdocs, do we also use the 80-character limit in a line? And if I encounter a link that will exceed the limit, as long as it won't be shown, it can exceed in the editor?

Just like below:
image

I noticed that part of the link has exceeded the limit. But since it is a link, so it doesn't matter I think?

Thank you very much!

squidfunk:
Yes, we limit everything to 80 characters for better readability, except for links which are not possible to limit.

Issue #6017, opened 2023-09-14
Code highlighting colours the whole line in --md-code-hl-color


mocdaniel:
### Context

I'm working on a freshly setup instance of Material for MkDocs and code highlighting seems to be off, compared to the results [shown in the docs](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#highlighting-specific-lines).

<img width="693" alt="strange code highlighting" src="https://github.com/squidfunk/mkdocs-material/assets/28064429/2434b4d4-350e-4879-bc93-937db90c5983">

### Bug description

Compared to the example from the docs, the result rendered on my end is

a) yellowish instead of blue

b) renders the whole line in `--md-code-hl-color` instead of just the left border and the remainder in `--md-code-hl-color` according to my dev tools:

```css
.highlight .hll {
    background-color: var(--md-code-hl-color);
    display: block;
    margin: 0 -1.1764705882em;
    padding: 0 1.1764705882em;
}
```

My related configuration looks like this, I went by the book, i.e. the docs, if not mistaken:

```yaml
markdown_extensions:
  - abbr
  - admonition
  - attr_list
  - md_in_html
  - pymdownx.details
  - pymdownx.highlight:
      use_pygments: true
      pygments_lang_class: true
      line_spans: __span
  - pymdownx.snippets
  - pymdownx.superfences
  - pymdownx.tabbed:
      alternate_style: true
```

### Related links

I took a good look at the `pygments`/`Highlight` configuration:
- [https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown-extensions/#highlight](https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown-extensions/#highlight)

as well as the docs regarding how to do highlighting and how to theme it:
- [https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#highlighting-specific-lines](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#highlighting-specific-lines)
- [https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#custom-syntax-theme](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#custom-syntax-theme)

I managed to get the bluish highlight working this way, but it still covered the whole line in the darker variant and omitted the `--light` variant altogether.

### Reproduction

[...]

PR #6024, opened 2023-09-15
fix: unwanted divergence public vs. insider #6017

alexvoss:
Hope this fits the bill. 

One question: I did build and I did note the instruction to commit the compiled output in CONTRIBUTING.md but I feel somewhat shy to push the compiled code as well as the source. Am I being too cautious? Happy to amend if you do feel that is the right approach.


squidfunk:
Thanks!

PR #6091, opened 2023-09-26
Enabled linting of MkDocs YAML config

vedranmiletic:
This might be controversial, so please bear with me.

YAML indentation rules are:

- 2 spaces per level
- optionally, arrays can have extra 2 space indentation

According to these rules, however you take the second one, this is wrong

``` yaml
nav:
  - Home: index.md
  - Getting started:
    - Installation: getting-started.md
    - Creating your site: creating-your-site.md
```

and should be either (`yaml.org`  website style)

``` yaml
nav:
- Home: index.md
- Getting started:
  - Installation: getting-started.md
  - Creating your site: creating-your-site.md
```

or

``` yaml
nav:
  - Home: index.md
  - Getting started:
      - Installation: getting-started.md
      - Creating your site: creating-your-site.md
```

I have to admit that it did throw me off at first, but it started to make sense soon afterward: namely, the dashes are always vertically four spaces separated from each other. (For the top-level keys, one could imagine dash being at character position -2 with space at character position -1.)

I like the second option more so I prepared the pull request accordingly. The configuration in the pull request could be adjusted to enforce the first option instead.

Anyhow, if you would agree that this style should be followed, this pull request updates `mkdocs.yml` to pass the linter checking and adds `yamllint` syntax checking during documentation deployment. The config for `yamllint` uses the default ruleset except for long lines, where a warning is issued instead of an error.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -57,6 +57,12 @@ jobs:
             pyproject.toml
             requirements.txt
 
+      - name: Install yamllint
+        run: pip install yamllint
+
+      - name: Check MkDocs YAML configuration
+        run: yamllint mkdocs.yml
+
       - name: Set up build cache
         uses: actions/cache/restore@v3
         with:
--- a/.yamllint.yaml
+++ b/.yamllint.yaml
@@ -1,0 +1,5 @@
+extends: default
+
+rules:
+  line-length:
+    level: warning
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -168,92 +168,92 @@ markdown_extensions:
 nav:
   - Home: index.md
   - Getting started:
-    - Installation: getting-started.md
-    - Creating your site: creating-your-site.md
-    - Publishing your site: publishing-your-site.md
-    - Customization: customization.md
-    - Conventions: conventions.md
-    - Browser support: browser-support.md
-    - Enterprise feedback: enterprise-support.md
-    - Philosophy: philosophy.md
-    - Alternatives: alternatives.md
-    - License: license.md
-    - Changelog:
-      - changelog/index.md
-      - How to upgrade: upgrade.md
-    - Contributing:
-      - contributing/index.md
-      - Reporting a bug: contributing/reporting-a-bug.md
-      - Reporting a docs issue: contributing/reporting-a-docs-issue.md
-      - Requesting a change: contributing/requesting-a-change.md
-      - Adding translations: contributing/adding-translations.md
-      - Asking a question: https://github.com/squidfunk/mkdocs-material/discussions
-    - Guides:
-      - Creating a reproduction: guides/creating-a-reproduction.md
-    - FAQ:
-      - Sponsoring: faq/sponsoring.md
+      - Installation: getting-started.md
+      - Creating your site: creating-your-site.md
+      - Publishing your site: publishing-your-site.md
+      - Customization: customization.md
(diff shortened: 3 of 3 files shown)


squidfunk:
Thanks for suggesting. Phew, I'm a fan of either of those options. Both look odd to me. In fact, I just adapted style of indentation from MkDocs itself, [which uses the same style that we use](https://www.mkdocs.org/user-guide/writing-your-docs/#configure-pages-and-navigation) (from their docs):

``` yaml
nav:
  - Home: 'index.md'
  - 'User Guide':
    - 'Writing your docs': 'writing-your-docs.md'
    - 'Styling your docs': 'styling-your-docs.md'
  - About:
    - 'License': 'license.md'
    - 'Release Notes': 'release-notes.md'
```

I'm not sure we want or need something like YAMLLint. As far as I understand, it will just enforce style, not schema, and the build will already fail if `mkdocs.yml` doesn't comply with what MkDocs expects. So what's the upside of using YAMLLint?

alexvoss:
I may be holding the wrong end of the stick here but this would not produce any change for the users of Material for MkDocs, right? Only effect would be to check the `mkdocs.yml` of the documentation. Not sure much is gained here but @squidfunk would have an additional constraint to keep in mind.

vedranmiletic:
> Thanks for suggesting. Phew, I'm not a fan of either of those options. Both look odd to me.

I had the same thinking, but it started to make sense after looking at `yaml.org` style and the results produced by `yamlfmt` and similar tools.

> In fact, I just adapted the style of indentation from MkDocs itself, [which uses the same style that we use](https://www.mkdocs.org/user-guide/writing-your-docs/#configure-pages-and-navigation) (from their docs):
> 
> ```yaml
> nav:
>   - Home: 'index.md'
>   - 'User Guide':
>     - 'Writing your docs': 'writing-your-docs.md'
>     - 'Styling your docs': 'styling-your-docs.md'
>   - About:
>     - 'License': 'license.md'
>     - 'Release Notes': 'release-notes.md'
> ```

I figured. This style is something in between the two and doesn't fit either of the rules.

> I'm not sure we want or need something like YAMLLint. As far as I understand, it will just enforce style, not schema, and the build will already fail if `mkdocs.yml` doesn't comply with what MkDocs expects. So what's the upside of using YAMLLint?

Same as with other linters, automated checking of style. Schema validation is a separate thing, that can be done with `yamale`.

> I may be holding the wrong end of the stick here but this would not produce any change for the users of Material for MkDocs, right?

No, unless they use `mkdocs.yml` from the repo as an inspiration. It would be formatted slightly differently.

> Only effect would be to check the `mkdocs.yml` of the documentation. Not sure much is gained here but @squidfunk would have an additional constraint to keep in mind.

Well, the gain is the same as with linters for other languages which are already used (eslint, stylelint), you get automated style checking of the code and don't have to worry about doing it manually.

squidfunk:
As a further clarification: this does not impact anybody who wants to use this in their project – we just don't want to use it in our project, but you can always install YAMLLint and conform to the style it dictates.

vedranmiletic:
> Thanks for clarifying, @vedranmiletic . I'm actually a linter-skeptic. I use linters that enforce specific rules that you install as a safety net (e.g. no uncaught Promises in JS), but I never (up to now) used rules that force me into using some specific level of indentation or formatting.

Oh, I did not know that. Then it makes sense. Thanks anyhow for considering my contribution.

PR #6093, opened 2023-09-27
Fix a broken link in the documentation

iBug:
Fixes this link:

> image

I'm not sure what the originally intended link target is, so I replaced it with one that I find suitable.


squidfunk (changes requested):
Thanks for spotting and fixing! Could you change the link to point here? It's clearer then that this is the default value from which it is sourced:

https://squidfunk.github.io/mkdocs-material/setup/changing-the-fonts/#regular-font

squidfunk (approved):

squidfunk:
Thanks!

PR #6095, opened 2023-09-27
added instructions for checking out specific releases

alexvoss:
Added instructions for checking out specific releases and also moved the bit about $GH_TOKEN to a place above the pip install instructions - IMHO it did not make sense to have it after where it is needed. Hope this fits the bill.

### DIFF
--- a/docs/insiders/getting-started.md
+++ b/docs/insiders/getting-started.md
@@ -30,20 +30,31 @@ your GitHub account in order to access the Insiders repository programmatically
   [Generate a new token]: https://github.com/settings/tokens/new
   [scopes]: https://docs.github.com/en/developers/apps/scopes-for-oauth-apps#available-scopes
 
+Some of the instructions below require that the `GH_TOKEN` environment
+variable is set to the value of the personal access token you
+generated in the previous step. Note that the personal access token
+must be kept secret at all times, as it allows the owner to access
+your private repositories.
+
 ## Installation
 
 ### with pip
 
-Material for MkDocs Insiders can be installed with `pip`:
+Material for MkDocs Insiders can be installed with `pip`. You will
+normally want to install the latest release but can also install a
+specific older release. Pick the corresponding tag from the [list of
+tags] for the Insiders repository. In the `pip` command below, replace
+the tag with the one you want to install and make sure to have the
+`GH_TOKEN` variable set as instructed above.
 
 ``` sh
-pip install git+https://${GH_TOKEN}@github.com/squidfunk/mkdocs-material-insiders.git
+pip install git+https://${GH_TOKEN}@github.com/squidfunk/mkdocs-material-insiders.git@9.4.2-insiders-4.42.0
 ```
 
-The `GH_TOKEN` environment variable must be set to the value of the personal
-access token you generated in the previous step. Note that the personal access
-token must be kept secret at all times, as it allows the owner to access your
-private repositories.
+To install the latest development version, simply remove the last `@`
+symbol in the command and everything that follows it.
+
+[list of tags]: https://github.com/squidfunk/mkdocs-material-insiders/tags
 
 ### with docker
 


docs/insiders/getting-started.md
```diff
@@ -30,20 +30,31 @@ your GitHub account in order to access the Insiders repository programmatically
...
+variable is set to the value of the personal access token you
+generated in the previous step. Note that the personal access token
+must be kept secret at all times, as it allows the owner to access
+your private repositories.
+
 ## Installation
 
 ### with pip
 
-Material for MkDocs Insiders can be installed with `pip`:
```
squidfunk, line 37: I think we should move this into content tabs, similar to what we did in our getting started guide:
https://squidfunk.github.io/mkdocs-material/getting-started/#with-pip-9x

squidfunk (changes requested):

docs/insiders/getting-started.md
alexvoss: You mean the complete pip/Docker/git alternatives?

squidfunk: No, the "Latest / 9.x" content tabs ☺️

Bildschirm­foto 2023-09-27 um 20 36 31

alexvoss: Funny, that is what I had at first then then thought that it was not necessary. Will re-create that version for consistency's sake. Updating in a mo

alexvoss: I am also looking at the instructions for upgrading, adding instructions for git there as well. Got the text and will go about testing the instructions before committing. Is there anything else that should be done while I am at it?

squidfunk: From the top of my head:

- Installing Insiders Latest
- Installing Insiders specific tag
- Upgrading Insiders

We should probably adjust the structure to mirror what we have here:
https://squidfunk.github.io/mkdocs-material/changelog/

- Changelog (section)
- How to upgrade

alexvoss: Like so?

image

alexvoss: Ok, pushed those changes, hoping they match what you imagined. Tested the upgrade (and downgrade!) instructions and they should work. You may want to sanity check. 

On thing I note is that for the `git` version there does not seem to be a need to specify `--upgrade`, presumably because of the `-e`? I would leave it out in case it gets in the way of a downgrade. I am not that deeply familiar with how `pip` works in cases like this.

docs/insiders/upgrade.md
```diff
@@ -0,0 +1,69 @@
+# How to upgrade
+
+When upgrading Insiders, you should always check the version of Material for
+MkDocs which makes up the first part of the version qualifier, e.g.Insiders
+`4.x.x` is currently based on `8.x.x`:
+
+```
+8.x.x-insiders-4.x.x
```
squidfunk, line 8: We're already at 9.x.x 😉

docs/insiders/upgrade.md
```diff
@@ -0,0 +1,69 @@
...
+    creates a branch with the tag as a name - to avoid a warning by
+    `git` that you are in a "detached head" state but also to make it
+    easy for you to switch between different versions.
+
+    Now, change back to the parent directory in which your Git
+    repository lives and run `pip`:
+
+    ```
+    cd .. 
+    pip install -e --upgrade mkdocs-material
```
squidfunk, line 66: Is `--upgrade` necessary here? I'd assume that `pip` doesn't check versions and will always install what you give it once you set `-e` (which means "editable")

mkdocs.yml
```diff
@@ -253,8 +253,8 @@ nav:
     - Tooltips: reference/tooltips.md
   - Insiders:
     - insiders/index.md
-    - Getting started:
-      - Installation: insiders/getting-started.md
-      - Changelog: insiders/changelog.md
+    - Getting started with Insiders: insiders/getting-started.md
+    - Changelog: insiders/changelog.md
```
squidfunk, line 257: Could we put this into a section, similar to how we have it in our setup guide?

``` yaml
nav:
  ...
  - Changelog:
    - insiders/changelog/index.md # <- we'd need to rename the file, but URLs should stay the same
    - How to upgrade: insiders/upgrade.md
```

The structure would then be identical.

squidfunk (changes requested):

docs/insiders/upgrade.md
```diff
@@ -0,0 +1,69 @@
+# How to upgrade
+
+When upgrading Insiders, you should always check the version of Material for
+MkDocs which makes up the first part of the version qualifier, e.g.Insiders
```
squidfunk, line 4: Minor typo: "e.g.Insiders" - "e.g., Insiders"

docs/insiders/upgrade.md
```diff
@@ -0,0 +1,69 @@
...
+    this is done, you can run `pip` to install that version.
+
+    You can look up the tags using `git tag --sort -refname` or you
+    can consult the [list of tags]. Then, checkout the tag you want to
+    use by replacing the one given in the command below (twice)and running 
+    it from your workspace:
+
+    ``` 
+    cd mkdocs-material 
+    git checkout tags/9.4.2-insiders-4.42.0 -b 9.4.2-insiders-4.42.0
```
squidfunk, line 53: I think this can be simplified to:

```
git checkout 9.4.2-insiders-4.42.0
```

That's at least why I use when switching tags.

mkdocs.yml
```diff
@@ -253,8 +253,8 @@ nav:
     - Tooltips: reference/tooltips.md
   - Insiders:
     - insiders/index.md
-    - Getting started:
-      - Installation: insiders/getting-started.md
-      - Changelog: insiders/changelog.md
+    - Getting started with Insiders: insiders/getting-started.md
```
squidfunk, line 256: I'd drop "with Insiders" - we have it in the page title and in the menu it's redundant, because it's the first navigation item of the section that says "Insiders".

alexvoss:
Sorry, I made individual little commits here - can you squash them when merging the PR?

docs/insiders/upgrade.md
alexvoss: creating the new branch prevents the warning about the 'detached head'. It might also help a user to keep track of what releases they have looked at. The reason I included the `tags/` part is that this is how it is usually described online (yes, I had to look up how to check out a tagged version). It might be a toss-up whether starting with the simplest solution is better or starting with the more complicated one and letting people figure out they can simplify it? Happy to take your guidance here.

squidfunk:
Jup, we're going to squash the PR when we resolved the remaining conversation.

docs/insiders/upgrade.md
squidfunk: Thanks for the clarification! Yes, detached head state always sounds kind of scary, especially to git newcomers, but the question is what we're optimizing here for. IMHO, users will rarely need to push new commits into Insiders (which they will not be able to in detached head state), because I think they usually just intend to pin their Insiders version rather than basing more commits off that specific tag. Specifying the tag twice leaves more room for errors, as you need to adjust both tags, so I'm rather leaning towards keeping it simple. I'd say we go for simple now and adjust the instructions once we encounter users with problems.

However, quick googling lead me to:

```
git:master $ git switch -c 9.2.0                     
Switched to a new branch '9.2.0'
git:9.2.0 $ 
```

If you can confirm that this has the same outcome as you suggested, we can just use `switch` ☺️

squidfunk: For clarification: `switch` will create a new branch with the same name as the tag, based of the tag.

mkdocs.yml
squidfunk: Ah! Yes, we need to adjust the shortcodes hook as well – good catch!

https://github.com/squidfunk/mkdocs-material/blob/b77769009ece2326476d1a02d8fa04e501f73af0/src/overrides/hooks/shortcodes.py#L148

docs/insiders/upgrade.md
alexvoss: I think that the `-c` option does need the branch name and a starting point, so `tags/<tagname>` would still be needed. The `switch` command does seem preferable to `checkout`, though. Perhaps the instructions should avoid calling the branch the same as the tag. Another option is to use the `--detach` option on `checkout`, which avoids the warning.

squidfunk: Ah nice, then let's use `--detach`, if we can omit the repeated tag use.

alexvoss:
Oh. I did not mean to just push this through all the way to mkdocs-material. I should have asked for a re-review and let you do that. Given what I was working on, I don't imagine any harm was done?!

alexvoss:
@squidfunk, I managed to break the documentation build by not committing the files under `/material`. Will try to fix and apologies again for inadvertently pushing this through. Lack of experience with GitHub.

squidfunk:
No problem, I'll take care of it.

Issue #6105, opened 2023-10-01
Not clear how to disable `navigation.instant` for local pages


VelocityRa:
### Description

When using `navigation.instant`, the Table of Contents becomes un-scrollable. It only shows the top part, with no scrollbar. The bug is presumably only visible if there are many entries (more than a page view's worth, vertically). It goes away if I remove `navigation.instant` from my config.

I use both `gh-deploy` (for online docs), and ship built html files to my users (for local browsing).

More of a question (or bug report?) but, It's not clear how I would only use instant navigation when deploying to GH pages without hacks, like programmatically appending `- navigation.instant` to my `mkdocs.yml` before the `gh-deploy` on my CI runner.

### Related links

- https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#instant-loading
- https://www.mkdocs.org/user-guide/deploying-your-docs/#local-files

### Proposed change

One of:
- Fix instant loading on local pages
- Disable instant loading on local pages (loading locally is instant anyway)
- Document a way / workaround for disabling instant loading for `file://` browsing.

### Before submitting

- [X] I have read and followed the [documentation issue guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-docs-issue/).
- [X] I have attached the links to the affected sections of [the documentation](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-docs-issue/#related-links)

alexvoss:
The [building for offline usage page](https://squidfunk.github.io/mkdocs-material/setup/building-for-offline-usage/) says that instant navigation needs to be turned off, amongst other things. You may want to give the [group plugin](https://squidfunk.github.io/mkdocs-material/plugins/group/) a try. 

I realize the information is not conveniently in one place.  Would adding a line or two here have helped? https://squidfunk.github.io/mkdocs-material/creating-your-site/#building-your-site

squidfunk:
Thanks for reporting. Since you're deliberately building for offline usage, the [building for offline usage guide](https://squidfunk.github.io/mkdocs-material/setup/building-for-offline-usage/) is the canonical place for information around that topic. The page lists all features that are not compatible with offline usage, not because Material for MkDocs doesn't support them, but because browsers don't allow for the underlying implementation of those features to work. This is all expained in the [limitations](https://squidfunk.github.io/mkdocs-material/setup/building-for-offline-usage/#limitations) sections of the offline plugin.

[...]

PR #6116, opened 2023-10-02
 Not clear how to disable navigation.instant for local pages #6105 

alexvoss:
I have added a note about building for offline usage and a link to the building for offline usage page. 

That page could, in turn, do with an addition to show how the group plugin can be used to turn off functionality that will not work offline. I suggest developing this as an example. Happy to do it if you agree this is the right approach.

### DIFF
--- a/docs/creating-your-site.md
+++ b/docs/creating-your-site.md
@@ -225,3 +225,10 @@ or your private web space.
 
   [GitHub Pages]: publishing-your-site.md#github-pages
   [GitLab pages]: publishing-your-site.md#gitlab-pages
+
+If you intend to distribute your documentation as a set of files to be
+read from a local filesystem rather than a web server (such as in a
+`.zip` file), please read the notes about [building for offline
+usage].
+
+  [building for offline usage]: setup/building-for-offline-usage.md


squidfunk:
Thanks for the PR! I'm not sure this is the best location, because it's the basic getting up and running guide which you read when you first set up Material for MkDocs, and authors will likely dive deep into the setup, plugins, and reference section, forgetting about this. I think the offline guide is the best location for this, but we cannot cross reference the offline guide at each and every feature that does not work with it.

alexvoss:
I agree that people may forget about it but now at least it is there for those who consider an offline delivery format. 

"we cannot cross reference the offline guide at each and every feature that does not work with it" - agreed. That is why I suggest strengthening the offline guide by showing how it is done there. People may still ask why plugin/feature X does not work but now we have a place to point them to. Could also be an example to avoid blowing up the documentation more. Do with the PR what you think is right. I thought that having the link to the building for offline page was useful to have.

squidfunk:
Fair points. Let's add it and reconsider once we found a good way to better cluster our documentation into use cases ☺️

Issue #6085, opened 2023-09-26
Additional dependencies for development


alexvoss:
### Description

On the [page on setting up a development environment](https://squidfunk.github.io/mkdocs-material/customization/#environment-setup), two additional dependencies need to be added, I think:

```
pip install mkdocs-rss-plug
pip install 'mkdocs-material[imaging]'
```

After installing these, I can build the documentation again.

### Related links

- [Environment Setup](https://squidfunk.github.io/mkdocs-material/customization/#environment-setup)

### Proposed change

add those additional dependencies required to build the documentation (as part of testing that the theme is working)

### Before submitting

- [X] I have read and followed the [documentation issue guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-docs-issue/).
- [X] I have attached the links to the affected sections of [the documentation](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-docs-issue/#related-links)

alexvoss:
Ok, I see what I did there. I was trying to produce a pull request for functionality, so was using the Insiders edition. Should the development environment not support both? If so then we would also need a mention of `libcairo` and any other library dependencies. If the recommended dependencies do the trick then all the better!

squidfunk:
Yes, we can feature both, but must then also explain and link to the [image processing](https://squidfunk.github.io/mkdocs-material/plugins/requirements/image-processing/) requirements guide. You could just create content tabs for both editions, similar to what we're doing in our [publishing guide](https://squidfunk.github.io/mkdocs-material/publishing-your-site/#with-github-actions-material-for-mkdocs).

squidfunk:
For explanation:

```
Community
  recommended

Insiders
  recommended
  imaging
    libcairo
    pngquant
```

We should keep it as simple as possible and only install what's absolutely required.

squidfunk:
Resolved via 699097679313b6e1d8dc508ed9a147bc46ba9df9 by @alexvoss – love the new [environment setup guide](https://squidfunk.github.io/mkdocs-material/customization/#environment-setup)!

squidfunk:
By the way, one thing I noticed: this can likely be shortened to:

```
pip install -e ".[recommended]"
pip install nodeenv
```

... and:

```
pip install -e ".[recommended, imaging]"
pip install nodeenv
```

[...]

PR #6125, opened 2023-10-03
fix: Additional dependencies for development #6085 

alexvoss:
I added instructions for working on Insiders and also changed the instructions for installing optional dependencies to use the definitions in `pyproject.toml`. 

As I keep forgetting to activate a `venv` and end up installing things globally, I added a note about getting `pip` to refuse installing things outside a `venv`. Seems I am not the only person who keeps making that mistake, so I thought it would be useful for people here as many of our customers will not be hard-core Pythonistas.

Hope this fits the bill. I got stuck on this a little bit because I looked for ways to make things run on Windows but that is a battle we can fight another day, perhaps when we find someone who is native in that environment.

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -242,7 +242,7 @@ directly in the source of the theme and recompile it.
 First, clone the repository for the edition you want to work on. Note
 that you need to be a member of [Insiders] to access its repository.
 
-  [Insiders]: insiders.md
+  [Insiders]: insiders/index.md
 
 === "Material for MkDocs"
 
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -258,9 +258,15 @@ sponsor first to gain access.
     Insiders documentation] and make it available in the `$GH_TOKEN` 
     variable. 
 
+    ``` sh
+    git clone https://${GH_TOKEN}@github.com/squidfunk/mkdocs-material-insiders.git # (1)!
     ```
-    git clone https://${GH_TOKEN}@github.com/squidfunk/mkdocs-material-insiders.git
-    ```
+
+    1. If you are using SSH keys for authenticating with GitHub, you can
+       clone Insiders with this command:
+       ```
+       git clone git@github.com:squidfunk/mkdocs-material-insiders.git
+       ```
 
     [as described in the Insiders documentation]: insiders/getting-started.md#requirements
 
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -239,8 +239,9 @@ directly in the source of the theme and recompile it.
 
 ### Environment setup
 
-First, clone the repository for the edition you want to work on. Note
-that you need to be a member of [Insiders] to access its repository.
+First, clone the repository for the edition you want to work on. If
+you want to clone the Insiders repository, you need to become a
+sponsor first to gain access.
 
   [Insiders]: insiders/index.md
 
@@ -251,7 +252,7 @@ that you need to be a member of [Insiders] to access its repository.
     cd mkdocs-material
     ```
 
-=== "Insiders (with token)"
+=== "Insiders"
 
     You will need to have a GitHub access token [as described in the
     Insiders documentation] and make it available in the `$GH_TOKEN` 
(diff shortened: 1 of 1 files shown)


docs/customization.md
```diff
@@ -239,12 +239,38 @@ directly in the source of the theme and recompile it.
 
 ### Environment setup
 
-First, clone the repository:
+First, clone the repository for the edition you want to work on. Note
+that you need to be a member of [Insiders] to access its repository.
```
squidfunk, line 243: Have we clarified in this document that there are two editions? Maybe we should clarify it with an "if", because it sounds a little too exclusive. Something like "If you want to clone the Insiders repository, you need to become a sponsor first to gain access.

docs/customization.md
```diff
@@ -239,12 +239,38 @@ directly in the source of the theme and recompile it.
...
+    Insiders documentation] and make it available in the `$GH_TOKEN` 
+    variable. 
+
+    ```
+    git clone https://${GH_TOKEN}@github.com/squidfunk/mkdocs-material-insiders.git
+    ```
+
+    [as described in the Insiders documentation]: insiders/getting-started.md#requirements
+
+=== "Insiders (with SSH key)"
```
squidfunk, line 266: We're nowhere mentioning cloning with SSH in our docs, so could we move this into a code annotation and remove the "(with token)"? From my experience giving too many options to choose from leads to confusion. Let's recommend the token approach as the canonical way, as most users are used to that, since we advertise it on our docs. I just try to make less work for us 😅

docs/customization.md
```diff
@@ -254,17 +280,46 @@ python -m venv venv
 source venv/bin/activate
 ```
 
+!!! note "Get pip to refuse working outside a `venv`"
```
squidfunk, line 283: Very good idea, but I think the headline is a little misleading. Maybe something like:

Ensure pip always runs in a virtual environment

squidfunk (changes requested):
Great work! Only some minor changes.

docs/customization.md
```diff
@@ -263,15 +264,6 @@ that you need to be a member of [Insiders] to access its repository.
 
     [as described in the Insiders documentation]: insiders/getting-started.md#requirements
 
-=== "Insiders (with SSH key)"
```
squidfunk: I meant to put that into a code annotation, similar to. Sorry if this was unclear. It's definitely valuable information and we should mention it, but I'd regard it as side content to the interested reader:
https://squidfunk.github.io/mkdocs-material/getting-started/#with-pip-9x

alexvoss: Ah, sorry, you mean like so?
```
=== "Insiders"
    
    You will need to have a GitHub access token [as described in the
    Insiders documentation] and make it available in the `$GH_TOKEN`
    variable. 
  
    ``` sh
    git clone https://${GH_TOKEN}@github.com/squidfunk/mkdocs-material-insiders.git # (1)!
    ```

    1. If you are using SSH keys for authenticating with GitHub, you can
       clone Insiders with this command:
       ```
       git clone git@github.com:squidfunk/mkdocs-material-insiders.git
       ``` 
       
    [as described in the Insiders documentation]: insiders/getting-started.md#requirements
```

I had to remove the line before the code snippet in the annotation to force it to be rendered inside the annotation and not as a next line. The [documentation for the code annotation feature](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#adding-annotations) does not say anything about the rules for defining the annotation? Is this down to indentation? (may need to better learn Markdown's rules on that)

alexvoss: I've gone and pushed this version, hope it fits the bill.

squidfunk: > I had to remove the line before the code snippet in the annotation to force it to be rendered inside the annotation and not as a next line. The [documentation for the code annotation feature](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#adding-annotations) does not say anything about the rules for defining the annotation?

I think this happens because you need to indent by four spaces, not three, like so:

https://github.com/squidfunk/mkdocs-material/blob/5ba794405ff2781d57899d85eba3b0cda0e553c6/docs/getting-started.md?plain=1#L33-L55

squidfunk (approved):

PR #6127, opened 2023-10-03
Fix wrong documentation link to Built-in projects plugin

SimonMarquis:
See https://squidfunk.github.io/mkdocs-material/plugins/#management


squidfunk:
Thanks!

PR #6133, opened 2023-10-04
Small change in Danish translation

cfaagaard:


saltydk:
Should probably be 
```json
"readtime.other": "# minutters læsetid",
```

squidfunk:
Thanks for the PR! So what's it gonna be? I'm not proficient in Danish 😉

saltydk:
My suggestion is how we would write it in Danish when referring to multiple minutes, which I assume was the intent here.

squidfunk:
Im not sure how to resolve this. If you both could please discuss what makes sense and we could reach a conclusion, that would be amazing @saltydk @cfaagaard

saltydk:
My suggestion is the correct way if the intent is multiple minutes, which I assumed to be the case.

squidfunk:
Thanks @saltydk! Since OP hasn't responded, we go with your suggested improvement. Added in 7fa5ea060, which supersedes this PR.

cfaagaard:
Sorry. Been offline. I agree with @saltydk .

PR #6134, opened 2023-10-04
Allow hiding page title using front matter

ti-mo:
For special documents like landing pages, it can be necessary to hide the first `<h1>`. Unfortunately, this can currently only be done by overriding partials/content.html, which I'd rather not carry a copy of in my documentation.

I think `material/templates/partials/content.html` needs to be regenerated, but I couldn't immediately find how to do that.

### DIFF
--- a/docs/setup/setting-up-the-header.md
+++ b/docs/setup/setting-up-the-header.md
@@ -25,6 +25,26 @@ theme:
     - header.autohide
 ```
 
+### Hiding the page title
+
+<!-- md:version 9.4.4 -->
+<!-- md:flag metadata -->
+
+In documents missing a `<h1>` at the top of the page, Material for MkDocs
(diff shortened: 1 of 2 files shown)


squidfunk:
Thanks for the PR. However, this will only hide the auto-generated title, not the title you define in Markdown. IMHO, it's not consistent with what the expected authoring experience is, as it depends heavily on what the user writes. Please use CSS for that, it's trivial. Just add the following lines __to the Markdown of the page for which you want to hide the title__:

``` md
<style>
  .md-content .md-typeset h1 {
    display: none;
  }
</style>

Your Markdown content
```

Additionally, a page should always have a `h1` for SEO purposes. If you have this case repeatedly, you can use [Snippets](https://facelessuser.github.io/pymdown-extensions/extensions/snippets/).

ti-mo:
Thanks! That works, too. :)

PR #6136, opened 2023-10-05
added support for new entries in mkdocs.yml (to fix  #6129)

alexvoss:
`exclude_docs`, `not_in_nav` and `validation` were introduced in MkDocs Version 1.5

The schema for the `validation` field probably needs to be be looked at again at it covers only the simplest use case where `nav` and `links` are present.


alexvoss:
@squidfunk, will work on this while traveling, hope to have it complete tonight. Please do me the favour and have a look. One thing I was not sure about was whether the repetition of `"enum": ["warn", "info", "ignore"]` in `validation.json` was the right things to do. I tend to think that when a reference replaces a simple line then it is more confusing than it is worth but would take your lead on this. We could have a `warninglevels` definition, which I guess would also not be difficult to interpret.

docs/schema/validation.json
```diff
@@ -0,0 +1,44 @@
...
+  "title": "Validation settings",
+  "markdownDescription": "https://www.mkdocs.org/user-guide/configuration/#validation",
+  "type": "object",
+  "properties": {
+    "nav": {
+      "title": "Settings for validation of navigation",
+      "markdownDescription": "https://www.mkdocs.org/user-guide/configuration/#validation",
+      "properties": {
+        "omitted_files": {
+          "type": "string",
```
squidfunk, line 12: I think there's no need to set `type: string` when you set `enum` ☺️ All instances of settings where we define `enum` values don't have a `type `defined, so I suggest to drop it.

docs/schema/validation.json
```diff
@@ -0,0 +1,44 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema",
+  "title": "Validation settings",
+  "markdownDescription": "https://www.mkdocs.org/user-guide/configuration/#validation",
+  "type": "object",
+  "properties": {
+    "nav": {
+      "title": "Settings for validation of navigation",
```
squidfunk, line 8: Let's just drop the "Settings for", as I think it's redundant. Just "Validation of navigation". Same below.

squidfunk (changes requested):
Thanks! I'd say we should deduplicate it by using a `$ref`, because MkDocs defines them as a single type, and if that ever changes, we would have to remember to adjust 6 locations. Otherwise only very minor suggestions. Great work!

alexvoss:
Right, hope this fits the bill. One other thing I noticed needed doing was to add `"additionalProperties": false` in places to make sure mis-spelled stuff is actually caught

docs/schema/validation.json
```diff
@@ -0,0 +1,53 @@
...
+      "markdownDescription": "https://www.mkdocs.org/user-guide/configuration/#validation",
+      "additionalProperties": false,
+      "properties": {
+        "not_found": { "$ref": "#/$defs/not_found" },
+        "absolute_links": { "$ref": "#/$defs/absolute_links" },
+        "unrecognized_links": { "$ref": "#/$defs/unrecognized_links" }
+      }
+    }
+  },
+  "$defs": {
```
squidfunk, line 32: Please use `"definitions"` and not `$defs`, as we do in the rest of our schema, e.g. here:

https://github.com/squidfunk/mkdocs-material/blob/57e598b96e8a976eec0559de99ab700bf2d6c38e/docs/schema/theme.json#L457-L468

docs/schema/validation.json
```diff
@@ -0,0 +1,53 @@
...
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "omitted_files": { "$ref": "#/$defs/omitted_files" },
+    "not_found": { "$ref": "#/$defs/not_found" },
+    "absolute_links": { "$ref": "#/$defs/absolute_links" },
+    "nav": {
+      "title": "validation of navigation",
+      "markdownDescription": "https://www.mkdocs.org/user-guide/configuration/#validation",
+      "additionalProperties": false,
```
squidfunk, line 14: Please move `"additionalProperties"` below `"properties"`, as in the rest of our schema, e.g., here:

https://github.com/squidfunk/mkdocs-material/blob/57e598b96e8a976eec0559de99ab700bf2d6c38e/docs/schema/plugins/info.json#L36-L46

squidfunk (changes requested):
Almost there! Sorry to be so pedantic, but I'm trying very hard to keep it consistent 😉

docs/schema/validation.json
alexvoss: The reason I put it at the top is that it is actually important information for how what follows is to be interpreted.  Happy to follow your style, though. I think the bit about `$defs` is more important.

docs/schema/validation.json
squidfunk: You're right, I didn't know that 🤦‍♂️

> Although we can identify any subschema using JSON Pointers or named anchors, the $defs keyword gives us a standardized place to keep subschemas intended for reuse in the current schema document.

Could you open another PR then, renaming all `definitions` to `$defs`? Consistency is really important.

squidfunk (approved):

squidfunk:
Great, thanks!

PR #6146, opened 2023-10-06
Schema

alexvoss:
ok, 2nd attempt to get the "definitions" renamed to "$defs" and to add "unrecognized_links"

### DIFF
--- a/docs/schema/extensions/markdown.json
+++ b/docs/schema/extensions/markdown.json
@@ -73,10 +73,10 @@
           "type": "object",
           "properties": {
             "toc": {
-              "$ref": "#/definitions/toc"
+              "$ref": "#/$defs/toc"
             },
             "markdown.extensions.toc": {
-              "$ref": "#/definitions/toc"
+              "$ref": "#/$defs/toc"
             }
           },
           "additionalProperties": false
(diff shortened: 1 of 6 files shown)


squidfunk (approved):

squidfunk:
Thanks!

alexvoss:
Funny, GitHub again claims that my branch is ahead of the master even though its changes have been merged. It tells me that there is nothing to sync from squidfunk:master, so no way I can tell it to look again and figure out that my changes have been incorporated there. It tells me that I can delete the whole fork, which is not really what I want but did in the end because I did not trust what was going on behind the scenes at GitHub. 

I do wonder what the right approach is for doing a stream of PRs?

a) ignore this and see what happens when we try to do subsequent PRs
b) delete the branch once it is merged and open a new one for new work
c) delete the whole fork and start with a blank slate
d) try and find a way to resolve this that does not use the GitHub UI
e) invest effort to try and understand just how this works and find The Solution

I don't want to do another PhD but if it saves a lot of hassle downstream then (e) might be the best. If you have/know of any good resources for starting this please let me know. 

Otherwise, the pragmatic approach would be (b). I would sync the changes from your master branch into mine through the GitHub Web UI and then delete the work branch on GitHub and locally.

squidfunk:
IMHO, the best way is:

1. Pull changes from upstream
2. Create a new branch
3. Apply your changes
4. Create PR with branch in upstream

It's never a good idea to create a PR based on your master, because the master might be impacted by other branches or new upstream changes. Thus, for each PR, always create a new branch. Once your branch has been merged, pull master and create a new branch. If your branch hasn't been merged yet and you work on a commit that is based on an open PR, create a new branch from that branch.

PR #6172, opened 2023-10-11
Create Document how to build locally, stage a change for PR, and subm…

Youngemmy5956:
…it a PR for modifying the code.md

Adding my own contribution

### DIFF
--- a/Document how to build locally, stage a change for PR, and submit a PR for modifying the code.md
+++ b/Document how to build locally, stage a change for PR, and submit a PR for modifying the code.md
@@ -1,0 +1,39 @@
+Documenting the Global Packages Needed to Be Installed
+Before you can start modifying the code and submitting a PR, you'll need to have certain global packages installed on your system. These typically include Node.js and package managers like npm or yarn. You can document the required global packages like this:
+
+Node.js: Ensure you have Node.js installed. You can download it from the official website: Node.js. Verify the installation using node -v and npm -v.
+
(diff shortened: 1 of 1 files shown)


squidfunk:
cc @alexvoss

alexvoss:
Hi, thanks for getting in touch. As it happens, we just had discussions about how to improve the documentation for contributors. Please note there is existing documentation about [setting up a development environment](https://squidfunk.github.io/mkdocs-material/customization/#environment-setup). We need to start with this and evolve it to cover all the points you touch on in the document you sent (as well as possibly other ones). This will happen over the course of the next two weeks I think, at least I should make a good start on this soon. 

If you don't mind, I'd suggest I ask you to review the documentation before I ask @squidfunk to review it for inclusion in the project. That would be really helpful. If you find issues I have not addressed, you could add your own PRs based on what is there. 

@squidfunk, can we keep this open for a few days until I have managed to put up a PR for a skeleton new version of the existing docu?

squidfunk:
Well first, not to criticize the work of the OP, but it looks like something generated with ChatGPT or some other generative AI. It currently provides only limited value since it is completely generic and does not talk at all about our project. I tagged you, @alexvoss, because we talked about a PR guide, so this was just meant as an FYI, but coming back to this I realize that it only makes sense to _maybe_ use as a checklist.

So, thanks @Youngemmy5956, but this is too generic.

PR #6175, opened 2023-10-11
Bumped mermaid version to 10.5.0

TKFRvisionOfficial:


alexvoss:
Hi, have you discussed bumping Mermaid with @squidfunk? I cannot seem to find a discussion on this. Apologies if I have missed something. There would need to be a review as to whether this will not break things before a request to bump the version can be accepted.

TKFRvisionOfficial:
Yeah, so I screwed up. Buming over major versions will not work like 99% of the time, and I sadly haven't won the dependency lottery. 😂. Next time I'll test before publishing a pr... Im sorry

PR #6183, opened 2023-10-12
Escape snippet syntax in code block example

sisp:
I've fixed the [code block example in the docs which demonstrates how to embed external files through the snippet syntax](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#embedding-external-files). The snippet syntax was not [escaped](https://facelessuser.github.io/pymdown-extensions/extensions/snippets/#escaping-snippets-notation), so the plugin actually embedded the content of the referenced file.

### DIFF
--- a/docs/reference/code-blocks.md
+++ b/docs/reference/code-blocks.md
@@ -460,7 +460,7 @@ from within a code block:
 
 ```` markdown title="Code block with external content"
 ``` title=".browserslistrc"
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! I didn't know this was possible by now – perfect!

sisp:
I didn't know either, but then I stumbled over the `pymdownx` docs. :smile:

PR #6189, opened 2023-10-14
Install extras from local package for development environment setup

sisp:
I've fixed a documentation error in the [development environment setup](https://squidfunk.github.io/mkdocs-material/customization/#environment-setup) instructions about installing Python package extras. Before, the extras were installed from the latest _release_ which may lead to a broken development installation if the latest _development version_ had changed the extras in some way. Now, the extras shall be installed also from the latest development version.

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -301,16 +301,14 @@ Then, install all Python dependencies:
 === "Material for MkDocs"
 
     ```
-    pip install -e .
-    pip install "mkdocs-material[recommended]"
+    pip install -e ".[recommended]"
     pip install nodeenv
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks! [Had that in my inbox for a week now](https://github.com/squidfunk/mkdocs-material/issues/6085#issuecomment-1749118807), and now you did it. Perfect! 🚀

sisp:
Ah, I hadn't seen that issue. Good to know it had been on your list already.

alexvoss:
Thanks, that removed it from my guilt pile as well!

PR #6200, opened 2023-10-16
Use of document$ observable

alexvoss:
Added tip about using the `document$` observable to ensure the DOM is fully loaded. Hope I got the details right and that this is sufficient to give people the leg up they will need. One thing I am not sure about is whether we need to / should add a note about ordering of the invocations of the callback - or any other fatal assumptions people might make?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,5 +62,6 @@ example.zip
 TODO
 tmp
 
-# IDEs
+# IDEs & Editors
 .idea
+*~
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -54,6 +54,29 @@ extra_javascript:
   - javascripts/extra.js
 ```
 
+!!! tip "Make sure the page is fully loaded"
+
+    It is likely that you will want to run your JavaScript code only
+    once the page has been fully loaded by the browser. This means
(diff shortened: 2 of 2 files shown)


alexvoss:
Not sure why the commit re vim backup files ended up in this? That change is already in the master. I did sync my fork and pulled from it before making today's changes and committing them to a new branch.

squidfunk:
Did you also merge master into this `observable` branch?

alexvoss:
It is possible I was on the `gitignore` branch that I still have locally. So, I might have pulled with no effect on master. Got to remember to delete local branches when deleting them on GitHub! I seem to remember there is a way to delete a branch on GitHub from the command line. I need to script this stuff - just don't do this often enough and been working on some completely unrelated things over the weekend...

squidfunk:
Thanks, LGTM!

squidfunk:
I changed the title of the box to make it clearer. "Make sure the page is fully loaded" is not actionable enough, because IMHO, it's easier to anchor the user at what she wants to do, i.e., integrate a third-party library. Additionally I changed the wording of the demo integration to be a little nicer - IMHO "whatever" is not a really nice word 😉

PR #6202, opened 2023-10-16
docs(setup): add missing code annotation

robingenz:
It seems that the anchor is missing here.


squidfunk:
Good catch, jup! Thanks!

PR #6220, opened 2023-10-21
+doc: mention icon on custom page status and example

alexvoss:
Based on https://github.com/squidfunk/mkdocs-material/discussions/6208, I added a note that a custom tag that needs an icon will need customization in `extra.css` and provided a link to the example in the text. I was not sure what you meant by "dumbing down the example" - it is already minimal, no?

While we are at it, the README pages for examples probably should have the download links just as they appear in the documentation, right?

### DIFF
--- a/docs/reference/index.md
+++ b/docs/reference/index.md
@@ -130,10 +130,17 @@ status: new
 ...
 ```
 
-The following status identifiers are currently supported:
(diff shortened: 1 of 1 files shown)


squidfunk (approved):

squidfunk:
LGTM! Thanks!

PR #6222, opened 2023-10-21
Use admonition for documenting a tip

sisp:
I've moved a plain [tip regarding Python intro](https://squidfunk.github.io/mkdocs-material/getting-started/#with-pip) into a tip admonition which is more visually pleasing IMO and more consistent with the rest of the fantastic docs.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -75,10 +75,12 @@ MkDocs on GitHub Pages in a step-by-step guide.
 
 ---
 
-__Tip__: If you don't have prior experience with Python, we recommend reading
-[Using Python's pip to Manage Your Projects' Dependencies], which is a really
-good introduction on the mechanics of Python package management and helps you
-troubleshoot if you run into errors.
+!!! tip
+
+    If you don't have prior experience with Python, we recommend reading
+    [Using Python's pip to Manage Your Projects' Dependencies], which is a
+    really good introduction on the mechanics of Python package management and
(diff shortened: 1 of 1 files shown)


squidfunk:
I completely forgot this PR! It somehow got deleted from my inbox, which is why I'm late to the party. LGTM! Thanks!

sisp:
No worries, it happens. :wink: Thanks for liking and merging it. :pray:

squidfunk:
Perfect. If you or somebody else has a better primer on Python, happy to replace it. realpython.com tries to force oyu into their paywalled content and/or forces you to sign up after reading 1-2 articles.

alexvoss:
I was hoping that [Code Academy](https://www.codecademy.com) would provide but [all I could find was this](https://www.codecademy.com/resources/docs/python/pip) and it seems a bit thin. The Python documentation has a [useful combo of describing `pip` and `venv`](https://packaging.python.org/en/latest/tutorials/installing-packages/#creating-virtual-environments). Regarding the RealPython tutorial, I think it tries to do too much and does not distinguish between different use cases. Just my 2p.

squidfunk:
Hmm, maybe a good idea for a blog article on our own blog then!

alexvoss:
You read my mind but I am slowly learning to prioritize - mainly by sticking things into a backlog, which I will do with this discussion. Might become a side-product of thinking about making producing tutorial more efficient by reusing content.

PR #6224, opened 2023-10-22
Specify container image in GitLab CI job instead of globally

sisp:
I've moved the `image` setting for GitLab CI from the global level down to the `pages` CI job specification in the docs on publishing a site via GitLab Pages. I think this is a better example as the `python:latest` image might only apply to the `pages` job. Specifying it locally is definitely correct whereas specifying it globally _may_ be correct, but we cannot know as it depends on the rest of the CI config.

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -152,9 +152,9 @@ contents:
 === "Material for MkDocs"
 
     ``` yaml
-    image: python:latest
     pages:
       stage: deploy
+      image: python:latest
       script:
         - pip install mkdocs-material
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! I trust you on this one, as I have no experience with GitLab CI.

sisp:
Just following up briefly: It even turns out that the [`image` keyword at the _global_ level has been deprecated](https://docs.gitlab.com/ee/ci/yaml/#deprecated-keywords), so moving it into the job specification is better also from that perspective.

Issue #5758, opened 2023-07-25
ESM Support - Challenge


elliot-huffman:
### Context

The statement in the old issue is that it will hurt browser support. I would like to challenge this.

### Description

According to can I use (the same database you use, a great one)
https://caniuse.com/es6-module, if you enable support for this, you are still targeting 95%+ of all users, up by 3% from your target https://squidfunk.github.io/mkdocs-material/browser-support/ of 92%.

You are also putting your project at risk because security issues that have been identified and fixed on Mermaid.JS are not being implemented on your project:
e.g. https://github.com/knsv/mermaid/issues/847 (this example is where more bypasses were found for the old issue and were patched again to prevent more bypasses.)

### Related links

[#5193](https://github.com/squidfunk/mkdocs-material/issues/5193)

### Use Cases

Keeping Deps Up to date and reduce project complexity by allowing to update more deps to the latest version and use newer tech.

### Before submitting

- [X] I have read and followed the [change request guidelines](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/).
- [X] I have verified that [my idea is a change request and not a bug report](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/#its-not-a-bug-its-a-feature).
- [X] I have ensured that, to the best of my knowledge, [my idea will benefit the entire community](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/#benefit-for-the-community).
- [X] I have included relevant links to [the documentation](https://squidfunk.github.io/mkdocs-material), related [issues](https://github.com/squidfunk/mkdocs-material/issues), and [discussions](https://github.com/squidfunk/mkdocs-material/discussions) to underline the need for my idea.

facelessuser:
The main issue that is preventing the latest versions of Mermaind (that uses ESM modules) is that MkDocs doesn't support them.  Once MkDocs supports them, themes like Material can then add support. See relevant discussion: https://github.com/squidfunk/mkdocs-material/discussions/5708.

elliot-huffman:
correct me if I am wrong but wasn't this fixed in 1.5.0?
https://github.com/mkdocs/mkdocs/issues/3164

facelessuser:
Yes, but there are no releases yet for 1.5.0. And once there is, there will need to be work here to add support.  I'm also not sure of the timeline for getting support into Material either, only that I believe there are plans to get it in sometime after 1.5.0.

[...]

PR #6265, opened 2023-10-31
Upgrade to Mermaid 10.6.1

lambda:
Bump the version of Mermaid up to 10.6.1. Per the release notes at https://github.com/mermaid-js/mermaid/blob/develop/CHANGELOG.md the render function is async now, so update to the new async render function.

There's also apparently a change to use ESM, but I didn't seem to have any trouble with just bumping the version and updating it to use async. It looks like it's working without any change to how we load it, but I'm not much of a frontend developer so I could be missing something.


mampersat:
+1

lambda:
@manselmi Rebased and updated to pull in 10.6.1

@squidfunk Looks like you're aware of this, are there any concerns that I should address? It should be fairly easy to test out. I saw the discussion about ESM on on other PRs, but there didn't seem to need to be any updates to how this was loaded, and it works in the browsers I have tested. I don't have any older browsers easily available to test against.

squidfunk:
Thanks for the PR! As mentioned in https://github.com/squidfunk/mkdocs-material/issues/5758#issuecomment-1792078444, I need some time to test this thoroughly. I'm currently focusing on #6307, but as soon as we have a first prototype deployed, I'll take some time to revisit this PR ☺️

squidfunk:
So, thanks for your work! This PR looks good, and my testing shows everything we supported up to now is still supported. New diagram types might need additional styles for better integration, so PRs welcome.

PR #6267, opened 2023-11-01
added short description + links to MkDocs docu

alexvoss:
I thought the description of how to deploy from your local machine to GH Pages was a bit brief.

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -145,10 +145,10 @@ mkdocs gh-deploy --force
 This will build your documentation and deploy it to a branch
 `gh-pages` in your repository. See [this overview in the MkDocs
 documentation] for more information. For a description of the
-arguments, see [this section in the MkDocs documentation].
+arguments, see [the documentation for the command].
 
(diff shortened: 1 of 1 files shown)


docs/publishing-your-site.md
```diff
@@ -142,6 +142,14 @@ the following command from the directory containing the `mkdocs.yml` file:
 mkdocs gh-deploy --force
 ```
 
+This will build your documentation and deploy it to a branch
+`gh-pages` in your repository. See [this overview in the MkDocs
```
squidfunk, line 146: "See this section" is not a good link name from an A11y point of view. We try to minimize those, because when you step with <kbd>tab</kbd> through the docs, your screenreander will not provide enough context. We definitely have some parts where we also do it like that, but when updating our documentation we should always try to give links discernible names, so that they can be understood on their own.

squidfunk (changes requested):
Good addition! Minor changes.

docs/publishing-your-site.md
alexvoss: Hope the wording is better now. Also tried to avoid the repetition of "MkDocs documentation".

squidfunk (approved):

PR #6268, opened 2023-11-01
Work around collapsible block icon rotation bug

sisp:
I've added a workaround for a [Firefox bug that renders a subpixel outline when rotating a mask image element](https://bugzilla.mozilla.org/show_bug.cgi?id=1671784) – in this case the collapsible block icon:

bug

| Before | After |
| - | - |
| before | after |

The fix is similar but not identical to that of #5937. There, the icon was contained in a tight parent element to which the workaround styles were applied as suggested in the upstream Firefox issue. Here, the "parent" element is `<summary>`, which spans the entire row, and the icon is defined via its `::after` pseudo-element. So, applying the `border-radius: 0.01px` hack makes no sense as the row has a real border radius. But it turns out that adding `overflow: hidden` is enough in this case to prevent the subpixel outline.


sisp:
/ping @squidfunk Just in case you missed this PR. 😉

src/templates/assets/stylesheets/main/extensions/pymdownx/_details.scss
```diff
@@ -69,6 +69,10 @@
     display: block;
     min-height: px2rem(20px);
     padding-inline-end: px2rem(36px);
+    // Hack: Work around Firefox bug that renders a subpixel outline when
+    // rotating a mask image element.
+    // https://bugzilla.mozilla.org/show_bug.cgi?id=1671784
```
squidfunk, line 74: Could you minify the link please? I use [t.ly](https://t.ly/) nowadays, as bit.ly became unusable. Please reference the link similar to how we do it in other parts of our code base, e.g.:

https://github.com/squidfunk/mkdocs-material/blob/c0755bf2471a3476d9592de99384abe476c4a645/src/plugins/blog/readtime/__init__.py#L34

squidfunk (changes requested):

src/templates/assets/stylesheets/main/extensions/pymdownx/_details.scss
squidfunk: Oh, and maybe you could also do that here, that slipped my attention:

https://github.com/squidfunk/mkdocs-material/blob/c3683b240f665c2cca87d1cf59e784e0f827b244/src/templates/assets/stylesheets/main/components/_tooltip.scss#L165-L167

sisp: Done. :slightly_smiling_face:

squidfunk (approved):

squidfunk:
Thanks for your work!

PR #6272, opened 2023-11-01
git-committers 2.0: add token documentation

ojacques:
As we are getting closer to having git-committers be included in mkdocs-material (yeah!), I took some time to update [mkdocs-git-committers-plugin-2 plugin](https://github.com/ojacques/mkdocs-git-committers-plugin-2) so that it:

- Does not use an undocumented GitHub endpoint to fetch contributors, but GitHub's GraphQL V4 API
- Support private repositories
- Does not use BeautifulSoup nor lxml: only `requests`, and `gitpython` as prerequisites (no change on those prerequisites)

As a consequence, a GitHub token is needed.

This PR:

- Updates the documentation for the `token` configuration item
- Update `pyproject.toml` to point to 2.0 release

I hope I am not missing something. Otherwise, let me know.

### DIFF
--- a/docs/setup/adding-a-git-repository.md
+++ b/docs/setup/adding-a-git-repository.md
@@ -274,6 +274,7 @@ plugins:
   - git-committers:
       repository: squidfunk/mkdocs-material
       branch: main
+      token: !ENV GH_TOKEN
 ```
 
 The following configuration options are supported:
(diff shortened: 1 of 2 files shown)


squidfunk:
> As a consequence, a GitHub token is needed.

Sadly, this might be a no-no for many users.

ojacques:
> Sadly, this might be a no-no for many users.

Understood. I will look to preserve the current "hack", and add support for a REST API version as fallback in case the unsupported URI goes away, or to add support for private repos. The REST API has a [rate limit](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limits) of 60 requests/hour/IP address when not authenticated.

Closing this PR for now.

squidfunk:
Hmm, I absolutely understand the idea to work around the rate limit. However, if we could keep the token-free variant, and allow the user to provide one once the rate limit is reached, that might be a better solution from a user experience perspective. After all, it should be possible to query the REST API with a token as well, right? Other than that I understand the move to GraphQL – you get much more data with one request.

ojacques:
> However, if we could keep the token-free variant, and allow the user to provide one once the rate limit is reached, that might be a better solution from a user experience perspective.

I like this. It also potentially makes it compatible with GitLab hosted documentation (not many public, but I stumble on a lot on-prem doc on self-hosted GitLab).

Issue #6276, opened 2023-11-02
KaTeX not rendering correctly


letian-w:
## Bug

KaTeX does not render array environment correctly.

The following is rendered correctly.

```
$$
\begin{array}{rcl}
E(X) & = & \displaystyle\sum_{k=0}^n k \cdot ({}_n C_k) p^k q^{n-k}\\\\
     & = & \displaystyle\sum
\end{array}
$$
```

The following results in the entire math section being rendered as raw code in a paragraph.

```
$$
\begin{array}{rcl}
E(X) & = & \displaystyle\sum_{k=0}^n k \cdot ({}_n C_k) p^k q^{n-k}\\\\
     & = & \displaystyle\sum_{k=0}
\end{array}
$$
```

Note the only difference is that I added `_{k=0}` to the second line of the equation.

## Expecting

Copy and paste the following code into [this KaTeX demo tool](https://sixthform.info/katex/examples/demo.html) to see how it should be rendered.

```
\[
\begin{array}{rcl}
E(X) & = & \displaystyle\sum_{k=0}^n k \cdot ({}_n C_k) p^k q^{n-k}\\\\
     & = & \displaystyle\sum_{k=0}
\end{array}
\]
```

### Related links

- [Reporting a bug](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/)
- [How it should look like](https://sixthform.info/katex/examples/demo.html)

### Reproduction

[bug.zip](https://github.com/squidfunk/mkdocs-material/files/13234393/bug.zip)

### Steps to reproduce

1. get the right dependencies
    ```
    mkdocs==1.5.3
    mkdocs-material-insiders==9.4.2
    ```
3. run `mkdocs serve`

### Browser

Chrome

### Before submitting

- [X] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
- [X] I have attached links to [the documentation](https://squidfunk.github.io/mkdocs-material/), and possibly related [issues](https://github.com/squidfunk/mkdocs-material/issues) and [discussions](https://github.com/squidfunk/mkdocs-material/discussions).
- [X] I assure that I have [removed all customizations](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/#remove-customizations) before submitting this bug report.
[...]

PR #6277, opened 2023-11-02
Update math.md for KaTeX settings

letian-w:
KaTeX also needs Arithmatex to prevent conflict with the markdown parser.

See #6276 for reference.


squidfunk:
Thanks! Definitely looks like an oversight.

PR #6305, opened 2023-11-07
Update changing-the-colors.md regarding auto dark mode

rafrafek:
I think more than one operating system allows to automatically switch between light and dark theme, e.g. iOS, iPadOS, MacOS 😎


squidfunk:
Thanks! Indeed, there are one or two by now ☺️

Issue #6087, opened 2023-09-26
Accessibility: Progress indicator/loading UI for "Instant loading" configuration


netanel-haber-deci:
### Description

Right now, there is no indicator to the user that another page is loading when [Instant Loading](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#instant-loading) is on. Sometimes pages are large and the experience is suboptimal.
It would be useful to be able to either get a progress indicator OOTB, or to be able to toggle/configure such a spinner. Maybe an overlay with a slight change in opacity for further enhancement would be useful as well.
If this already exists via a more low-level construct (specifying UI during a hook, etc.), the request is to add documentation for this use-case.
I quite like the [determinate linear indicator](https://m3.material.io/components/progress-indicators/specs#9194f993-31bb-44cb-8a46-f47be0245b91), shown at the very top of the page - this is pretty standard in plenty of websites.

### Related links

- [Requesting a change](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/)
- [Instant Loading](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#instant-loading)
- [Material Progress Indicators](https://m3.material.io/components/progress-indicators/overview)

### Use Cases

This feature would increase accessibility can directly benefit all users of the instant loading feature, especially for navigations with large payloads.

### Visuals

- [Determinate linear indicator](https://firebasestorage.googleapis.com/v0/b/design-spec/o/projects%2Fm3%2Fimages%2Fl4yjc2jk-linear-progress-indicator-Determinate_3P.mp4?alt=media&token=f05ea51f-14e1-437b-988a-7a67c78b225e)

### Before submitting

- [X] I have read and followed the [change request guidelines](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/).
- [X] I have verified that [my idea is a change request and not a bug report](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/#its-not-a-bug-its-a-feature).
- [X] I have ensured that, to the best of my knowledge, [my idea will benefit the entire community](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/#benefit-for-the-community).
[...]

PR #6322, opened 2023-11-12
Getting request total off of Content-Length header rather than event to bypass chrome bug

netanel-haber:
https://bugs.chromium.org/p/chromium/issues/detail?id=463622

### DIFF
--- a/src/templates/assets/javascripts/browser/request/index.ts
+++ b/src/templates/assets/javascripts/browser/request/index.ts
@@ -85,7 +85,12 @@ export function request(
     // Handle download progress
     if (typeof options?.progress$ !== "undefined") {
       req.addEventListener("progress", event => {
-        options.progress$!.next((event.loaded / event.total) * 100)
+        if(event.lengthComputable){
+          options.progress$!.next((event.loaded / event.total) * 100)
+        } else { // https://bugs.chromium.org/p/chromium/issues/detail?id=463622
+          const totalFromHeader = Number(req.getResponseHeader("Content-Length")) ?? 0
+          options.progress$!.next((event.loaded / totalFromHeader) * 100)
+        }
       })
 
       // Immediately set progress to 5% to indicate that we're loading
--- a/src/templates/assets/javascripts/browser/request/index.ts
+++ b/src/templates/assets/javascripts/browser/request/index.ts
@@ -85,10 +85,10 @@ export function request(
     // Handle download progress
     if (typeof options?.progress$ !== "undefined") {
       req.addEventListener("progress", event => {
-        if(event.lengthComputable){
+        if (event.lengthComputable) {
           options.progress$!.next((event.loaded / event.total) * 100)
         } else { // https://bugs.chromium.org/p/chromium/issues/detail?id=463622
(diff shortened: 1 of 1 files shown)


netanel-haber:
Hey :) I'm the original requester of https://github.com/squidfunk/mkdocs-material/issues/6087

src/templates/assets/javascripts/browser/request/index.ts
```diff
@@ -85,7 +85,12 @@ export function request(
     // Handle download progress
     if (typeof options?.progress$ !== "undefined") {
       req.addEventListener("progress", event => {
-        options.progress$!.next((event.loaded / event.total) * 100)
+        if(event.lengthComputable){
```
squidfunk, line 88: Please add whitespace around the braces ☺️

squidfunk (changes requested):
Thanks, LGTM! Minor cosmetics.

netanel-haber:
> Thanks, LGTM! Minor cosmetics.

Hey - just making sure - will this behavior be tested somehow? Is there a simple way for me to test it myself locally?

squidfunk:
Well, I'm not sure _how_ we should test this automatically, since it's a browser bug. We need to rely on manual testing and best effort. I read your PR as "it works".

netanel-haber:
> Well, I'm not sure _how_ we should test this automatically, since it's a browser bug. We need to rely on manual testing and best effort. I read your PR as "it works".

Loud and clear. Do you happen to have some recommended setup for local manual debugging/testing with a dummy website?

squidfunk:
Not really, I think, we currently have no automated regression tests, and testing side effects like this is particularly challenging.

src/templates/assets/javascripts/browser/request/index.ts
netanel-haber: Done.

netanel-haber:
Done checking, seems to work.
For what it's worth, what I did is the following, off the top of my head:

**Setting up the environment**:
I ran `npm run build`, then searched the minified js for the "progress" listener, then I swapped the augmented implementation in the minified js I have in my work Python docs project (after running `mkdocs build` locally). 

**Testing**:
* `If` branch (`event.lengthComputable===true`):
   Confirmed the loader still works in a non-buggy browser (firefox), in which I confirmed beforehand that `event.lengthComputable` is indeed `true`, so that tests the first branch.
* `else` branch:
   Checked in chrome and verified that the loader actually works, which was the initial goal of the PR.
  
Seems wonky and brittle to me, so if you do happen to have tips for the next PR, they would be much appreciated 😄

squidfunk:
Thanks for testing! In order to test something like this automatically, we'd need to set up automated regression tests. We'll be working on setting up a test infrastructure next year, so we'll consider testing the frontend components as well.

netanel-haber:
Actually, I was asking about an easier manual testing flow. Meaning, given a change/commit in the code, how do you usually pull up the changes live with some dummy website?

squidfunk:
You can use our own documentation or any other site. I usually use `pip install -e <path-to-project>` to test any site against a specific commit.

Issue #6344, opened 2023-11-15
long unbreakable headings break mobile table-of-contents view


thesuperzapper:
### Bug description

If you have a very long unbreakable heading in a page, the table of contents view (in the mobile sidebar) will not wrap it, and it will be scrollable off-screen.

I am not sure what the most desirable behavior here is, e.g. force wrapping (even in the middle of words), or making an ellipsis, or something else.

![Screenshot 2023-11-15 at 15 13 29](https://github.com/squidfunk/mkdocs-material/assets/5735406/bbbaad08-7e80-45b9-bb08-862a192d92ae)

### Related links

- https://github.com/squidfunk/mkdocs-material/issues/84

### Reproduction

[9.4.8-long-headings-break-toc.zip](https://github.com/squidfunk/mkdocs-material/files/13370590/9.4.8-long-headings-break-toc.zip)

### Steps to reproduce

1. Create a very long heading in a page (which is not breakable with spaces, dashes, etc)
2. Open page on mobile
3. View sidebar table of contents
4. Notice that the heading link goes off page in a way which is not scrollable on mobile that

### Before submitting

- [X] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
- [X] I have attached links to [the documentation](https://squidfunk.github.io/mkdocs-material/), and possibly related [issues](https://github.com/squidfunk/mkdocs-material/issues) and [discussions](https://github.com/squidfunk/mkdocs-material/discussions).
- [X] I assure that I have [removed all customizations](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/#remove-customizations) before submitting this bug report.
- [X] I have attached a __.zip file__ with a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/).

alexvoss:
I assume that the long unbreakable title is generated from source code so you cannot just avoid it? I am not entirely sure there is a generic solution for this, so would suggest a customization. You could write a [hook](https://www.mkdocs.org/user-guide/configuration/#hooks) that plugs into the [`on_nav` event](https://www.mkdocs.org/dev-guide/plugins/#on_nav) to modify the navigation where necessary?

alexvoss:
[...]

Issue #6281, opened 2023-11-02
Bug in right nav styles on safari v17.1


tlambert03:
### Bug description

commit 521f7272ee266526102261933e383e6f74eed062 seems to have introduced a bug wherein the right nav appears clipped in safari (Version 17.1 (19616.2.9.11.7)) after page refresh.  resizing the window across any breakpoint fixes it)

https://github.com/squidfunk/mkdocs-material/assets/1609449/916443f0-c595-412f-9a87-350ab972781c

The issue is visible as of v9.4.2 and git bisect put the issue at 521f7272ee266526102261933e383e6f74eed062

chrome, Firefox, edge work fine.

### Related links

- This is the same as #6273 which was closed

I was unable to find any other related issues since sept 23, which is when 521f7272ee266526102261933e383e6f74eed062 was merged

### Reproduction

[9.4.7-right-nav.zip](https://github.com/squidfunk/mkdocs-material/files/13243633/9.4.7-right-nav.zip)

(edited to a more minimal lockfile)

### Steps to reproduce

This bug can be seen on a brand new install. with only the theme added mkdocs.yml
```sh
mkdir test
cd test
pip install --upgrade --force-reinstall mkdocs-material
mkdocs new .
echo "\ntheme:\n  name: material" >> mkdocs.yml
mkdocs serve
```

In safari v17.1 the right nav is missing until you resize the window (see example above). 

```
pip install 'mkdocs-material<9.4.2'
mkdocs serve
```

fixed...

```
pip install 'mkdocs-material==9.4.2'
mkdocs serve
```

broken

### Browser

Safari

### Before submitting

- [X] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
- [X] I have attached links to [the documentation](https://squidfunk.github.io/mkdocs-material/), and possibly related [issues](https://github.com/squidfunk/mkdocs-material/issues) and [discussions](https://github.com/squidfunk/mkdocs-material/discussions).
- [X] I assure that I have [removed all customizations](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/#remove-customizations) before submitting this bug report.
- [X] I have attached a __.zip file__ with a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/).

squidfunk:
[...]

PR #6351, opened 2023-11-17
added md-ellipsis span to toc-item.html

alexvoss:
This should fix https://github.com/squidfunk/mkdocs-material/issues/6344.

In the public version, long TOC items were not wrapped in `.md-ellipsis` and so just cut off. Seemed an unintended divergence between the two versions to me.


squidfunk:
Thanks for the PR. Indeed, this looks like an oversight on my side.

squidfunk:
The issue is resolved, see https://github.com/squidfunk/mkdocs-material/issues/6281#issuecomment-1817804434

PR #6371, opened 2023-11-20
Replaced `enum` by `const` for constant value types in schemas

sisp:
I've refactored the schema files a bit to use the more concise [`const` keyword](https://json-schema.org/understanding-json-schema/reference/const) instead of `enum` with a single array item for constant value types. All schemas are based on JSON Schema Draft 7, and `const` was introduced in Draft 6, so it's a supported keyword.

### DIFF
--- a/docs/schema/assets/fonts.json
+++ b/docs/schema/assets/fonts.json
@@ -7,11069 +7,7907 @@
     {
       "title": "ABeeZee",
       "markdownDescription": "https://fonts.google.com/specimen/ABeeZee",
-      "enum": [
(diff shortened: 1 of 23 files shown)


sisp:
I just realized that the fonts schema is generated and I hadn't updated that code to use `const` instead of `enum`. Should be ready for a review now. :slightly_smiling_face:

squidfunk:
Thanks for the PR! I wasn't aware of `const`, but it sounds like a better fit than `enum` for all of those cases ☺️

Issue #6307, opened 2023-11-07
Towards better documentation search 


squidfunk:
## Background

As you may have read in one of my recent comments, we're currently revising our search implementation. The current search is based on [Lunr.js](https://github.com/olivernn/lunr.js), which is also the search engine that MkDocs has been using the time Material for MkDocs started in 2016. In the beginning, we felt that this was a good fit, as Lunr.js allows searching in the browser without the need for an external service. This makes deploying documentation much simpler, since search is and should always be a central component to each and every good documentation site.

In the past years, we've invested hundreds of hours into making search better. With the help of our awesome sponsors, we were able to ship [rich search previews](https://squidfunk.github.io/mkdocs-material/blog/2021/09/13/search-better-faster-smaller/#rich-search-previews), [support for more sophisticated tokenizers](https://squidfunk.github.io/mkdocs-material/blog/2021/09/13/search-better-faster-smaller/#tokenizer-lookahead), [support for Chinese](https://squidfunk.github.io/mkdocs-material/blog/2022/05/05/chinese-search-support/), as well as [better highlighting](https://squidfunk.github.io/mkdocs-material/blog/2021/09/13/search-better-faster-smaller/#accurate-highlighting). Additionally, we made search almost [twice as fast](https://squidfunk.github.io/mkdocs-material/blog/2021/09/13/search-better-faster-smaller/#benchmarks). However, in order to progress, and solve the many open issues that are related to search, we decided to throw out Lunr.js. There are several reasons for that, the most important of which that it is unmaintained since 2020. Additionally, Lunr.js only allows ranking with [BM25](https://en.wikipedia.org/wiki/Okapi_BM25), which is a good basis, but almost all issues that are related to weird rankings are caused by the fact that BM25 is not ideal for stable typeahead search. It was meant for full-word retrieval and is almost impossible to tame for the many different use cases that we've seen in the wild. Again, we've invested a lot of time to improve the situation, but we've reached an end where this doesn't make sense anymore.

This is the reason why we're currently releasing so few new features, because we're putting our entire energy in finishing the new search implementation. We're already almost en-par with Lunr.js' functionality, but now have an __entirely modular architecture__, which will allow us to swap out everything. Yes, I mean __everything__: the ranking algorithm, wildcard matching, the inverted index implementation, yada, yada, yada. Solving the documentation search problem is a personal affair for me. I really hate that there's not yet a solution that works reliably, can run anywhere, and is modular so it can be easily customized.

[...]

PR #6372, opened 2023-11-20
Search: Research Preview 2 🧪

squidfunk:
> [!IMPORTANT]
> __This PR presents an early preview of our new search feature for gathering user feedback.__ It is not yet ready for production and is subject to modifications. Currently, it lacks support for search suggestions and does not cater to Japanese, due to the absence of segmentation functionality.

### Overview of this PR

This PR offers the second iteration of our revamped search functionality. While still in development and devoid of UI changes (for now), it marks a departure from the current implementation. It builds upon the previous [research preview](https://github.com/squidfunk/mkdocs-material/pull/6321), forming an intersection instead of a union, and employs a new, experimental ranking method. __We are eager to know whether you find this change enhances or diminishes the search results.__

The ranking function is a variant of BM-25 that takes the proximity of matched terms into account. Our testing shows, that typeahead results should be much more stable, and results should be much more sensical. For example:

__Lunr.js (current `master`)__

screenshot-squidfunk-github-io-mkdocs-material-setup-1700488454381

__This PR__:

screenshot-localhost-3000-mkdocs-material-setup-setting-up-site-search-1700488460680

As noted, the new ranking function is still experimental, and we have several ideas to make it even more concise. It should, however, be already better than what we currently have, as it can now find consecutive words more easily.

### Current limitations

- No Chinese and Japanese segmentation
- No stemming
- No special syntax, e.g. `"foo bar"` or `+foo -bar`
- No boosting

All of those will be resolved before we release the new search.

### Your feedback is essential 🙋‍♀️

We want to co-create the search feature with our community. Our goal is to deliver an unparalleled search experience to _all_ Material for MkDocs users, not just our sponsors. This journey begins with a revamped interface and, more critically, heightened accuracy to help users find what they need quickly and effortlessly. __After establishing a solid foundation, we will explore many exciting extensions, which, as always, will first be released exclusively to our sponsors__. But first, we aim to elevate the experience for everyone.

To test this version featured in the PR, you can install it using `pip`:

``` sh
pip install git+https://github.com/squidfunk/mkdocs-material.git@spike/search-preview-2
```

We warmly invite you to share your honest feedback on this PR! Please note that this PR is specifically designed for feedback collection and will not be merged. Your insights are invaluable to us! ☺️


strausmann:
I just wanted to test Preview 2 on our project. Unfortunately, I get this error message.
```bash
ERROR - Config value 'plugins': The "meta" plugin is not installed
```
The branch will support all other features or not?

squidfunk:
It appears you're trying to install it on an Insiders project. Since this is the community edition, it is currently not compatible with Insiders projects. Thus, you'd have to disable the plugins that are only available in Insiders. Once we have a stable version (and are beyond mere research previews), we'll release an alpha version on both editions.

pawamoy:
I did not test extensively, but it looks like the results are better with the new search.

- Some entries that were previously displayed too low (8th position) are now appearing higher (3rd position). It seems results for searches with multiple words are better than before. 
- Most results are similar to the previous version, with the ordering just a bit different (first and second results switched).
- Some searches would previously yield lots of pages with "Missing: words" before a good result. Now those results are not shown and only the matching page is shown :+1: 

Sorry if my phrasing does not make sense, let me know if you'd like some screenshots or anything else :slightly_smiling_face:

karengermond:
I tested this and am wondering if there will be a mechanism for boosting topics/sections or some other method like that for controlling search results?  I'm finding that the new search isn't performing better than lunr because I made extensive use of boosting which worked well in many cases.

karengermond:
> ```shell
> pip install git+https://github.com/squidfunk/mkdocs-material.git@spike/search-preview-2
> ```

We just have nonstandard data (proprietary so I can't make it available).  Here's an example: we have a language construct "on" which is used in a heading H2. But the term "on" is used in a whole bunch of other headings as H1s. All of those hits come up first in the new search and the old search.  That's a really tough one so I'd need to do something to boost the section. 

I have a question -- how do I go back to the old version of mkdocs with lunr?

karengermond:
Thanks!

day-jeff:
Search is dramatically improved! Nice work. 

I want to deploy to our internal production site ASAP. Any idea when you might integrate into the main version? (Since search is important for us, and most of the recent updates are not, I'm tempted to deploy the test version.)

squidfunk:
I would not encourage you to deploy this version. This is only a research preview and we don't offer support for bugs that you discover when deploying it into production. Please see https://github.com/squidfunk/mkdocs-material/issues/6307#issuecomment-1889319317 for when search will be available. You can probably imagine that I'm being asked that every other day 😅

squidfunk:
Closing. Next search preview will be ready soon ☺️

PR #6373, opened 2023-11-20
Update setting-up-tags.md

letian-w:
Settings in mkdocs.yml need to be configured for tags.md to be recognized.


squidfunk:
Thanks! After cutting out the plugin configuration, the next step for the documentation is to rewrite the usage guide for each plugin. This is definitely something that is currently missing, so thanks for the PR!

Issue #6378, opened 2023-11-21
YAML schema missing markdown extensions


AdrianoKF:
### Bug description

The JSON schema definition for the `mkdocs.yml` format does not recognize a couple of Markdown extensions:

- The [Neoteroi MkDocs plugins](https://www.neoteroi.dev/mkdocs-plugins)
- `sane_lists` from `python-markdown`

I can prepare a PR adding the relevant plugins to the schema (however, I might be missing others that I haven't run into just yet).

### Related links

- Neoteroi MkDocs plugin: https://www.neoteroi.dev/mkdocs-plugins
- `sane_lists` markdown extension: https://python-markdown.github.io/extensions/sane_lists/

### Reproduction

Not including, since this is not an issue with a generated site.

### Steps to reproduce

Here's a snippet from my `mkdocs.yml`, both lines trigger an error message saying the value is not accepted in VSCode:

```yaml
markdown_extensions:
  - neoteroi.cards  # https://www.neoteroi.dev/mkdocs-plugins/cards/
  - sane_lists
```

Full error message:

```
Value is not accepted. Valid values: "markdown.extensions.abbr", "abbr", "markdown.extensions.admonition", "admonition", "markdown.extensions.attr_list", "attr_list", "markdown.extensions.callouts", "callouts", "markdown.extensions.def_list", "def_list", "markdown.extensions.footnotes", "footnotes", "markdown.extensions.md_in_html", "md_in_html", "markdown.extensions.tables", "tables", "markdown.extensions.toc", "toc", "pymdownx.arithmatex", "pymdownx.betterem", "pymdownx.caret", "pymdownx.critic", "pymdownx.details", "pymdownx.emoji", "pymdownx.highlight", "pymdownx.inlinehilite", "pymdownx.keys", "pymdownx.magiclink", "pymdownx.mark", "pymdownx.smartsymbols", "pymdownx.snippets", "pymdownx.superfences", "pymdownx.tasklist", "pymdownx.tilde".yaml-schema: Abbreviations – Python Markdown(1)
```

### Before submitting

- [X] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
- [X] I have attached links to [the documentation](https://squidfunk.github.io/mkdocs-material/), and possibly related [issues](https://github.com/squidfunk/mkdocs-material/issues) and [discussions](https://github.com/squidfunk/mkdocs-material/discussions).
[...]

PR #6381, opened 2023-11-21
Add mkdocs.yml JSON schema support for `sane_lists` and Neoteroi markdown extensions

AdrianoKF:
As suggested in #6378, this PR adds support for several markdown extensions to the JSON schema for `mkdocs.yml`:

- [`sane_lists`](https://python-markdown.github.io/extensions/sane_lists/) from `python-markdown`
- The Cards, Timeline, Gantt, and Spantable markdown extensions from [`neoteroi-mkdocs`](https://www.neoteroi.dev/mkdocs-plugins/) (I've added a separate `external/` folder under the folder containing the schemas for markdown extensions - does that make sense?)

### DIFF
--- a/docs/schema/extensions.json
+++ b/docs/schema/extensions.json
@@ -10,6 +10,9 @@
       },
       {
         "$ref": "extensions/pymdownx.json"
+      },
+      {
+        "$ref": "extensions/external/neoteroi.json"
       }
     ]
   },
--- a/docs/schema/extensions/external/neoteroi.json
+++ b/docs/schema/extensions/external/neoteroi.json
@@ -1,0 +1,27 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema",
+  "title": "Neoteroi markdown extensions",
+  "markdownDescription": "https://www.neoteroi.dev/mkdocs-plugins",
+  "oneOf": [
+    {
+      "title": "Cards – Neoteroi Markdown",
+      "markdownDescription": "https://www.neoteroi.dev/mkdocs-plugins/cards/",
+      "const": "neoteroi.cards"
+    },
+    {
+      "title": "Timeline – Neoteroi Markdown",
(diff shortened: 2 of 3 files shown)


squidfunk:
Thanks! Could I please ask you to split the configuration similar to what we did for plugins here?
https://github.com/squidfunk/mkdocs-material/blob/master/docs/schema/plugins.json

The idea is to group external Markdown extensions that are not natively supported by Material for MkDocs under a dedicated group inside the schema. This makes it more clear what we support and what is supported through third parties.

AdrianoKF:
Sure, that makes perfect sense! Now that I see that the plugins schema already references some external schemas by URL, I might split the Neoteroi part of this PR off and submit it as a PR on their repo instead (and then change this PR accordingly).

As for `sane_lists`, I would leave it in the current place, since it is part of the official `python-markdown` extensions (although not explicitly mentioned in the mkdocs-material docs)

squidfunk:
> Sure, that makes perfect sense! Now that I see that the plugins schema already references some external schemas by URL, I might split the Neoteroi part of this PR off and submit it as a PR on their repo instead (and then change this PR accordingly).

That would in fact be perfect and favorable over hosting the external schema here. In fact, we try only to host external schemas for extensions we mention on our docs, so this would be even better!

squidfunk:
Could you please split this PR into adding `sane_lists` and the other extension? We can merge `sane_lists` right aways, but we'll wait until the other schema has been added to the upstream repository.

AdrianoKF:
In fact, the Neoteroi maintainer has merged the PR already. I have updated this PR with the external URL accordingly.

Happy to split this PR, if you'd prefer it for bookkeeping reasons, but otherwise it should be go to go as is. 👍🏻

squidfunk:
Perfect, that was quick! No need to split into a separate PR – as it's already merged, all good.

PR #6409, opened 2023-11-29
Removed Support for Ko-fi

katharinalisalin:
This PR removes all documentation of the payment platform Ko-fi, as this payment option will no longer be available.


squidfunk:
LGTM!

PR #6422, opened 2023-12-04

MeDott29:
updated to the version of ci.yml in James Willet's tutorial 
https://www.youtube.com/watch?v=Q-YA_dA8C20&ab_channel=JamesWillett

it fixed my github page https://medott29.github.io/

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -22,35 +22,35 @@ contents:
 === "Material for MkDocs"
 
     ``` yaml
-    name: ci # (1)!
-    on:
-      push:
-        branches:
-          - master # (2)!
-          - main
-    permissions:
-      contents: write
-    jobs:
-      deploy:
-        runs-on: ubuntu-latest
-        steps:
-          - uses: actions/checkout@v4
-          - name: Configure Git Credentials
-            run: |
-              git config user.name github-actions[bot]
-              git config user.email 41898282+github-actions[bot]@users.noreply.github.com
-          - uses: actions/setup-python@v4
-            with:
-              python-version: 3.x
-          - run: echo "cache_id=$(date --utc '+%V')" >> $GITHUB_ENV # (3)!
-          - uses: actions/cache@v3
-            with:
-              key: mkdocs-material-${{ env.cache_id }}
-              path: .cache
-              restore-keys: |
-                mkdocs-material-
-          - run: pip install mkdocs-material # (4)!
-          - run: mkdocs gh-deploy --force
+name: ci 
+on:
+  push:
+    branches:
+      - master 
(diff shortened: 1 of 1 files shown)


kamilkrzyskow:
Looking past the formatting changes the commit changed the cache key to `github.ref`, which I tried using in #5460 and it was used in the past in this repo too.

As I wrote in the comment here
https://github.com/squidfunk/mkdocs-material/pull/5460#issuecomment-1531524310

Despite the docs mentioning that it would use a name as `refs/pulls/number/merge` it kept using the `refs/branches/master`, and because it kept using the same key, cache was saved only once and was constantly reused because of the primary-key hit.

Perhaps, I overlooked something 🤔, I will have to investigate the workflow run again first before I'm certain 😵‍💫.

Also @MeDott29 could you describe what did the change fix on your website? Was there some error before?

kamilkrzyskow:
Looking at https://github.com/MeDott29/MeDott29.github.io/actions/workflows/ci.yml
I see that both CI runs were successful. First uses the method described in the docs, and the second uses `github.refs`

Both of them didn't save any cache because the `.cache` directory wasn't found.

However, I still don't see any major mistake with the current workflow, so waiting for feedback 🙋‍♂️

kamilkrzyskow:
Looking at it again, the first CI run based on the current docs example created the `gh-pages` branch.
https://github.com/MeDott29/MeDott29.github.io/actions/runs/7080008473/job/19267303556#step:8:1
So the Deploy workflow didn't trigger because it wasn't configured to watch the `gh-pages` branch.
The second CI run force pushed to the `gh-pages` and triggered a deployment, so I think that's what you meant by "it fixed my github page", however it's not related with any change in the workflow file.

squidfunk:
@kamilkrzyskow thank you so much for investigating. I guess this is one of the reasons why caching is considered one of the hardest parts of software development. Closing until proven that we really need to update the documentation and change our approach to caching.

kamilkrzyskow:
Current setup updates cache once per week and allows the user to change it to update it daily, as explained within the code annotation: https://squidfunk.github.io/mkdocs-material/publishing-your-site/#__codelineno-0-21

The only valid change for even more frequent caches would be to split the single `cache` action into separate `cache/restore` and `cache/save` actions, as discussed previously: https://github.com/squidfunk/mkdocs-material/pull/5463#issuecomment-1731594584
But iirc the danger here is that forcing a save, when `.cache` doesn't exist would end up with a `mkdocs-material-` key without the hash part, which would be an exact primary key hit later on, but I could've made some testing error back then and have this bad memory now 😓 Lack of a `.cache` dir should prevent cache creation, but that shall be tested if needed in the future :v:

PR #6435, opened 2023-12-04
blog only folder tree

copdips:


squidfunk (changes requested):
Thanks for the PR! This should not go into the plugin API reference, but into the [blog setup guide](https://squidfunk.github.io/mkdocs-material/setup/setting-up-a-blog/), as it's a concrete setup example. Could you please move it there?

squidfunk:
No, it should not go into the blog plugin guide. Please move it to setup. I'm closing this PR, and you can create a new one if you wish to proceed, but we need to keep the introduction for the blog plugin as short and simple as possible. It is only a reference not a setup guide.

PR #6438, opened 2023-12-04
blog only folder tree

copdips:
Hello,

What about this time, I'm trying to put the blog only part under the `setup/setting-up-a-blog/#rss`, hereunder is the rendering:

image


squidfunk:
Thanks! It's not the perfect place, but we're planning to do a rewrite of this section anyway, so we can definitely host it there for now.

PR #6445, opened 2023-12-06
Fixing links to blog HTML templates

Cvaniak:
Just simple update of not working links.


squidfunk:
Thanks! We moved those templates around at some point and forgot to update the links.

PR #6467, opened 2023-12-08
Add raw date as title attribute

Sija:
Hi,

This PR adds raw dates as title attributes to the `{{ date }}` values (showing approximate, relative date) shown in `source-file.html` partial.


squidfunk:
Thanks for the PR. We're currently not considering this change. However, you can easily apply those changes to your own installation by using [theme extension](https://squidfunk.github.io/mkdocs-material/customization/#extending-the-theme).

PR #6468, opened 2023-12-08
Dockerfile: Write to /etc/gitconfig for all users

iBug:
Change how `git config` is handled in the Docker image.

As it's currently written, `git config --global` writes to `/root/.gitconfig`. This file is not visible if a container is created specifying another user (e.g. `docker run -u 1000:1000`), which is a common practice when using Docker on user-owned files instead of root-owned files.

By replacing `--global` with `--system`, Git writes to `/etc/gitconfig`, whose content applies to all users on this system (container image).

I'm also curious why these `git config` commands are given on a separate `RUN` directive - seems unnecessary to create another layer just for a single `.gitconfig` / `gitconfig` file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -83,11 +83,11 @@ RUN \
   find ${PACKAGES} \
     -type f \
     -path "*/__pycache__/*" \
-    -exec rm -f {} \;
-
-# Trust directory, required for git >= 2.35.2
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. I think this should be okay to merge. Additionally, we could definitely move this into the prior RUN directive, so you're invited to do so ☺️

PR #6471, opened 2023-12-09
Recommend another way to store snippets

ofek:
This is what I've always done; feel free to close if this is too much for that note.

Example: https://github.com/pypa/hatch/tree/master/docs/.snippets


squidfunk:
Thanks for suggesting. While perfectly viable, this is not something we want to recommend to users.

PR #6479, opened 2023-12-11
Details / Collapsible blocks - Rotate Arrows

lenisko:
Wouldn't it be more intuitive having arrows point the way that detail's section will expand/hide?
The current solution is confusing for some, they don't actually know that section is expandable.

**Proposed change**

2023-12-11_02-22-50-firefox

**Current form**

2023-12-11_02-27-36-firefox


squidfunk:
Thanks for suggesting. This is a matter of opinion, e.g., several SSGs also use this pattern for expandable sidebars, like Material for MkDocs as well. However, it is trivial to achieve with customization, so please just use an [additional stylesheet](https://squidfunk.github.io/mkdocs-material/customization/#additional-css).

PR #6494, opened 2023-12-12
show email addresses

fghaas:
- Only show authors' email addresses if so configured in git-authors
- Only show authors' email addresses if so configured in git-authors


src/templates/partials/source-file.html
```diff
@@ -54,9 +56,7 @@
     </span>
     <nav>
       {% for author in authors %}
-        <a href="mailto:{{ author.email }}">
-          {{- author.name -}}
-        </a>
+        {%- if show_email_address %}<a href="mailto:{{ author.email }}">{{- author.name -}}</a>{%- else -%}{{- author.name -}}{% endif -%}
```
squidfunk, line 59: This is not really readable. Please adhere to the surrounding formatting, so move things onto separate lines. There's also no need to put `show_email_address` in a separate variable I think.

squidfunk (changes requested):
Minor cosmetic changes, otherwise good!

src/templates/partials/source-file.html
fghaas: Roger that, updated.

src/templates/partials/source-file.html
```diff
@@ -54,9 +55,11 @@
     </span>
     <nav>
       {% for author in authors %}
-        <a href="mailto:{{ author.email }}">
+        {%- if git_authors.config.show_email_address %}
```
squidfunk, line 58: One more spot: We don't need whitespace control here. Please remove the `-` on all tags that you added.

squidfunk (changes requested):

src/templates/partials/source-file.html
fghaas: Does that include the one instance of `{{- author.name -}}` that I duplicated? In other words, should the hyphens be dropped from that one as well?

squidfunk: Yes. The following line is the only line where this is necessary, because we want the `,` to stick to the author name:

https://github.com/squidfunk/mkdocs-material/blob/2ca5cf0ea0627982f595c14cb8080447f9a9d24c/src/templates/partials/source-file.html#L60

fghaas: I've just pushed 7f0cdd24fa91a5e7c396a4a80898a00b86ce4c5c which is the most minimal patch that I can think of, with no whitespace management added at all. But that introduces the extra space before the comma, no matter if we're rendering a link or not. So please do not merge that as-is.

fghaas: That fixes things only for `show_email_address: false`; but still breaks (i.e. introduces the extraneous whitespace) for the default of `show_email_address: true`.

fghaas: This does the right thing, regardless of how `show_email_address` is set:

```diff
       {% for author in authors %}
+        {%- if git_authors.config.show_email_address %}
         <a href="mailto:{{ author.email }}">
           {{- author.name -}}
         </a>
+        {%- else -%}
+          {{- author.name -}}
+        {% endif -%}
         {%- if loop.revindex > 1 %}, {% endif -%}
       {% endfor %}
```

Okay if I update my topic branch with that?

squidfunk: I'm sure you don't need all of those whitespace control characters, but okay, let's keep the whitespace control. I'll look into in the future.

fghaas: I sympathise with that, but if you instruct me to "remove the `-` on all tags that [I] added" and simultaneously "we want the `,` to stick to the author name", I can only do one or the other. :)

fghaas:
> Is it okay to merge now?

Well I'd still been trying to wrap my head around how to add tests for this, hence the Draft mode. I've looked at [what I think is the relevant part of the docs](https://squidfunk.github.io/mkdocs-material/contributing/making-a-pull-request/#testing-and-reviewing-changes), but most of those recommendations don't seem to apply here since neither mkdocs-material's own docs nor the examples repo use `git-authors`. 

I've tested it against the repo I co-maintain (https://github.com/citynetwork/docs/pull/257) and there all our tests pass. Is that sufficient?

squidfunk:
Yes, that should be fine. Thanks!

squidfunk (approved):

PR #6502, opened 2023-12-13
Fixed privacy plugin to exclude Giscus client script

sisp:
The [Giscus client script cannot be self-hosted](https://github.com/giscus/giscus/issues/900#issuecomment-1427794340), as the Giscus widget fails to work then. Since [Material for MkDocs officially supports Giscus](https://squidfunk.github.io/mkdocs-material/setup/adding-a-comment-system/) and the privacy plugin has become available to everybody thanks to meeting the "Goat's Horn" funding goal &ndash; although without the [`assets_exclude` setting](https://squidfunk.github.io/mkdocs-material/plugins/privacy/#config.assets_exclude), which would allow excluding the Giscus client script URL in userland &ndash; I think it's fair to support usage of the privacy plugin along with Giscus for everybody. Also Insiders users, who use the pivacy plugin, would benefit as Giscus would simply work out of the box also for them.

To make the privacy plugin work with Giscus, I've added a special case that excludes the Giscus client script from getting self-hosted.

WDYT, @squidfunk?

### DIFF
--- a/src/plugins/privacy/plugin.py
+++ b/src/plugins/privacy/plugin.py
@@ -209,6 +209,10 @@ class PrivacyPlugin(BasePlugin[PrivacyConfig]):
 
     # Check if the given URL is excluded
     def _is_excluded(self, url: URL, initiator: File | None = None):
+        # Skip Giscus client script - see https://t.ly/QTJ3s
+        if url.geturl() == "https://giscus.app/client.js":
+            return True
+
         if not self._is_external(url):
             return True
 


squidfunk:
Thanks for the PR. However, we really shouldn't add more and more exceptions to the plugin, with [Mermaid.js being one of them](https://github.com/squidfunk/mkdocs-material/blob/9b98c7772560f4cd98faca6ef000cb141b1cc9ee/src/plugins/privacy/plugin.py#L93-L102), but which is natively built into the theme. Giscus is not the only case that does not work, as mentioned in our documentation. [MathJax is another one](https://squidfunk.github.io/mkdocs-material/plugins/privacy/#__code_11_annotation_1). What if users discover other third-party scripts that don't work with the plugin?

I understand that the [`assets_include`](https://squidfunk.github.io/mkdocs-material/plugins/privacy/#config.assets_include) and [`assets_exclude`](https://squidfunk.github.io/mkdocs-material/plugins/privacy/#config.assets_exclude) directives mandate the use of [Insiders](https://squidfunk.github.io/mkdocs-material/insiders/), as they form the latest effort of making the privacy plugin even more flexible, but please note that we must keep this project sustainable and can't yet release them into the wild. Please understand that the need for communication and support on this project has skyrocketed in the past months because of massive increasing adoption. I can't sugar-coat this: we really need more funds to grow our team. The privacy plugin is already very powerful, in that it allows you to use external assets almost without thinking about them, including Google Fonts. There are [limitations](https://squidfunk.github.io/mkdocs-material/plugins/privacy/#limitations).

Additionally, excluding Giscus from the downloaded assets while downloading everything else sounds like an esoteric case to me, as it will again, mandate a cookie banner to comply with GDPR. The fact that Giscus cannot be self-hosted is an upstream issue that must be addressed in Giscus. We cannot solve it on our side.

If somebody wants this functionality nonetheless, forking the theme is always an option, taking this PR as a basis. The new privacy directives will be released once the [Jalapeño funding goal](https://squidfunk.github.io/mkdocs-material/insiders/#20000-jalapeno) is hit. I hope for your understanding.

sisp:
Thanks for your comprehensive feedback, @squidfunk! :pray: I understand your position. Indeed, an upstream fix would be preferable.

If anybody stumbles over this issue, a (non-GDPR compliant) workaround is to construct the Giscus `<script>` tag dynamically using JavaScript with a string-concatenated client script URL to avoid being matched by the privacy plugin:

```diff
-<script
-  src="https://giscus.app/client.js"
-  data-repo="<username>/<repository>"
-  data-repo-id="..."
-  data-category="..."
-  data-category-id="..."
-  data-mapping="pathname"
-  data-reactions-enabled="1"
-  data-emit-metadata="1"
-  data-theme="light"
-  data-lang="en"
-  crossorigin="anonymous"
-  async
->
-</script>
+<script>
+  var script = document.createElement("script");
+  script.setAttribute("src", "https://" + "giscus.app" + "/client.js");
+  script.setAttribute("data-repo", "<username>/<repository>");
+  script.setAttribute("data-repo-id", "...");
+  script.setAttribute("data-category", "...");
+  script.setAttribute("data-category-id", "...");
+  script.setAttribute("data-mapping", "pathname");
+  script.setAttribute("data-reactions-enabled", "1");
+  script.setAttribute("data-emit-metadata", "1");
+  script.setAttribute("data-theme", "light");
+  script.setAttribute("data-lang", "en");
+  script.setAttribute("crossorigin", "anonymous");
+  script.setAttribute("async", true);
+  document.currentScript.insertAdjacentElement("afterend", script);
+</script>
```

PR #6519, opened 2023-12-17
Update docs about general availability of automatic light/dark mode

sisp:
Just a minor docs update on automatic light/dark mode being generally available now and not only via Insiders. I've actually only replaced `[Insiders]` by `Material for MkDocs`, the rest of the diff is related to changes due to the line length limit.


squidfunk:
Perfect, definitely an oversight in the latest merge. Thanks!

PR #6533, opened 2023-12-18
Python code improvements

sisp:
I've done a little Python code cleanup, split into several commits to separate the changes by type.

### DIFF
--- a/src/plugins/privacy/plugin.py
+++ b/src/plugins/privacy/plugin.py
@@ -62,11 +62,11 @@ class PrivacyPlugin(BasePlugin[PrivacyConfig]):
 
         # Initialize collections of external assets
         self.assets = Files([])
-        self.assets_expr_map = dict({
+        self.assets_expr_map = {
             ".css": r"url\((\s*http?[^)]+)\)",
             ".js": r"[\"'](http[^\"']+\.(?:css|js(?:on)?))[\"']",
             **self.config.assets_expr_map
-        })
+        }
 
     # Process external style sheets and scripts (run latest) - run this after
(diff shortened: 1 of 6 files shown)


src/plugins/privacy/plugin.py
```diff
@@ -247,7 +247,7 @@ def _parse_fragment(self, fragment: str):
 
     # Parse and extract all external assets from a media file using a preset
     # regular expression, and return all URLs found.
-    def _parse_media(self, initiator: File) -> "list[URL]":
+    def _parse_media(self, initiator: File) -> list[URL]:
```
squidfunk, line 250: Does this work with Python 3.8?

sisp: Oh, no, you're right, it doesn't. I've dropped that commit.

squidfunk: We'll add proper tests next year for all supported Python versions, that will safe-guard us against those cases.

sisp: How about also using a proper dependency manager like [Poetry](https://github.com/python-poetry/poetry), which supports a lock file and dev-dependencies, and [Ruff](https://github.com/astral-sh/ruff) as formatter and linter? Plus, of course, mypy for the type-checking that would have caught the above problem under Python 3.8.

squidfunk: Yes, we're on it, please see #6473. Note that my adventure into Python is kind of an accident, as I'd never would've thought that I will develop that many MkDocs plugins. The code base grew so large with the new capabilities, that we now need to professionalize the project. We will do this in 2024.

PR #6534, opened 2023-12-18
Improved Group plugin initialization for mike

kamilkrzyskow:
Should fix the crash when running `mike` with the `group` plugin.
Related discussion: https://github.com/squidfunk/mkdocs-material/discussions/6520

I didn't change anything unrelated to the crash, I shall wait for https://github.com/squidfunk/mkdocs-material/pull/6473 to have some sort of resolution to propose bigger styling changes :v:


squidfunk:
Perfect, thanks! Yes, we'll be moving to a more Python-idiomatic code base in the next year. Please don't create PRs yet, because we need to do this gradually. I'll be starting with the rewrite of the tags plugin.

PR #6541, opened 2023-12-19
Replaced mutable argument default by `None`

sisp:
I've fixed the use of a mutable argument default (here, a `dict` literal), which _can_ introduce bugs that are a bit hard to debug. Here, it doesn't seem to be a problem because `self.attrs` isn't mutated, but it's generally discouraged to use mutable argument defaults. See https://docs.astral.sh/ruff/rules/mutable-argument-default/#mutable-argument-default-b006 for more details.

### DIFF
--- a/src/plugins/search/plugin.py
+++ b/src/plugins/search/plugin.py
@@ -312,9 +312,9 @@ class Element:
     """
 
     # Initialize HTML element
-    def __init__(self, tag, attrs = {}):
+    def __init__(self, tag, attrs = None):
         self.tag   = tag
-        self.attrs = attrs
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks! Next time, please build the theme and also make sure that your changes are reflected in the `material` folder. For the plugins, it's mandatory to build the theme as well 😉

squidfunk:
FWIF, I'm referring to 175df6139.

sisp:
Oh, I wasn't aware that changes to the plugins also need a fresh build. 🫣 I'll try to remember it. :innocent:

PR #6549, opened 2023-12-21
added section on setting up navigation

alexvoss:
This is in response to https://github.com/squidfunk/mkdocs-material/discussions/6542 to add a section on setting up navigation for a blog so that the archive and category pages appear in the right place.

Hope I got the logic and level of detail right. Any feedback very welcome.


squidfunk:
Perfect, LGTM!

PR #6564, opened 2023-12-26
Add testing environments for Python

ofek:
As requested elsewhere from you, here is the initial config to set up a test matrix. You can enter a shell into the default environment (which will use the highest Python that can be found on PATH) by running `hatch shell` and test scripts can be executed like `hatch run cov`. The matrix can run tests in each like `hatch run test:cov` or individually like `hatch run test.py3.12:cov`.

- I am aware that there is no such tests directory yet, this is just to show how it would happen
- I haven't taken a look at CI so let me know if/when you would like me to add stuff there
- This uses pytest (rather than unittest as I saw in the other commit to insiders) as that is basically the standard for testing in Python now

### DIFF
--- a/hatch.toml
+++ b/hatch.toml
@@ -1,0 +1,21 @@
+[envs.default]
+dependencies = [
+  "coverage[toml]>=6.5",
+  "pytest",
+  "pytest-mock",
+  "pytest-randomly",
+]
+[envs.default.scripts]
+test = "pytest {args:tests}"
+test-cov = "coverage run -m pytest {args:tests}"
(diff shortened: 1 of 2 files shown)


squidfunk:
Thanks for much! However, could you do the PR on Insiders? We'll start with testing there, since Insiders is our "`master`", containing all plugins, not only the ones available in the community edition ☺️ Once we have a good coverage there, we'll merge it into the community edition as well, but it's not economical to maintain two test code bases right now.

ofek:
Gotcha!

PR #6584, opened 2024-01-03
Update creating-your-site.md

DanielHabenicht:
I just wanted to clarify some confusion I had while setting up my documentation site. 

As I am on Windows I tried the Windows option first only to find out later that this command was specific to the command line and not the default terminal (Powershell) in Windows 11.

This should make it clear that this command is aimed only at Windows CommandLine users and not generally Windows users.

I was unsure if I should open a PR for this, as it is such a small change. If it is no welcome just close it - no need for discussion.

### DIFF
--- a/docs/creating-your-site.md
+++ b/docs/creating-your-site.md
@@ -16,7 +16,7 @@ Alternatively, if you're running Material for MkDocs from within Docker, use:
     docker run --rm -it -v ${PWD}:/docs squidfunk/mkdocs-material new .
     ```
 
-=== "Windows Commandline"
+=== "Windows (cmd)"
 
     ```
     docker run --rm -it -v "%cd%":/docs squidfunk/mkdocs-material new .
--- a/docs/creating-your-site.md
+++ b/docs/creating-your-site.md
@@ -16,7 +16,7 @@ Alternatively, if you're running Material for MkDocs from within Docker, use:
     docker run --rm -it -v ${PWD}:/docs squidfunk/mkdocs-material new .
     ```
 
-=== "Windows"
+=== "Windows Commandline"
 
     ```
     docker run --rm -it -v "%cd%":/docs squidfunk/mkdocs-material new .


squidfunk:
Thanks for the PR! I'm not sure whether this really helps to further clarify. Maybe some other users can comment whether they ran into the same problems.

kamilkrzyskow:
As the command above already states that PowerShell will support the `${PWD}` notation, I don't think that adding Commandline to the bottom one with the `%cd%` will help much. Especially since `cmd` can be referred to as Command Prompt / Processor, also personally I never seen Commandline written together (referring to `cmd`). So it's hard to decide 🤔 
Windows 10 and 11 heavily promote PowerShell, so people using `cmd` is becoming a rarity, at least I think so.

So I would prefer `Windows (cmd)` to give more clarity. Sorry for adding my 2 cents on a single line change @DanielHabenicht 😅

DanielHabenicht:
That's fine. I also first thought about using `Windows (cmd)`. I will chang it.

squidfunk:
Thanks to you both. Honestly, I'm not a fan of adding multiple instructions for the same operating system and making the tabs longer and longer to scroll. Windows seems really odd to work with.

Can we maybe somehow leverage icons for disambiguation?

kamilkrzyskow:
I don't feel like 2-3 tabs add that much scrolling 🤔 
Since Windows 10 (some specific iteration) PowerShell is the default "Terminal", so in this case specifically, we could just remove the second tab for `Windows (cmd)` and set the title for the previous code block to `Unix / Windows (PowerShell)` and keep only one example. PowerShell could be also assumed the one and only default for the rest of the documentation.

https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#adding-a-title

However, in other instances where it's not about syntax and more about the actual path difference or some other thing and 2 tabs are required at minimum we could use icons with the custom hover title tooltips 🤩 I'm for using the `simple-windows10` (11 is just a square 🟦) and `simple-linux` icons. 

<details open>
<summary>Windows Icons</summary>

image

</details>


<details open>
<summary>Linux Icons</summary>

image

</details>

squidfunk:
Honestly, "Unix / Windows (Powershell)" feels even more weird 😅 Not a fan of Windows.

kamilkrzyskow:
> Not a fan of Windows

As a daily driver of Windows 10, I can say the same thing about Linux, after multiple tries to switch over. 
Never used macOS, way too expensive and overpriced for the most part, M1/M2 improved their image for me, but still too expensive.

To get back on topic 😅, if "Unix / Windows (Powershell)" isn't acceptable just use "Unix / PowerShell" with the capital S, or just "Terminal", and hope that since PowerShell is the default of Windows, there won't be many `cmd` users still around creating issues. Personally, over the years, I always preferred the old-school `cmd` over PowerShell, but nowadays I primarily use Git Bash for Windows with the occasional odd PowerShell script.

So IMO either accept the clarification of `Windows (cmd)` or delete the second example and leave only one for "Unix / PowerShell"

squidfunk:
Okay, thanks for the explanation. While I'm not yet happy with how things are not clearly cut, having two tabs that account to Windows, we can merge this as a bandaid and fix it later when we rework that section of the documentation.

squidfunk:
Oh, and I didn't want to bash Windows. I'm just not a fan, which was meant to outline my viewpoint. It clearly has its place in the desktop operating space. We try to support Windows as good as we can, but we currently need to rely on outside collaborators, because I'm not a user of it.

kamilkrzyskow:
Sure thing, also it didn't look like bashing Windows, and I really didn't plan to defend Windows via complaining about macOS/Linux, it just turned out like that when I read it again😅
I wanted to say that from a yearlong Windows users perspective it's "normal" instead of confusing to have both `cmd` and PowerShell, and for the most part Terminal usage isn't all that prevalent, so it's worth to make a clear distinction between them (as the merged commit).

Give it a few years, Windows 12 or 13 will come around and maybe PowerShell will be the only available Terminal app in the system.
I know that now a lot of developers just use Docker with WSL, so even on Windows they use Linux.
And even before that e.g. The Odin Project pushed students to use Linux on virtual machines, as most of their course work tools didn't work as well on Windows, as with Linux.

There will come a time when Unix-terminals will be the norm and developers will live in peace 😆

DanielHabenicht:
> There will come a time when Unix-terminals will be the norm and developers will live in peace 😆

Nice utopia. I think Copilot will take over Windows before that. The Documentation then has to include the prompt 😆 
"Dear Windows, please initialize a new MKDocs Page."

squidfunk:
> Dear Windows, please initialize a new MKDocs Material Page.

At least, MkDocs could remain relevant for some time for feeding knowledge to the machines 🤖

PR #6598, opened 2024-01-04
Update "short" date to be consistent

nicfv:
Update the "short" date to 1/31/23 since it appears that all other dates on this page are January 31, 2023.

image


squidfunk:
Thanks for the PR! Could I maybe ask you to update all dates to 2024? That would be an even better PR ☺️

nicfv:
Sure, why not! Done and done.

I couldn't find any other pages where it would make sense to change the date. Let me know if you think I missed something.

squidfunk:
Awesome, thanks for helping us keeping the docs up-to-date!

PR #6606, opened 2024-01-05
Fix typo in SmartSymbols section

JimmyCHWang:
The text following "e.h." are listing examples of SmartSymbols, so "e.h." could be a typo for "e.g." in the description.


squidfunk:
Jup, good catch, thanks!

PR #6612, opened 2024-01-06
Explicitly mention setting defaults is for insiders only (meta plugin)

sanspace:
As the meta plugin is reserved for insiders only, this section could also mention that so people don't run into the below error.

```
$ mkdocs serve
ERROR   -  Config value 'plugins': The "meta" plugin is not installed

Aborted with 1 configuration errors!
```

I am not sure if I'm adding all the labels accurately. I got it from the meta plugin page under [configuration](https://squidfunk.github.io/mkdocs-material/plugins/meta/#configuration).  Please edit as you see fit. Thanks.

### DIFF
--- a/docs/setup/setting-up-a-blog.md
+++ b/docs/setup/setting-up-a-blog.md
@@ -535,6 +535,11 @@ This will disable automatic reading time computation.
 
 #### Setting defaults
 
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks, makes sense!

PR #6653, opened 2024-01-15
fix: Custom Elements support for getActiveElement

### LINKED ISSUE #6652: Add support for Custom Element
### Context

I added Custom Elements with shadow root via custom elements because I don't want the styles of mkdocs has influence on forms I am adding. When I have a form inside my custom component, the global keyboard handler catches some keys I am typing like `s` and takes focus to search.
The reason is if there is input focused inside the shadow root, the parent's `document.activeElement` would point to the custom element itself, not to it's underlying element.

Since 'the bug' is caused by extra javascript, I cannot mark it as a bug, so I am asking about adding support for custom elements.

### Description

While having additional functionality via extra_javascript is a great feature, it lacks a full support for adding custom elements. I was able to add the custom element, but the keyboard shortcut is not aware about such type of element. 
There is a check for activeElement which works differently for elements with shadowRoot. I am asking to add support for such kind of elements here:
https://github.com/squidfunk/mkdocs-material/blob/44982451f7759ed65e31d1151b3c49566b8c6ca0/src/templates/assets/javascripts/browser/element/_/index.ts#L116-L120

The active element of shadowRoot can be accessed via `document.activeElement.shadowRoot.activeElement`.

### Related links

- [Custom elements](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements)
- [Shadow DOM v1 - Self-Contained Web Components # Handling Focus](https://web.dev/articles/shadowdom-v1#handling_focus).

### Use Cases

It will allow to have forms inside custom elements

### Visuals

_No response_

### Before submitting

- [X] I have included relevant links to [the documentation](https://squidfunk.github.io/mkdocs-material/), related [issues](https://github.com/squidfunk/mkdocs-material/issues), and [discussions](https://github.com/squidfunk/mkdocs-material/discussions) to underline the need for my idea.

vtsvetkov-splunk:
## Before changes:

Focusing shadowRooted input vs regular input
image

After changes:
image

Unfortunately, that would not work if shadowRoot has closed mode. In this case internals of the custom element is not accessible via an external document.

The issue is described here: [Shadow DOM v1 - Self-Contained Web Components](https://web.dev/articles/shadowdom-v1#handling_focus).

Should we check recursively?

Resolves #6652

### DIFF
--- a/src/templates/assets/javascripts/browser/element/_/index.ts
+++ b/src/templates/assets/javascripts/browser/element/_/index.ts
@@ -114,6 +114,12 @@ export function getOptionalElement<T extends HTMLElement>(
  * @returns Element or nothing
  */
 export function getActiveElement(): HTMLElement | undefined {
+  const isHTMLElement = document.activeElement instanceof HTMLElement;
+  const hasShadowRoot = isHTMLElement && document.activeElement?.shadowRoot;
+
+  if (hasShadowRoot) {
+    return document.activeElement?.shadowRoot?.activeElement || undefined;
+  }
   return document.activeElement instanceof HTMLElement
     ? document.activeElement || undefined
     : undefined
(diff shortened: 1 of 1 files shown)


vtsvetkov-splunk (commented):
reuse var

squidfunk:
Superseded by be95f4901 – a simpler implementation. Tagged you as a co-author, @vtsvetkov-splunk .

Issue #6656, opened 2024-01-15
Add docs on Deploying mkdocs to fly.io


userbradley:
### Context

At the bottom of the `publishing-your-site` page there is a section of community provided sites.

I would like to add my writeup for `fly.io` to this 

### Description

I've written a kb article on deploying mkdocs to fly, as my site runs on fly.

In my KB I've mentioned that if users have issues they are to reach out, but I think I may clarify this is to reach out to me (@userbradley ) opposed to you as it's not officially supported. Let me know if the wording needs to be fixed

### Related links

- https://documentation.breadnet.co.uk/cloud/fly/mkdocs-on-fly/

### Use Cases

Helping users to publish mkdocs on another hosting provider

### Visuals

I could not find the `fly.io` icon on SimpleIcons, so I am no sure what the next port of call is. I may open a feature request on SI but that may take a while

### Before submitting

- [X] I have read and followed the [change request guidelines](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/).
- [X] I have verified that [my idea is a change request and not a bug report](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/#its-not-a-bug-its-a-feature).
- [X] I have ensured that, to the best of my knowledge, [my idea will benefit the entire community](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/#benefit-for-the-community).
- [X] I have included relevant links to [the documentation](https://squidfunk.github.io/mkdocs-material/), related [issues](https://github.com/squidfunk/mkdocs-material/issues), and [discussions](https://github.com/squidfunk/mkdocs-material/discussions) to underline the need for my idea.

squidfunk:
Thanks for suggesting. You can create a PR on the publishing guide, if you like, and add your guide in alphabetical order.

PR #6663, opened 2024-01-18
[Publishing guide] Adding Fly.io

userbradley:
This PR adds a link to a guide I wrote on my docs site on how to Deploy mkdocs-material to Fly.io

Implements https://github.com/squidfunk/mkdocs-material/issues/6656 

---

Commits are signed to verify I contributed for any license etc


squidfunk:
Thanks! Could we use `:material-airballoon:` as an icon until we get an official logo in Simple icons? It is quite similar and would be more consistent with the style we currently have.

userbradley:
Added that. Good catch @squidfunk! Let me know if there's anything else in the mean time!

squidfunk:
LGTM.

PR #6668, opened 2024-01-21
Add Codeberg pages guide to "Publishing your site"

Andre601:
Made a guide explaining how to publish your MkDocs page using Codeberg Pages and woodpecker-ci. May be of use for anyone who uses that repo host and wants to use MkDocs.


squidfunk:
Thanks!

PR #6695, opened 2024-01-25
feat: Improve image visibility/UX DOCS-649 

nicklem:
Highlights image borders with a box-shadow.


nicklem:
Wrong destination repo, apologies

PR #6720, opened 2024-01-31
Update GitHub Actions workflow

uschmidt83:
Update action versions for 'setup-python' and 'cache'.

Set 'fetch-depth: 0' for 'checkout' action to obtain correct document dates when using 'mkdocs-git-revision-date-localized-plugin' (cf. https://github.com/timvink/mkdocs-git-revision-date-localized-plugin/blob/v1.2.2/mkdocs_git_revision_date_localized_plugin/ci.py#L37-L48)

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -35,15 +35,17 @@ contents:
         runs-on: ubuntu-latest
         steps:
           - uses: actions/checkout@v4
+            with:
+              fetch-depth: 0
           - name: Configure Git Credentials
             run: |
               git config user.name github-actions[bot]
               git config user.email 41898282+github-actions[bot]@users.noreply.github.com
-          - uses: actions/setup-python@v4
+          - uses: actions/setup-python@v5
             with:
               python-version: 3.x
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for suggesting. There are two changes in this PR that need discussion. We're happy to update the versions of the GitHub Actions, but we should not add fetch depth. The reason is that not all users are using the git revision date plugin, and adding fetch depth will significantly slow down your build, as it now fetches the entire history. This _must_ be a deliberate decision. The publishing your site guide aims to be as minimal as possible.

If you remove this change, we're happy to merge your PR.

uschmidt83:
> The reason is that not all users are using the git revision date plugin, and adding fetch depth will significantly slow down your build, as it now fetches the entire history. 

Sure. I didn't realize that it would (cumulatively) be a big deal.

I also overlooked the _When using build environments_ box under the [Document dates](https://squidfunk.github.io/mkdocs-material/setup/adding-a-git-repository/#document-dates).

squidfunk:
Thanks! Yeah, it would be a huge waste of resources.

PR #6727, opened 2024-02-01
fix: error on empty fields in search index

Philip-Larsen-Donnelly:
If fields in the search index exist but are empty, the search can throw an error and return no results. This can happen, for example, if the markdown includes a section heading without any text, resulting in an empty "text" field in the search index. If the search term is found in the section heading "title" field, the highlighting code will then throw an error when processing the "text" field.

### DIFF
--- a/src/templates/assets/javascripts/integrations/search/_/index.ts
+++ b/src/templates/assets/javascripts/integrations/search/_/index.ts
@@ -244,7 +244,7 @@ export class Search {
 
           /* Highlight matches in fields */
           for (const field of this.index.fields) {
-            if (typeof doc[field] === "undefined")
+            if (typeof doc[field] === "undefined" || doc[field].length === 0)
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. Could you please first open an issue with a minimal reproduction, so we can understand what's happening? Before merging such changes, we need to reproduce them and assess the impact. Thanks!

Philip-Larsen-Donnelly:
Thanks for the response. So, now I tried reproducing the problem on a vanilla install of mkdocs and mkdocs-material, and I can't. I have quite a hacked version of the search, but this seemed a very low-level issue.
I'll close this PR while I don't have a way to reproduce the problem in a standard environment.
Sorry about that!

squidfunk:
Thanks! We may have fixed this in a subsequent release.

PR #6737, opened 2024-02-04
Github Action Plugin Support

TinnyTerr:
I know that it has a footnote explaining how to add this to the action however people, including me when I had not thought about this reason the action was failing, may easily miss this and not think to come back to it purely on the thought that its been there for a while and should be correct.

Overview of the small change being:

Adding 
` - run: pip install $(echo "mkdocs get-deps")` 
and updating the dropdown footnote attached to it. I see no problem but if you'd prefer to keep it the same then go ahead.
(I'd love a look over the footnote as I'm not the best at documenting the best)

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -49,7 +49,7 @@ contents:
               path: .cache
               restore-keys: |
                 mkdocs-material-
-          - run: pip install mkdocs-material 
+          - run: pip install mkdocs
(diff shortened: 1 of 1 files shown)


TinnyTerr:
After further inspection, this iteration doesn't work. I'm closing this until I find a working one as this outputs 
`pip install get-deps` instead of the expected plugins

PR #6753, opened 2024-02-08
Added a emoji

sounar97:
https://github.com/pydantic/pydantic/issues/8743 
Tried to solve the above issue by adding the emoji.


squidfunk:
Thanks for suggesting. However, we're currently not considering merging this PR.

PR #6765, opened 2024-02-11
docs: fix typos

AlessandroMiola:
Fix minor typos


squidfunk:
Thanks for the PR. However, this is not correct. The label indicates what mode will be switched to when it is clicked. Thus, no error or problem on our side and we rather keep it like it is.

PR #6787, opened 2024-02-15
Sort pages on tag list by title

RealOrangeOne:
The list seems to be ordered by URL, which doesn't necessarily match the title, which can look odd. Instead, explicitly sort by title, which is shown in the list anyway.

### DIFF
--- a/src/plugins/tags/plugin.py
+++ b/src/plugins/tags/plugin.py
@@ -19,6 +19,7 @@
 # IN THE SOFTWARE.
 
 import logging
+import operator
 import sys
 
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! However, note that sorting is [already implemented in Insiders](https://squidfunk.github.io/mkdocs-material/plugins/tags/#config.listings_sort_by) as a setting. I'm not sure we should change the default sorting (it would impact many projects), and other users might prefer other ways of sorting. That's why we added it as a setting, so everybody can configure it to its needs.

When [the next funding goal](https://squidfunk.github.io/mkdocs-material/insiders/#16000-chipotle) is hit, the new tags plugin will be released into the community edition, which includes configurable listings – you'll be able to sort tags and configure listings + define an allowed list of tags. You can use it immediately if you decide to [support our work by sponsoring](https://github.com/sponsor/squidfunk).

Please understand that we don't regard it to be a good idea to change default behavior when the new plugin is just around the corner, as we'd be making more work for us, as other users have other opinions.

PR #6800, opened 2024-02-19
Update Finnish translation

todeveni:
`blog.categories.in` is intentionally empty due to finnish grammar 🙂


squidfunk:
Excellent thanks! However, those changes need to be made in [`src/templates/partials/languages/fi.html`](https://github.com/squidfunk/mkdocs-material/blob/61430da21437b1c2728cb70ce4122b2f83ce5ae2/src/templates/partials/languages/fi.html), as otherwise your changes will get overwritten on the next build, as [mentioned in our documentation](https://squidfunk.github.io/mkdocs-material/customization/#development-mode) ☺️

todeveni:
My bad, new PR coming up 🙂

PR #6801, opened 2024-02-19
Update finnish translation

todeveni:
This time edit a file that doesn't have `This file was automatically generated - do not edit` in it


squidfunk:
Almost right - in fact, we need both. This is why you need to go through the theme development process. However, never mind, I'll build the theme locally and push the generated files. Thanks again!

PR #6803, opened 2024-02-19
Update icons-emojis.md

jjasghar:
The link no longer works, it seems X has removed it completely.


squidfunk:
Good catch. Meh. Could you replace the link to point to https://github.com/twitter/twemoji?

jjasghar:
Fixed :)

squidfunk:
Thanks!

Issue #6817, opened 2024-02-22
Social plugin makes Mkdocs build crashing when system dependencies are not met


Guts:
### Context

Our Mkdocs website is collaborative and many of writers are using Windows without any ability to install the CairoSVG dependencies at system level (https://squidfunk.github.io/mkdocs-material/plugins/requirements/image-processing/#cairo-graphics).

### Bug description

Expected behavior: try/except the exception raised by cairosvg and display a error, warning or even info message to inform that social cards have been disabled, but let the build going on.

### Related links

- https://squidfunk.github.io/mkdocs-material/setup/setting-up-social-cards/
- https://squidfunk.github.io/mkdocs-material/plugins/requirements/image-processing/#cairo-graphics

### Reproduction

Sorry I cant' generate a minimal reproduction since the build is failing. Still, here comes a manual zip: [material-social-card-build-crash.zip](https://github.com/squidfunk/mkdocs-material/files/14378567/material-social-card-build-crash.zip)

### Steps to reproduce

Install:

```powershell
pip install --upgrade --force-reinstall mkdocs-material[imaging]
```

Given this mkdocs.yml:

```yaml
site_name: My Docs
site_url: https://example.com

theme:
  name: material

plugins:
  - info
  - social:
      enabled: true
      cards: true

```

Running build:

```powershell
(.venv) PS C:\Users\username\Documents\GitHub\Perso\material-social-card-build-crash> mkdocs build
Traceback (most recent call last):
  File "C:\Users\username\AppData\Local\Programs\Python\Python310\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\username\AppData\Local\Programs\Python\Python310\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\Users\username\Documents\GitHub\Perso\material-social-card-build-crash\.venv\Scripts\mkdocs.exe\__main__.py", line 7, in <module>
  File "C:\Users\username\Documents\GitHub\Perso\material-social-card-build-crash\.venv\lib\site-packages\click\core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "C:\Users\username\Documents\GitHub\Perso\material-social-card-build-crash\.venv\lib\site-packages\click\core.py", line 1078, in main
    rv = self.invoke(ctx)
[...]

PR #6818, opened 2024-02-22
fix: handle OSError importing cairosvg to avoid build crash because of missing system dependencies

Guts:
Closes: #6817


squidfunk:
Thanks for the PR. However, we're not considering merging it.

The reason is that if we wouldn't throw the error, it might go silent in CI builds, with users wondering why social cards are not building. We're already experiencing high demand in support for setting up Cairo, and this would lead to even more issues why social cards are not building, as users might forget to install the dependencies, so the build will succeed. Thus, we cannot succeed the build if the dependencies are not available.

A better solution is to only enable the social cards plugin only in CI using an [environment variable](https://www.mkdocs.org/user-guide/configuration/#environment-variables), as the `CI` environment variable is already defined in most CI build environments, e.g., GitHub Actions and GitLab:

``` yaml
plugins:
  - social:
      enabled: !ENV CI
```

This will disable generation of social cards locally (which makes no sense anyway, as you cannot post local links to Twitter or other social media platforms) and keep them when built within CI. We always need to keep the economics of maintaining the project in sight when considering PRs for merging. I hope for your understanding. Thank you.

Guts:
I get your rationale.  
However, there are 2 commits in my PR precisely because I wanted to separate :

1. the [first](ee8a6480d60d5ab0090b3ef2f1c9a8ad8c7c935a) prevents mkdocs from crashing (cf the issue) but let the actual plugin's behavior (exiting the build with an exception)
1. the [second](30b1d6cee5344e7f4354f4eb8b77232310996ca1) suggests avoiding the error and continuing the build; which will fail if the user has opted for strict mode, which is how the Mkdocs ecosystem works.

According to your argument, I can understand that the second point is debatable and why not assume that end users are not aware of strict mode and don't read the logs (I have the same experience on my own projects). But the former should be integrated to reduce support queries following a mkdocs build crash, shouldn't it?

For the record, I'm already using the environment variable to activate/deactivate the plugin, but I haven't figured out how to activate it only on a Linux environment or depending on the presence of Cairo dependencies.

squidfunk:
Okay! Let's fix the `OSError` then – could you remove the second commit or open a new PR?

kamilkrzyskow:
I see that @Guts has reacted, but no change 🤔So was "you" in "could you remove the second commit or open a new PR?" directed at me specifically or both of us 😅 
Will you adjust the PR @Guts or am I free to do it myself? I think I'm eligible to make changes with "Maintainers are allowed to edit this pull request. ", because I have push rights 🤔

Guts:
Sorry, I did not have time past days and not much either in coming next.

So feel free to make the changes!

squidfunk:
It was targeted towards @Guts, the OP of the PR, but I don't mind you making the changes, @kamilkrzyskow ☺️

kamilkrzyskow:
I made the errors a bit more explicit:
```bash
ERROR   -  Required dependencies of "social" plugin not found.
Install with: pip install "mkdocs-material[imaging]"
No module named 'PIL'
No module named 'cairosvg'
```

```bash
ERROR   -  "cairosvg" Python module is installed, but it crashed with:
no library called "cairo-2" was found
no library called "cairo" was found
no library called "libcairo-2" was found
cannot load library 'libcairo.so.2': error 0x7e.  Additionally, ctypes.util.find_library() did not manage to locate a library called 'libcairo.so.2'
cannot load library 'libcairo.2.dylib': error 0x7e.  Additionally, ctypes.util.find_library() did not manage to locate a library called 'libcairo.2.dylib'
cannot load library 'libcairo-2.dll': error 0x7e.  Additionally, ctypes.util.find_library() did not manage to locate a library called 'libcairo-2.dll'
---
Did you install the external CairoSVG library?
Did you add the binaries path to your system PATH?
Did you restart the Terminal to assure the PATH was reloaded?
```

material/plugins/social/plugin.py
```diff
@@ -49,13 +49,24 @@
...
 from zipfile import ZipFile
+
+from .config import SocialConfig
+
 try:
-    from cairosvg import svg2png
     from PIL import Image, ImageDraw, ImageFont
-except ImportError:
-    pass
+except ImportError as e:
```
squidfunk, line 57: So I'm not sure I'm for using this import error concatenation pattern, but I'm short on time now thinking of a better approach, so I'm happy to test it out. Is that a common practice?

material/plugins/social/plugin.py
```diff
@@ -76,10 +87,20 @@ def on_config(self, config):
...
+                "Required dependencies of \"social\" plugin not found.\n"
+                "Install with: pip install \"mkdocs-material[imaging]\"\n"
+                + str("\n".join(DEP_IMPORT_ERRORS))
+            )
+
+        if CAIROSVG_ERROR:
             raise PluginError(
-                "Required dependencies of \"social\" plugin not found. "
-                "Install with: pip install \"mkdocs-material[imaging]\""
+                "\"cairosvg\" Python module is installed, but it crashed with:\n"
```
squidfunk, line 99: What if we create a troubleshooting section in our image processing documentation and link to it? I think this might be a better approach because we have essentially infinite space as opposed to a terminal, it is easier to maintain and format, and we can cater to more OSs.

We then would only include the link here.

squidfunk (changes requested):
Thanks for your work! I'd say we move out the troubleshooting info to our documentation.

material/plugins/social/plugin.py
kamilkrzyskow: > Is that a common practice?

Converting the error to string is a valid practice, but I don't know if it's common. The idea with gathering the errors to print them out later came from MkDocs <-> Plugin relationship. Errors that are local to the plugin should be properly moved upstream to the event flow and I wanted to be more specific about the message that says "dependencies not installed", so I needed to store the errors from the config load to report them later on.
If you feel it's weird I can just revert it :v:

material/plugins/social/plugin.py
kamilkrzyskow: OK, I will write up a small troubleshooting guide and put a link. Should I use the t.ly link shortener again?

material/plugins/social/plugin.py
squidfunk: Let's just keep it like that for now and revisit it later, all good! I just wanted to ask.

material/plugins/social/plugin.py
squidfunk: Yes, [t.ly](https://t.ly/) is the one we're using because it seems to be the most sensible to use right now.

material/plugins/social/plugin.py
kamilkrzyskow: Funnily enough, when researching things for the troubleshooting guide I found out that the cairocffi (dependency of cairosvg) also gathers the exceptions and the result we see are a mix of formatted strings and errors joined together 😅 

So the commonality increased :v:
https://github.com/Kozea/cairocffi/blob/f1984d644bbc462ef0ec33b97782cf05733d7b53/cairocffi/__init__.py#L39-L40

kamilkrzyskow:
The current errors look like this:
```bash
ERROR   -  Required dependencies of "social" plugin not found:
           - ModuleNotFoundError("No module named 'PIL'")
           - ModuleNotFoundError("No module named 'cairosvg'")

           --> Install with: pip install "mkdocs-material[imaging]"
```
```bash
ERROR   -  "cairosvg" Python module is installed, but it crashed with:
           no library called "cairo-2" was found
           no library called "cairo" was found
           no library called "libcairo-2" was found
           cannot load library 'libcairo.so.2': error 0x7e.  Additionally, ctypes.util.find_library() did
           not manage to locate a library called 'libcairo.so.2'
           cannot load library 'libcairo.2.dylib': error 0x7e.  Additionally, ctypes.util.find_library()
           did not manage to locate a library called 'libcairo.2.dylib'
           cannot load library 'libcairo-2.dll': error 0x7e.  Additionally, ctypes.util.find_library() did
           not manage to locate a library called 'libcairo-2.dll'

           --> Check out the troubleshooting guide: https://t.ly/dlwQ7
```

The troubleshooting guide first and foremost recommends to restart the Terminal windows to reload PATH and stuff.
Then goes into a bit more detail, as best as I could write it out. I didn't check the macOS script yet (with GitHub Actions would be my only option). Perhaps I should provide some simple command instead of the script 🤔 
In Windows Powershell there is 
```powershell
$env:Path -split ';'
```
and it will display a list, just like the one in the Python script, but without the caironames etc.

I'll change the state of the PR to draft, since I wait for feedback :v:

kamilkrzyskow:
Made a test repository and run the scripts on each system:
https://github.com/kamilkrzyskow/test-macos-actions/actions/runs/8103418921

Everything works as expected, however it kind of sucks I can't create a simple Linux script for easier troubleshooting.
Unless I would create it specifically for Ubuntu / GitHub Actions and tell other disctro users to deal with it on their own.

---

I also think I might've misunderstood and the troubleshooting guide should've been at the bottom of the page, not inside the cairo section? 🤔

docs/plugins/requirements/image-processing.md
```diff
@@ -93,6 +93,189 @@ The following environments come with a preinstalled version of [Cairo Graphics]:
   [Docker image]: https://hub.docker.com/r/squidfunk/mkdocs-material/
   [GitHub Actions]: ../../publishing-your-site.md#with-github-actions
 
+#### Troubleshooting { id="cairosvg-troubleshooting" }
```
squidfunk, line 96: Why the anchor name? I'd say we'll keep it as troubleshooting to make room for more to add later + keep the anchor links stable. If we feel that we need more granularity, we can move troubleshooting one way up and add further headlines.

docs/plugins/requirements/image-processing.md
```diff
@@ -93,6 +93,189 @@ The following environments come with a preinstalled version of [Cairo Graphics]:
...
+    to maybe find the issue.
+
+    A [known workaround][cffi-issue] is to add the homebrew lib path
+    directly before running MkDocs:
+
+    ```bash
+    export DYLD_FALLBACK_LIBRARY_PATH=/opt/homebrew/lib
+    ```
+
+    ??? tip "Python Debug macOS Script"
```
squidfunk, line 138: Just an idea: we _could_ create a `etc/debug` folder in our repository and move the script there. That way we can version it + we could, if necessary, use Snippets to integrate it. In the future, we might also somehow move this logic into the social plugin (or better: into our own utilities folder, so we can use it in other plugins as well) and/or bundle the script the theme (not now, let's keep it simple for now).

What do you think? I'm not sure if this makes sense, just wanted to bounce ideas.

docs/plugins/requirements/image-processing.md
kamilkrzyskow: OK, so I misunderstood, I'll move the Troubleshooting to the bottom of the page as another H2 tag, a sibling of Dependencies, and change the link to `#troubleshooting` 👌

docs/plugins/requirements/image-processing.md
kamilkrzyskow: There is already the https://github.com/squidfunk/mkdocs-material/tree/master/includes directory. We could move the scripts there into the `etc/debug` structure, and later include them with Snippets, as you said. 

I'm still not sure about the Linux scripts, because GitHub Actions finds the library with https://www.unix.com/man-page/Linux/1/ld/ but I'm not sure if that will always be the case with the cairo library or if it depends on how someone installs it 😑 

> social plugin (or better: into our own utilities folder, so we can use it in other plugins as well) and/or bundle the script the theme (not now, let's keep it simple for now).

The scripts I provided are supposed to display the paths that are checked or the way they're checked for the user that wants to troubleshoot them. They are somewhat generalized, since they run with the `ctypes.util` built-in and not cairoffi directly, but I'm doubtful they would be useful as utils in the plugins. 

Another idea I had, was to ask you to create a mkdocs-material/troubleshooting repository and link to it from that troubleshooting guide, and there would be CI similar to my own repository, which people could copy to execute the actions and see the scripts in action, but this idea is a bit "out there" doesn't really focus on the matter at hand of getting it to work on the users PC, and requires "GitHub Actions proficiency" of the user.

squidfunk: Yes, let's move it into `include/debug` and use Snippets for now.

> Another idea I had, was to ask you to create a mkdocs-material/troubleshooting repository and link to it from that troubleshooting guide, and there would be CI similar to my own repository, which people could copy to execute the actions and see the scripts in action, but this idea is a bit "out there" doesn't really focus on the matter at hand of getting it to work on the users PC, and requires "GitHub Actions proficiency" of the user.

Interesting idea, I'll discuss it with the core team next time we speak ☺️

kamilkrzyskow:
Thanks to Snippets the Markdown became more readable, and allowed for a bit longer Linux script.
I've focused on Ubuntu and Manjaro, as the other BSD or SunOS Linux variants aren't that common.
Moved the structure and adjusted the t.ly link in the scripts.

The updated scripts can be seen running here:
https://github.com/kamilkrzyskow/test-macos-actions/actions/runs/8159277013

kamilkrzyskow:
Did another force push to span out the text to 80 characters in some places,

Guts:
Impressive and really valuable work 👏!

Thanks @kamilkrzyskow!

kamilkrzyskow:
The last thing I'm unsure about, as this wasn't in the change request before, is that I did set CAPITAL-cased global variables, and from the rest of the code base it seems that global `log` variables are lowercase, so there is no distinction between global and local case?

squidfunk (approved):
LGTM! Thanks for putting the time in, great work!

One final remark when viewing the documentation though: I think we should not use Snippets, as the scripts are very long and that adds a lot of unnecessary noise to the documentation. This was my fault, recommending it but I also did not know the scripts will get so long. The scripts do not _explain_ anything, and are more or less intended to be "just run" by the user helping with troubleshooting, so I would say we just link to the `raw.githubusercontent` versions of them, ideally directly triggering a download if that is possible.

kamilkrzyskow:
> The scripts do not explain anything, and are more or less intended to be "just run" by the user helping with troubleshooting

Yes, that was my intention from the start, as I believe adding a lot of comments / print statements, would just make it longer, and a non-techy person would still not read it before running it. There is only one print at the end, saying if it found the path. 

> so I would say we just link to the raw.githubusercontent versions of them, ideally directly triggering a download if that is possible.

I will look into a command to download the script and pipe them to Python. I've seen some PowerShell scripts do that, but without Python. I think the same should be possible with a `curl` command on Unix.

squidfunk:
> I will look into a command to download the script and pipe them to Python. I've seen some PowerShell scripts do that, but without Python. I think the same should be possible with a curl command on Unix.

Perfect. Also link to the script that is downloaded via curl, so users can inspect it. Anything else would be fishy.

kamilkrzyskow:
Done and a test here: https://github.com/kamilkrzyskow/test-macos-actions/actions/runs/8227806730
Ultimately, I also changed the UPPERCASE variables to lowercase, as I stumbled upon this in the search.py:
https://github.com/squidfunk/mkdocs-material/blob/c8bae9b5b4b46f0032a9f2dddb30adda68158c78/src/plugins/search/plugin.py#L33-L36
Not entirely the same, but still consistent with the lowercase `log`, so all variables are lowercase in the codebase, I hope I didn't mess it up now, as the previous review accepted the uppercased variables...

squidfunk (approved):
LGTM. I need to find a time slot to be able to merge this through to Insiders, but since the social plugin is entirely different there, I expect it to take some time. Please bear with me.

kamilkrzyskow:
I can also make a PR towards the Insiders branch, if you want to to have the dependency logic the same there, but since the plugins have branched out into two different versions, I don't think they have to merge the change at the same time.

kamilkrzyskow:
I had a look at the source code in Insiders, and I see that the dependency check and related error was moved out of `on_config` to `_generate` which is invoked concurrently after `on_page_markdown` 😵💫
I'm not sure why the change was made, the comment claims Pillow to be the minimal requirement, but there are no safety checks done for cairosvg, so it's assumed it's always installed anyway.

I would move out the safety logic back to `on_config` like the community version of the plugin, as I don't see advantages to throwing an error only after the concurrency jobs are invoked 🤔

squidfunk:
Yes, that's why I said I need a little more time to investigate. I think I put it there because of a better editing experience when crafting social card layouts. I'm quite sure I put it in `on_config` first and moved it out because of several oddities, but I can't remember entirely – sorry, too many changes happened since then and my comment is not clear enough (that's why I stress to write good comments explaining design decisions so often).

We need to test it thoroughly before making any change to it, i.e. check how the social plugin behaves in all of those cases (building, serving, cairo error, pillow not available). I would like to work on this myself. Again, please give me some time.

vedranmiletic:
Thanks for this, social plugin now works again inside pipenv venv on FreeBSD.

PR #6823, opened 2024-02-24
docs: update integration of KaTeX instructions

sinaatalay:
I am using `KaTeX` in my `mkdocs-material` documentation. I had a problem: `KaTeX` never loads the first time; you need to refresh the page to make it load. 

I found the solution, rather than using this (which is given by the official documentation of material-mkdocs):
```javascript
document$.subscribe(({ body }) => { 
  renderMathInElement(body, {
    delimiters: [
      { left: "$$",  right: "$$",  display: true },
      { left: "$",   right: "$",   display: false },
      { left: "\\(", right: "\\)", display: false },
      { left: "\\[", right: "\\]", display: true }
    ],
  })
})
```

I used this:

```javascript
document$.subscribe(() => { 
  renderMathInElement(document.body, {
    delimiters: [
      { left: "$$",  right: "$$",  display: true },
      { left: "$",   right: "$",   display: false },
      { left: "\\(", right: "\\)", display: false },
      { left: "\\[", right: "\\]", display: true }
    ],
  })
})
```

It solved the problem. I don't know JavaScipt, so I don't know how it solved it, but it worked.

This pull request updates the documentation's instructions on KaTeX integration and fixes this issue for possible KaTeX users.

### DIFF
--- a/docs/reference/math.md
+++ b/docs/reference/math.md
@@ -101,8 +101,8 @@ supports a subset of LaTeX syntax and can render math to HTML and SVG. To use
 === ":octicons-file-code-16: `docs/javascripts/katex.js`"
 
     ``` js
-    document$.subscribe(({ body }) => { // (1)!
-      renderMathInElement(body, {
+    document$.subscribe(() => { // (1)!
+      renderMathInElement(document.body, {
         delimiters: [
           { left: "$$",  right: "$$",  display: true },
           { left: "$",   right: "$",   display: false },


squidfunk:
Thanks for the PR! I'm not a user of KaTeX, so I need to rely on the community to tell whether this change makes sense. @adhadse did the [original KaTeX integration](https://github.com/squidfunk/mkdocs-material/commit/c1948b22ea3d46716b1e384268235e0d7390ac5a) – can you confirm that this is a better solution?

squidfunk:
It turned out to be a regression of [9.5.5](https://github.com/squidfunk/mkdocs-material/releases/tag/9.5.5). Fixed in de46c41. Thus, this PR can be closed.

squidfunk:
Released as part of [9.5.12](https://squidfunk.github.io/mkdocs-material/changelog/#9.5.12).

PR #6826, opened 2024-02-24
Better file gathering and validation for the info plugin

### LINKED ISSUE #6750: The `info` plugin does not includes inherited configurations
### Context

_No response_

### Bug description

When you create a report with the `info` plugin having a `INHERIT` key in your Mkdocs configuration, the generated ZIP does not includes inherited configurations.

### Related links

- https://www.mkdocs.org/user-guide/configuration/#configuration-inheritance
- https://squidfunk.github.io/mkdocs-material/plugins/info/

### Reproduction

Create a minimal example, generate your report building and check that the inherited configuration file is not added.

### Steps to reproduce

1. `mkdocs new .`
2. Change `mkdocs.yml` content by:
```yaml
INHERIT: 'parent.yml'
theme: material
plugins:
    - info
```
3. Create `parent.yml` file with the content:
```yaml
site_name: Foo
```
4. `mkdocs build`
5. Give a name and extract the result.
6. The *parent.yml* file is missing.

### Browser

_No response_

### Before submitting

- [X] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
- [X] I have attached links to [the documentation](https://squidfunk.github.io/mkdocs-material/), and possibly related [issues](https://github.com/squidfunk/mkdocs-material/issues) and [discussions](https://github.com/squidfunk/mkdocs-material/discussions).
- [X] I assure that I have [removed all customizations](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/#remove-customizations) before submitting this bug report.
- [X] I have attached a __.zip file__ with a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/).

kamilkrzyskow:
Fixes #6750

Took a bit longer than I expected to make a minimal viable improvement, or perhaps I was running circles in my head for too long instead of actually coding via trial and error 😅 

I also played a bit too much with the validation for different things relative to the root directory. 
I'll simplify it, but I don't want to remove it completely. 

As for the `info.gitignore` I hope the capabilities aren't too limiting compared to real `.gitignore` logic.
It's likely I'm missing something when it comes to `fnmatch`.

Once I'm closer to finishing, I'll also squash some of the commits to keep only relevant final changes.

> It's worth commenting that Snippets or any markdown_extension use the default working directory,
so the base_path is always relative to `mkdocs build` command, not relative to the config file. By default it also
doesn't crash when a include snippet path isn't resolved, just quietly omits it.

Task list:

- [x] Handle INHERIT configs
- [x] Read every file, not only the ones that MkDocs sees
- [x] Exclude paths we don't want
- [x] Basic validation for paths to assure `mkdocs build` at the top root directory.
  ~~Currently it's not basic enough, rather convoluted.~~ 
- [x] Improve fnmatch handling
- [x] Test with the community multi-lang project https://github.com/squidfunk/mkdocs-material/discussions/2346
- [x] Test with a `projects` plugin docs, like the examples repository
- [x] Fix styling (rough 80 lines), naming consistency etc.
- [x] Squash commits into categorized ones
- [x] Switch to `regex` from `fnmatch`

### DIFF
--- a/src/plugins/info/plugin.py
+++ b/src/plugins/info/plugin.py
@@ -23,6 +23,7 @@ import logging
 import os
 import platform
 import requests
+import site
 import sys
 
 from colorama import Fore, Style
@@ -31,6 +32,7 @@ from io import BytesIO
 from markdown.extensions.toc import slugify
 from mkdocs.plugins import BasePlugin, event_priority
 from mkdocs.utils import get_theme_dir
+import regex
 from zipfile import ZipFile, ZIP_DEFLATED
 
 from .config import InfoConfig
@@ -48,6 +50,9 @@ class InfoPlugin(BasePlugin[InfoConfig]):
 
         # Initialize incremental builds
         self.is_serve = False
+
+        # Initialize empty members
+        self.exclusion_patterns = []
 
     # Determine whether we're serving the site
     def on_startup(self, *, command, dirty):
@@ -111,12 +116,42 @@ class InfoPlugin(BasePlugin[InfoConfig]):
         example, _ = os.path.splitext(example)
         example = "-".join([present, slugify(example, "-")])
 
+        # Load exclusion patterns
+        self.exclusion_patterns = _load_exclusion_patterns()
+
+        # Exclude the site_dir at project root
+        if config.site_dir.startswith(os.getcwd()):
+            self.exclusion_patterns.append(_resolve_pattern(config.site_dir))
+
+        # Exclude the site-packages directory
+        for path in site.getsitepackages():
+            if path.startswith(os.getcwd()):
+                self.exclusion_patterns.append(_resolve_pattern(path))
+
         # Create self-contained example from project
         files: list[str] = []
         with ZipFile(archive, "a", ZIP_DEFLATED, False) as f:
             for abs_root, dirnames, filenames in os.walk(os.getcwd()):
+                # Prune the folders in-place to prevent
+                # scanning excluded folders
+                for name in list(dirnames):
+                    path = os.path.join(abs_root, name)
+                    if self._is_excluded(_resolve_pattern(path)):
+                        dirnames.remove(name)
+                        continue
+                    # Multi-language setup from #2346 separates the
+                    # language config, so each mkdocs.yml file is
+                    # unaware of other site_dir directories. Therefore,
+                    # we add this with the assumption a site_dir contains
+                    # the sitemap file.
+                    sitemap_gz = os.path.join(path, "sitemap.xml.gz")
+                    if os.path.exists(sitemap_gz):
+                        log.debug(f"Excluded site_dir: {path}")
+                        dirnames.remove(name)
                 for name in filenames:
                     path = os.path.join(abs_root, name)
+                    if self._is_excluded(_resolve_pattern(path)):
+                        continue
                     path = os.path.relpath(path, os.path.curdir)
                     f.write(path, os.path.join(example, path))
 
@@ -225,6 +260,15 @@ class InfoPlugin(BasePlugin[InfoConfig]):
         if self.config.archive_stop_on_violation:
             sys.exit(1)
 
+    # Exclude files, which we don't want in our zip file
+    def _is_excluded(self, posix_path: str) -> bool:
+        for pattern in self.exclusion_patterns:
+            if regex.match(pattern, posix_path):
+                log.debug(f"Excluded pattern '{pattern}': {posix_path}")
+                return True
+
+        return False
+
 # -----------------------------------------------------------------------------
 # Helper functions
 # -----------------------------------------------------------------------------
@@ -238,6 +282,34 @@ def _size(value, factor = 1):
         if abs(value) < 1000.0:
             return f"{color}{value:3.1f} {unit}"
         value /= 1000.0
+
+# Load info.gitignore, ignore any empty lines or # comments
+def _load_exclusion_patterns(path: str = None):
+    if path is None:
+        path = os.path.dirname(os.path.abspath(__file__))
+        path = os.path.join(path, "info.gitignore")
+
+    with open(path, encoding = "utf-8") as file:
+        lines = map(str.strip, file.readlines())
+
+    return [line for line in lines if line and not line.startswith("#")]
+
+# For the pattern matching it is best to remove the CWD
+# prefix and keep only the relative root of the reproduction.
+# Additionally, as the patterns are in POSIX format,
+# assure that the path is also in POSIX format.
+# Side-effect: It appends "/" for directory patterns.
+def _resolve_pattern(abspath: str):
+    path = abspath.replace(os.getcwd(), "", 1).replace(os.sep, "/")
+
+    if not path:
+        return "/"
+
+    # Check abspath, as the file needs to exist
+    if not os.path.isfile(abspath):
+        return path.rstrip("/") + "/"
+
+    return path
 
 # -----------------------------------------------------------------------------
 # Data
--- a/src/plugins/info/plugin.py
+++ b/src/plugins/info/plugin.py
@@ -18,6 +18,7 @@
 # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
 # IN THE SOFTWARE.
 
+import glob
 import json
 import logging
 import os
@@ -30,6 +31,7 @@ from colorama import Fore, Style
 from importlib.metadata import distributions, version
 from io import BytesIO
 from markdown.extensions.toc import slugify
+from mkdocs.config.defaults import MkDocsConfig
 from mkdocs.plugins import BasePlugin, event_priority
 from mkdocs.utils import get_theme_dir
 import regex
@@ -107,6 +109,27 @@ class InfoPlugin(BasePlugin[InfoConfig]):
             log.error("Please remove 'hooks' setting.")
             self._help_on_customizations_and_exit()
 
+        # Assure that config_file_path is absolute.
+        # If the --config-file option is used then the path is
+        # used as provided, so it is likely relative.
+        if not os.path.isabs(config.config_file_path):
+            config.config_file_path = os.path.normpath(os.path.join(
+                os.getcwd(),
+                config.config_file_path
+            ))
+
+        # Support projects plugin
+        projects_plugin = config.plugins.get("material/projects")
+        if projects_plugin:
+            abs_projects_dir = os.path.normpath(
+                os.path.join(
+                    os.path.dirname(config.config_file_path),
+                    projects_plugin.config.projects_dir
+                )
+            )
+        else:
+            abs_projects_dir = ""
+
         # Create in-memory archive and prompt author for a short descriptive
         # name for the archive, which is also used as the directory name. Note
         # that the name is slugified for better readability and stripped of any
@@ -127,6 +150,18 @@ class InfoPlugin(BasePlugin[InfoConfig]):
         for path in site.getsitepackages():
             if path.startswith(os.getcwd()):
                 self.exclusion_patterns.append(_resolve_pattern(path))
+
+        # Exclude site_dir for projects
+        if projects_plugin:
+            for path in glob.iglob(
+                pathname = projects_plugin.config.projects_config_files,
+                root_dir = abs_projects_dir,
+                recursive = True
+            ):
+                current_config_file = os.path.join(abs_projects_dir, path)
+                project_config = _get_project_config(current_config_file)
+                pattern = _resolve_pattern(project_config.site_dir)
+                self.exclusion_patterns.append(pattern)
 
         # Create self-contained example from project
         files: list[str] = []
@@ -310,6 +345,17 @@ def _resolve_pattern(abspath: str):
         return path.rstrip("/") + "/"
 
     return path
+
+# Get project configuration
+def _get_project_config(project_config_file: str):
+    with open(project_config_file, encoding="utf-8") as file:
+        config = MkDocsConfig(config_file_path = project_config_file)
+        config.load_file(file)
+
+        # MkDocs transforms site_dir to absolute path during validation
+        config.validate()
+
+        return config
 
 # -----------------------------------------------------------------------------
 # Data
--- a/src/plugins/info/plugin.py
+++ b/src/plugins/info/plugin.py
@@ -135,7 +135,13 @@ class InfoPlugin(BasePlugin[InfoConfig]):
                 json.dumps(
                     {
                         "system": platform.platform(),
-                        "python": platform.python_version()
+                        "architecture": platform.architecture(),
+                        "python": platform.python_version(),
+                        "command": " ".join([
+                            sys.argv[0].rsplit(os.sep, 1)[-1],
+                            *sys.argv[1:]
+                        ]),
+                        "sys.path": sys.path
                     },
                     default = str,
                     indent = 2
--- a/src/plugins/info/plugin.py
+++ b/src/plugins/info/plugin.py
@@ -23,11 +23,12 @@ import json
 import logging
 import os
 import platform
+import regex
 import requests
 import site
 import sys
-
 import yaml
+
 from colorama import Fore, Style
 from importlib.metadata import distributions, version
 from io import BytesIO
@@ -35,7 +36,6 @@ from markdown.extensions.toc import slugify
 from mkdocs.config.defaults import MkDocsConfig
 from mkdocs.plugins import BasePlugin, event_priority
 from mkdocs.utils import get_yaml_loader
-import regex
 from zipfile import ZipFile, ZIP_DEFLATED
 
 from .config import InfoConfig
@@ -110,21 +110,24 @@ class InfoPlugin(BasePlugin[InfoConfig]):
             log.error("Please remove 'hooks' setting.")
             self._help_on_customizations_and_exit()
 
-        # Assure that possible relative paths, which will be validated
-        # or used to generate other paths are absolute.
+        # Assure all paths that will be validated are absolute. Convert possible
+        # relative config_file_path to absolute. Its absolute directory path is
+        # being later used to resolve other paths.
         config.config_file_path = _convert_to_abs(config.config_file_path)
         config_file_parent = os.path.dirname(config.config_file_path)
 
-        # The theme.custom_dir property cannot be set, therefore a helper
-        # variable is used.
-        custom_dir = config.theme.custom_dir
-        if custom_dir:
-            custom_dir = _convert_to_abs(
-                custom_dir,
+        # Convert relative custom_dir path to absolute. The Theme.custom_dir
+        # property cannot be set, therefore a helper variable is used.
+        if config.theme.custom_dir:
+            abs_custom_dir = _convert_to_abs(
+                config.theme.custom_dir,
                 abs_prefix = config_file_parent
             )
+        else:
+            abs_custom_dir = ""
 
-        # Support projects plugin
+        # Extract the absolute path to projects plugin's directory to explicitly
+        # support path validation and dynamic exclusion for the plugin
         projects_plugin = config.plugins.get("material/projects")
         if projects_plugin:
             abs_projects_dir = _convert_to_abs(
@@ -134,29 +137,39 @@ class InfoPlugin(BasePlugin[InfoConfig]):
         else:
             abs_projects_dir = ""
 
-        # Load the current MkDocs config(s) to get access to INHERIT
+        # MkDocs removes the INHERIT configuration key during load, and doesn't
+        # expose the information in any way, as the parent configuration is
+        # merged into one. To validate that the INHERIT config file will be
+        # included in the ZIP file the current config file must be loaded again
+        # without parsing. Each file can have their own INHERIT key, so a list
+        # of configurations is supported. The INHERIT path is converted during
+        # load to absolute.
         loaded_configs = _load_yaml(config.config_file_path)
         if not isinstance(loaded_configs, list):
             loaded_configs = [loaded_configs]
 
-        # Validate different MkDocs paths to assure that
-        # they're children of the current working directory.
+        # We need to make sure the user put every file in the current working
+        # directory. To assure the reproduction inside the ZIP file can be run,
+        # validate that the MkDocs paths are children of the current root.
         paths_to_validate = [
             config.config_file_path,
             config.docs_dir,
-            custom_dir or "",
+            abs_custom_dir,
             abs_projects_dir,
             *[cfg.get("INHERIT", "") for cfg in loaded_configs]
         ]
 
+        # Convert relative hook paths to absolute path
         for hook in config.hooks:
             path = _convert_to_abs(hook, abs_prefix = config_file_parent)
             paths_to_validate.append(path)
 
+        # Remove valid paths from the list
         for path in list(paths_to_validate):
             if not path or path.startswith(os.getcwd()):
                 paths_to_validate.remove(path)
 
+        # Report the invalid paths to the user
         if paths_to_validate:
             log.error(f"One or more paths aren't children of root")
             self._help_on_not_in_cwd(paths_to_validate)
@@ -198,26 +211,36 @@ class InfoPlugin(BasePlugin[InfoConfig]):
         files: list[str] = []
         with ZipFile(archive, "a", ZIP_DEFLATED, False) as f:
             for abs_root, dirnames, filenames in os.walk(os.getcwd()):
-                # Prune the folders in-place to prevent
-                # scanning excluded folders
+                # Prune the folders in-place to prevent their processing
                 for name in list(dirnames):
+                    # Resolve the absolute directory path
                     path = os.path.join(abs_root, name)
+
+                    # Exclude the directory and all subdirectories
                     if self._is_excluded(_resolve_pattern(path)):
                         dirnames.remove(name)
                         continue
-                    # Multi-language setup from #2346 separates the
-                    # language config, so each mkdocs.yml file is
-                    # unaware of other site_dir directories. Therefore,
-                    # we add this with the assumption a site_dir contains
-                    # the sitemap file.
+
+                    # Projects, which don't use the projects plugin for
+                    # multi-language support could have separate build folders
+                    # for each config file or language. Therefore, we exclude
+                    # them with the assumption a site_dir contains the sitemap
+                    # file. Example of such a setup: https://t.ly/DLQcy
                     sitemap_gz = os.path.join(path, "sitemap.xml.gz")
                     if os.path.exists(sitemap_gz):
                         log.debug(f"Excluded site_dir: {path}")
                         dirnames.remove(name)
+
+                # Write files to the in-memory archive
                 for name in filenames:
+                    # Resolve the absolute file path
                     path = os.path.join(abs_root, name)
+
+                    # Exclude the file
                     if self._is_excluded(_resolve_pattern(path)):
                         continue
+
+                    # Resolve the relative path to create a matching structure
                     path = os.path.relpath(path, os.path.curdir)
                     f.write(path, os.path.join(example, path))
 
@@ -320,6 +343,11 @@ class InfoPlugin(BasePlugin[InfoConfig]):
         print(Style.NORMAL)
         print("  - extra_css")
         print("  - extra_javascript")
+        print(Fore.YELLOW)
+        print("  If you're using customizations from the theme's documentation")
+        print("  and you want to report a bug specific to those customizations")
+        print("  then set the 'archive_stop_on_violation: false' option in the")
+        print("  info plugin config.")
         print(Style.RESET_ALL)
 
         # Exit, unless explicitly told not to
@@ -327,20 +355,19 @@ class InfoPlugin(BasePlugin[InfoConfig]):
             sys.exit(1)
 
     # Print help on not in current working directory and exit
-    def _help_on_not_in_cwd(self, bad_paths):
+    def _help_on_not_in_cwd(self, outside_root):
         print(Fore.RED)
         print("  The current working (root) directory:\n")
         print(f"    {os.getcwd()}\n")
         print("  is not a parent of the following paths:")
         print(Style.NORMAL)
-        for path in bad_paths:
+        for path in outside_root:
             print(f"    {path}")
-        print()
-        print("  To assure that all project files are found")
-        print("  please adjust your config or file structure and")
-        print("  put everything within the root directory of the project.\n")
-        print("  Please also make sure `mkdocs build` is run in")
-        print("  the actual root directory of the project.")
+        print("  \nTo assure that all project files are found please adjust")
+        print("  your config or file structure and put everything within the")
+        print("  root directory of the project.\n")
+        print("  Please also make sure `mkdocs build` is run in the actual")
+        print("  root directory of the project.")
         print(Style.RESET_ALL)
 
         # Exit, unless explicitly told not to
@@ -370,17 +397,19 @@ def _size(value, factor = 1):
             return f"{color}{value:3.1f} {unit}"
         value /= 1000.0
 
-# To validate if a file is within the file tree,
-# it needs to be absolute, so that it is possible to
+# Get the absolute path with set prefix. To validate if a file is inside the
+# current working directory it needs to be absolute, so that it is possible to
 # check the prefix.
 def _convert_to_abs(path: str, abs_prefix: str = None) -> str:
     if os.path.isabs(path): return path
     if abs_prefix is None: abs_prefix = os.getcwd()
     return os.path.normpath(os.path.join(abs_prefix, path))
 
-# Custom YAML loader - required to handle the parent INHERIT config.
-# It converts the INHERIT path to absolute as a side effect.
-# Returns the loaded config, or a list of all loaded configs.
+# Get the loaded config, or a list with all loaded configs. MkDocs removes the
+# INHERIT configuration key during load, and doesn't expose the information in
+# any way, as the parent configuration is merged into one. The INHERIT path is
+# needed for validation. This custom YAML loader replicates MkDocs' loading
+# logic. Side effect: It converts the INHERIT path to absolute.
 def _load_yaml(abs_src_path: str):
 
     with open(abs_src_path, "r", encoding ="utf-8-sig") as file:
@@ -417,11 +446,9 @@ def _load_exclusion_patterns(path: str = None):
 
     return [line for line in lines if line and not line.startswith("#")]
 
-# For the pattern matching it is best to remove the CWD
-# prefix and keep only the relative root of the reproduction.
-# Additionally, as the patterns are in POSIX format,
-# assure that the path is also in POSIX format.
-# Side-effect: It appends "/" for directory patterns.
+# Get a normalized POSIX path for the pattern matching with removed current
+# working directory prefix. Directory paths end with a '/' to allow more control
+# in the pattern creation for files and directories.
 def _resolve_pattern(abspath: str):
     path = abspath.replace(os.getcwd(), "", 1).replace(os.sep, "/")
 
@@ -434,7 +461,7 @@ def _resolve_pattern(abspath: str):
 
     return path
 
-# Get project configuration
+# Get project configuration with resolved absolute paths for validation
 def _get_project_config(project_config_file: str):
     with open(project_config_file, encoding="utf-8") as file:
         config = MkDocsConfig(config_file_path = project_config_file)
--- a/src/plugins/info/plugin.py
+++ b/src/plugins/info/plugin.py
@@ -30,7 +30,6 @@ from importlib.metadata import distributions, version
 from io import BytesIO
 from markdown.extensions.toc import slugify
 from mkdocs.plugins import BasePlugin, event_priority
-from mkdocs.structure.files import get_files
 from mkdocs.utils import get_theme_dir
 from zipfile import ZipFile, ZIP_DEFLATED
 
@@ -115,14 +114,11 @@ class InfoPlugin(BasePlugin[InfoConfig]):
         # Create self-contained example from project
         files: list[str] = []
         with ZipFile(archive, "a", ZIP_DEFLATED, False) as f:
-            for path in ["mkdocs.yml", "requirements.txt"]:
-                if os.path.isfile(path):
+            for abs_root, dirnames, filenames in os.walk(os.getcwd()):
+                for name in filenames:
+                    path = os.path.join(abs_root, name)
+                    path = os.path.relpath(path, os.path.curdir)
                     f.write(path, os.path.join(example, path))
-
-            # Append all files visible to MkDocs
-            for file in get_files(config):
-                path = os.path.relpath(file.abs_src_path, os.path.curdir)
-                f.write(path, os.path.join(example, path))
 
             # Add information on installed packages
             f.writestr(
--- a/src/plugins/info/plugin.py
+++ b/src/plugins/info/plugin.py
@@ -27,13 +27,14 @@ import requests
 import site
 import sys
 
+import yaml
 from colorama import Fore, Style
 from importlib.metadata import distributions, version
 from io import BytesIO
 from markdown.extensions.toc import slugify
 from mkdocs.config.defaults import MkDocsConfig
 from mkdocs.plugins import BasePlugin, event_priority
-from mkdocs.utils import get_theme_dir
+from mkdocs.utils import get_yaml_loader
 import regex
 from zipfile import ZipFile, ZIP_DEFLATED
 
@@ -97,7 +98,7 @@ class InfoPlugin(BasePlugin[InfoConfig]):
         # hack to detect whether the custom_dir setting was used without parsing
         # mkdocs.yml again - we check at which position the directory provided
         # by the theme resides, and if it's not the first one, abort.
-        if config.theme.dirs.index(get_theme_dir(config.theme.name)):
+        if config.theme.custom_dir:
             log.error("Please remove 'custom_dir' setting.")
             self._help_on_customizations_and_exit()
 
@@ -109,26 +110,56 @@ class InfoPlugin(BasePlugin[InfoConfig]):
             log.error("Please remove 'hooks' setting.")
             self._help_on_customizations_and_exit()
 
-        # Assure that config_file_path is absolute.
-        # If the --config-file option is used then the path is
-        # used as provided, so it is likely relative.
-        if not os.path.isabs(config.config_file_path):
-            config.config_file_path = os.path.normpath(os.path.join(
-                os.getcwd(),
-                config.config_file_path
-            ))
+        # Assure that possible relative paths, which will be validated
+        # or used to generate other paths are absolute.
+        config.config_file_path = _convert_to_abs(config.config_file_path)
+        config_file_parent = os.path.dirname(config.config_file_path)
+
+        # The theme.custom_dir property cannot be set, therefore a helper
+        # variable is used.
+        custom_dir = config.theme.custom_dir
+        if custom_dir:
+            custom_dir = _convert_to_abs(
+                custom_dir,
+                abs_prefix = config_file_parent
+            )
 
         # Support projects plugin
         projects_plugin = config.plugins.get("material/projects")
         if projects_plugin:
-            abs_projects_dir = os.path.normpath(
-                os.path.join(
-                    os.path.dirname(config.config_file_path),
-                    projects_plugin.config.projects_dir
-                )
+            abs_projects_dir = _convert_to_abs(
+                projects_plugin.config.projects_dir,
+                abs_prefix = config_file_parent
             )
         else:
             abs_projects_dir = ""
+
+        # Load the current MkDocs config(s) to get access to INHERIT
+        loaded_configs = _load_yaml(config.config_file_path)
+        if not isinstance(loaded_configs, list):
+            loaded_configs = [loaded_configs]
+
+        # Validate different MkDocs paths to assure that
+        # they're children of the current working directory.
+        paths_to_validate = [
+            config.config_file_path,
+            config.docs_dir,
+            custom_dir or "",
+            abs_projects_dir,
+            *[cfg.get("INHERIT", "") for cfg in loaded_configs]
+        ]
+
+        for hook in config.hooks:
+            path = _convert_to_abs(hook, abs_prefix = config_file_parent)
+            paths_to_validate.append(path)
+
+        for path in list(paths_to_validate):
+            if not path or path.startswith(os.getcwd()):
+                paths_to_validate.remove(path)
+
+        if paths_to_validate:
+            log.error(f"One or more paths aren't children of root")
+            self._help_on_not_in_cwd(paths_to_validate)
 
         # Create in-memory archive and prompt author for a short descriptive
         # name for the archive, which is also used as the directory name. Note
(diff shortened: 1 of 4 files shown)


squidfunk (changes requested):
Thanks for the PR! I see that you went down the 'more specific' road. As I argued in https://github.com/squidfunk/mkdocs-material/issues/6750#issuecomment-1962275522, I don't think this is what we should do, because we'll be adding exceptions, exceptions and more exceptions. For instance, mkdocstrings usage is not covered in this solution, as far as I can tell. Additionally, the site package exclusion is based on git, which we should do programmatically, because we'll be adding more and more entries to the `.gitignore` as we discover other preferences for naming folders for virtual environments.

Moreover, configuration inheritance will likely not catch all cases, as we discussed in #6750. Collections of paths is based on a curated list of names of parameters, which is too labor intense, given that packing up of bug reports should 'just work'. Thus, as outlined in my last comment, I regard it as a better approach to go down the following road:

- Collect all files recursively from current directory
- Filter files that are in our `.gitignore` or in `site-packages` (this is essential)
- Iterate and refine our `.gitignore` if we find other files that we _always_ want to exclude

This should also allow for the projects plugin and other multi-configuration plugins to be accurately collected + is a solution that does not require maintenance from our side as the ecosystem grows.

kamilkrzyskow:
> I see that you went down the 'more specific' road.

Yes, I did, but I want to limit the specificity to MkDocs itself, as I mentioned:
> I also played a bit too much with the validation for different things relative to the root directory.
I'll simplify it, but I don't want to remove it completely.

which wasn't clear enough, sorry. I want to keep validation for:
- `docs_dir`, 
- `INHERIT` 
- `docs_dir` in the inherited config(s)
- `INHERIT` in the inherited configs(s)

I believe this is a good bare minimum to detect running `mkdocs build` from the wrong directory, it might be enough to check `docs_dir`, but `INHERIT` can influence `docs_dir` therefore I opted into all 4. 

In the code currently there is also the validation for markdown_extensions, which I checked mostly use 3 generic names for path options, but it will be removed soon, as by default, running `mkdocs build --config-file=configs/mkdocs.yml` will break Snippets if they were setup to work with `mkdocs build --config-file=mkdocs.yml`, so if Snippets don't support it, then we also don't have to.

I agree that `mkdocsstrings` shouldn't be validated via the `info` plugin as it's not a part of the python dependencies of `mkdocs-material`, Python Markdown Extensions are, so initially I wanted to support Snippets as well but it turned out unnecessary.

If you still insist that this simple validation is out of scope I'll remove it completely :v:

---

> Additionally, the site package exclusion is based on git

Not completely accurate. In the `info.gitignore` I added the mostly used generic venv names, as per the [Python.gitignore](https://github.com/github/gitignore/blob/main/Python.gitignore), however it's not the only place where it's checked:
https://github.com/squidfunk/mkdocs-material/blob/1e7c2f2b4ed1f40ee4b711057b79bb9a583a9640/src/plugins/info/plugin.py#L364-L368

> Note to self: I should probably add a comment about that in the `info.gitignore` 🤔 

I think the user might have different virtual environments, perhaps one with `mkdocs-material` and another with `mkdocs-material-insiders`, and they would often switch between the 2, so the `site.getsitepackages()` would not detect an active venv, but it would be detected by `info.gitignore`. I can see that this might be "overprotective" for the user, but I don't see downsides with this approach, unless the user decides to add a `docs/venv` directory with actual contents for reproduction...

> ... which we should do programmatically, 

Yes, I understand that there might be more custom generated paths in the project directory. I already add a `site_dir` exclusion pattern. 

https://github.com/squidfunk/mkdocs-material/blob/1e7c2f2b4ed1f40ee4b711057b79bb9a583a9640/src/plugins/info/plugin.py#L174

> Note to self: Now that I think about it, I forgot about the `.cache` directory, also since we don't support hooks `*.py` would also be reasonable 🤔 

I'm a bit busy today, but I will still try to clean up the PR before "tomorrow" comes 😅

squidfunk:
Checking for `docs_dir` is a good idea. However, I think supporting `INHERIT` _would_ require potentially going up a very long chain of inheritance, or we'll loose information. Letting `mkdocs` resolve `INHERIT` and only packing up the result will make it impossible for us to troubleshoot whether the problem actually stems from configuration inheritance. It's safer to just ignore it and just include all files that look like `mkdocs.yml` configurations. We could also include the command that was used for running `mkdocs build` when packing up the example, in order to disambiguate in case of multiple configurations. We're trying to make it less error prone than it currently is, and from my experience writing and maintaining the plugin, it means that we should do _less_ validation and pack up more selectively.

So, to sum up: the aim should be: pack up all files that are used to run `mkdocs build` _without losing any information_. If we pack up resolved stuff, we'll loose information, which might involve more time for debugging and troubleshooting.

> I agree that mkdocsstrings shouldn't be validated via the info plugin as it's not a part of the python dependencies of mkdocs-material, Python Markdown Extensions are, so initially I wanted to support Snippets as well but it turned out unnecessary.

Yet, packing up an example with `mkdocstrings` should work. We should not limit the info plugin to the plugins we officially support, because users might report bugs with plugins that we _might_ want to support by learning about them through bug reports. Thus, less validation, and a more forgiving packing up procedure is key, IMHO.

> I think the user might have different virtual environments, perhaps one with mkdocs-material and another with mkdocs-material-insiders, and they would often switch between the 2, so the site.getsitepackages() would not detect an inactive venv, but it would be detected by info.gitignore. I can see that this might be "overprotective" for the user, but I don't see downsides with this approach, unless the user decides to add a docs/venv directory with actual contents for reproduction...

Yet, only one environment is active. I doubt that users use two virtual environments in one project. It sounds like an anti-pattern to me, and IMHO, we should optimize for the common case where only a single environment is present, i.e., the active environment, which we should resolve as discussed.

> Yes, I understand that there might be more custom generated paths in the project directory. I already add a site_dir exclusion pattern.

Sounds good! We likely need to repeat this for all projects, i.e, all `mkdocs.yml` files we find. For completeness:  we cannot exclude HTML files, because they can be customizations (which you can also force the plugin to pack with another flag), or static templates in `docs_dir`.

> I'm a bit busy today, but I will still try to clean up the PR before "tomorrow" comes 😅

No hurry! I hope my reasoning makes sense. If not, don't hesitate to ask. I hope I could make it clearer why we need to make the plugin more resilient and less strict, yet keep it simple and powerful. Too much magic is like not a good idea for something that should work with minimal effort from a usage and maintenance perspective.

> EDIT: After some more pondering I think that docs/.../site/ or docs/.../env/ isn't such an unreasonable setup, so the current site/ or env/ exclusion would perhaps impact the zip creation negatively 😞 I will have to make sure that we can check for /site in the exclusion patterns for root/site, which is not possible currently.

The `venv` case should be caught by excluding site-packages, but the site directory inside the docs directory sounds like an anti-pattern to me. Is that even supported by MkDocs? Additionally, resolving the entire paths before removing files from the packing procedure should resolve both cases, shouldn't it? I mean: pack up `docs`, but don't pack up `docs/.../site`.

kamilkrzyskow:
I feel like there is some misunderstanding, because
> Letting `mkdocs` resolve INHERIT and only packing up the result will make it impossible for us to troubleshoot whether the problem actually stems from configuration inheritance. It's safer to just ignore it and just include all files that look like mkdocs.yml configurations.

indicates to me that you believe I'm using resolve INHERIT to get the parent.yml file and then include it in the zip file. If I'm wrong then sorry. I resolve INHERIT because MkDocs doesn't include it in MkDocsConfig, and I want to validate that it is inside the `mkdocs build` directory. If it's not I report it to the user, similar to the `version` or `customization` check.

Apart of that I don't use the INHERIT config, and use everything that is provided by the already validated and resolved MkDocsConfig and everything inside the current working directory of `mkdocs build`. So this should meet the criteria for:
> So, to sum up: the aim should be: pack up all files that are used to run mkdocs build without losing any information

I just want to avoid missing any files, so if there is any invalid path outside of CWD the user gets notified and the program ends.

> We could also include the command that was used for running mkdocs build when packing up the example, in order to disambiguate in case of multiple configurations.

Good idea. I will also consider adding some kind of output/log copy in the zip. But perhaps that's a feature for a future PR.

> we should optimize for the common case where only a single environment is present, i.e., the active environment, which we should resolve as discussed.

OK, I removed the static names, I also modified the exclusion logic, so that exclusion patterns for venvs are generated.

> We likely need to repeat this for all projects, i.e, all mkdocs.yml files we find. For completeness: we cannot exclude HTML files, because they can be customizations (which you can also force the plugin to pack with another flag), or static templates in docs_dir.

I considered excluding site_dir based on the inclusion of `sitemap.xml` because that's not a file that one can easily override.
However, I stuck to the exclusion pattern generation, so I needed to resolve the `mkdocs.yml` files for each project. This way I resovle each custom name a user could have set.
> Note to self: I forgot to add custom_dir handling for the customizations😰 

> No hurry! 

Thanks, but not only did I completely miss both deadlines I set in the PR, I suggested I would fix the plugin some months ago 🥶 So I'm trying to focus on this PR for now.

> The venv case should be caught by excluding site-packages, but the site directory inside the docs directory sounds like an anti-pattern to me. Is that even supported by MkDocs? Additionally, resolving the entire paths before removing files from the packing procedure should resolve both cases, shouldn't it? I mean: pack up docs, but don't pack up docs/.../site.

I meant paths that are called `docs/.../site` and `docs/.../venv` that have actual documentation `index.md` contents. Before the paths would be falsely excluded, now they won't.

---

Here is a slightly modified `examples` repository with the projects plugin:
[7-projects-examples.zip](https://github.com/squidfunk/mkdocs-material/files/14409372/7-projects-examples.zip)
if you like you can run it with the PR branch and `mkdocs build -v` to see the debug info about the excluded paths:
```
DEBUG   -  Excluded pattern '*.cache/': /.cache/ <-- static inside of our .gitignore
DEBUG   -  Excluded pattern '/site/': /site/ <-- dynamic added in code 
DEBUG   -  Excluded pattern '/venv/': /venv/ <-- dynamic added in code 
DEBUG   -  Excluded pattern '*.zip': /9.5.11+insiders.4.53.0-q.zip <-- static inside of our .gitignore
DEBUG   -  Excluded pattern '/examples/fonts_site_dir/': /examples/fonts_site_dir/ <-- dynamic added in code 
DEBUG   -  Excluded pattern '*__pycache__/': /examples/__pycache__/ <-- static inside of our .gitignore
DEBUG   -  Excluded pattern '/examples/blog-basic/site/': /examples/blog-basic/site/ <-- dynamic added in code 
DEBUG   -  Excluded pattern '*__pycache__/': /hooks/__pycache__/ <-- static inside of our .gitignore
DEBUG   -  Excluded pattern '*[!__init__].py': /hooks/archive.py <-- static inside of our .gitignore
```
The gathered project zip (inside of the zip above) isn't runnable because the `archive.py` hook is excluded and the config file still includes it.

I think it works fine, perhaps the exclusion is too lax. 
I still need to test the community multi-language setup
Add the command info, fix styling.

squidfunk:
> indicates to me that you believe I'm using resolve INHERIT to get the parent.yml file and then include it in the zip file. If I'm wrong then sorry. I resolve INHERIT because MkDocs doesn't include it in MkDocsConfig, and I want to validate that it is inside the mkdocs build directory. If it's not I report it to the user, similar to the version or customization check.
>
> Apart of that I don't use the INHERIT config, and use everything that is provided by the already validated and resolved MkDocsConfig and everything inside the current working directory of mkdocs build. So this should meet the criteria for:

I think I misread the implementation then, sorry! In that case, yes, we can try including the `INHERIT` validation and test it in the wild ☺️ 

> Good idea. I will also consider adding some kind of output/log copy in the zip. But perhaps that's a feature for a future PR.

Also a great idea, albeit it might be a catch-22?

> Thanks, but not only did I completely miss both deadlines I set in the PR, I suggested I would fix the plugin some months ago 🥶 So I'm trying to focus on this PR for now.

Again, no hurry. Albeit we're trying to fix bugs in the plugin, they happen only very rarely with most bug reports being complete and runnable (if users stick to our bug reporting guidelines and include the reproduction in the first place).

> I meant paths that are called docs/.../site and docs/.../venv that have actual documentation index.md contents. Before the paths would be falsely excluded, now they won't.

Ah yes, because of the ignore patterns. Yes! I read that this is another reason to be very careful when using exclusion patterns, so doing exclusion as programmatically as possible sounds like a safer approach.

> The gathered project zip (inside of the zip above) isn't runnable because the archive.py hook is excluded and the config file still includes it.

Hmm, so the hook should be included I think, albeit the plugin should warn you about using hooks. We also need to make sure that mkdocstrings works as expected. Here's an example that was manually packed because the info plugin did not include all files. We need to make sure that the packed archive has exactly the same context as this:
[9.5.9+insiders.4.52.2-link-style.zip](https://github.com/squidfunk/mkdocs-material/files/14413859/9.5.9%2Binsiders.4.52.2-link-style.zip)

> I think it works fine, perhaps the exclusion is too lax.
> I still need to test the community multi-language setup
> And ad the command info, fix styling.

We'll learn along the way ☺️

kamilkrzyskow:
> I think I misread the implementation then, sorry! In that case, yes, we can try including the INHERIT validation and test it in the wild ☺️

It's good we're finally on the same page. The confusion probably came from the [first commit](https://github.com/squidfunk/mkdocs-material/pull/6826/commits/4e21e1ef459a51b238caad4e50f99b1d1e0bf71a) I left for the time being and it will be squashed later..., it included the INHERIT path inclusion statically, but in a [further commit](https://github.com/squidfunk/mkdocs-material/pull/6826/commits/cabc3c6df5c79be4565444e3d1e811246800027a) I changed to the recursive path lookup. It's definitely not the cleanest PR (as usual), but I wanted to show that I started working on it, instead of make "assurances" in the Issue comments. Your comments definitely keep me within bounds 😄 

> Hmm, so the hook should be included I think, albeit the plugin should warn you about using hooks.

I forgot to mention that it did warn me about using hooks, so I enabled the config option that allows to create the archive despite bad validation. But when the user sees the warning they can just remove the `hooks` section from the config file, so I can't use `config.hooks` to dynamically ignore them from inclusion. 

> We also need to make sure that mkdocstrings works as expected. Here's an example that was manually packed because the info plugin did not include all files. We need to make sure that the packed archive has exactly the same contents as this:

From the example zip I see it uses a `__init__.py` script just like the file in the `examples` repository. The current exclusion pattern would include it, but I'll need to think about it some more, because there is surely a case where a `random_name.py` is used inside of the project that isn't a hook.

[9.5.11+insiders.4.53.0-strings.zip](https://github.com/squidfunk/mkdocs-material/files/14414288/9.5.11%2Binsiders.4.53.0-strings.zip)
> The contents are the same, but it contains 2 requirements.lock.txt and 2 platform.json 🤯 I didn't know a zip file can do that.

<details>
<summary> LOG </summary>

```bash
INFO    -  Started archive creation for bug report

Please name your bug report (2-4 words): strings
DEBUG   -  Excluded pattern '*.zip': /9.5.11+insiders.4.53.0-strings.zip
INFO    -  UserWarning: Duplicate name: '9.5.11+insiders.4.53.0-strings/requirements.lock.txt'
  File "C:\MyFiles\Python312\Lib\zipfile\__init__.py", line 1684, in _open_to_write
    self._writecheck(zinfo)
  File "C:\MyFiles\Python312\Lib\zipfile\__init__.py", line 1786, in _writecheck
    warnings.warn('Duplicate name: %r' % zinfo.filename, stacklevel=3)

INFO    -  UserWarning: Duplicate name: '9.5.11+insiders.4.53.0-strings/platform.json'
  File "C:\MyFiles\Python312\Lib\zipfile\__init__.py", line 1684, in _open_to_write
    self._writecheck(zinfo)
  File "C:\MyFiles\Python312\Lib\zipfile\__init__.py", line 1786, in _writecheck
    warnings.warn('Duplicate name: %r' % zinfo.filename, stacklevel=3)

INFO    -  Archive successfully created:

  9.5.11+insiders.4.53.0-strings/docs/index.md 404.0 B
  9.5.11+insiders.4.53.0-strings/mkdocs.yml 342.0 B
  9.5.11+insiders.4.53.0-strings/platform.json 282.0 B
  9.5.11+insiders.4.53.0-strings/platform.json 79.0 B
  9.5.11+insiders.4.53.0-strings/requirements.lock.txt 350.0 B
  9.5.11+insiders.4.53.0-strings/requirements.lock.txt 395.0 B
  9.5.11+insiders.4.53.0-strings/src/package/__init__.py 288.0 B

  9.5.11+insiders.4.53.0-strings.zip 3.4 kB
```

</details>

Perhaps the best approach would be to allow the user to set `config.hooks`, and later use the paths as exclusion patterns. This would not include hooks, but would include all other python scripts.
The project wouldn't be runnable because the config file expects the existence of the hooks, but as we have to remove the `info` plugin from the config inside the zip too, then perhaps this wouldn't be this big of an issue. 
Some warning should probably stay to make the user aware that we won't troubleshoot their custom hooks in a bug report. 🤔 

Another approach is a bit risky exclusion pattern again `*/hooks/*.py` where I expect the user to save the hooks indie of a directory called `hooks`, but I've seen cases where hooks are next to `mkdocs.yml`

---

Actually all of the above applies to `theme.custom_dir` as well. Because by default it prevents the plugin from running, but the user might just comment out `custom_dir` from their config, and leave the files next to the config... So the files will be included.

I guess the previous warnings were created because there was no exclusion logic to prevent files from being added to the zip. Since there is such a thing now, then perhaps the warnings shouldn't be as strict.

squidfunk:
> I forgot to mention that it did warn me about using hooks, so I enabled the config option that allows to create the archive despite bad validation. But when the user sees the warning they can just remove the hooks section from the config file, so I can't use config.hooks to dynamically ignore them from inclusion.

> Perhaps the best approach would be to allow the user to set config.hooks, and later use the paths as exclusion patterns. This would not include hooks, but would include all other python scripts.

> The project wouldn't be runnable because the config file expects the existence of the hooks, but as we have to remove the info plugin from the config inside the zip too, then perhaps this wouldn't be this big of an issue.

> Another approach is a bit risky exclusion pattern again */hooks/*.py where I expect the user to save the hooks indie of a directory called hooks, but I've seen cases where hooks are next to mkdocs.yml

> Actually all of the above applies to theme.custom_dir as well. Because by default it prevents the plugin from running, but the user might just comment out custom_dir from their config, and leave the files next to the config... So the files will be included.

The idea of the "ignore violation" flag is: warn the user about customizations and hooks and terminate unless the flag is set, so if the plugin progresses beyond the termination because of the flag, customizations and hooks should be packaged up as well. This might be necessary in very rare cases, e.g. for the customizations we advertise on our docs. Thus, customizations and hooks should never be ignored. The inline comments or our previous conversation might not have made this clear.

Additionally, there's no need to remove `info` from plugins. We shouldn't make `mkdocs.yml` transformations, as they might introduce new bugs which makes the plugin less resilient. This plugin needs to be maximally resilient.

> From the example zip I see it uses a __init__.py script just like the file in the examples repository. The current exclusion pattern would include it, but I'll need to think about it some more, because there is surely a case where a random_name.py is used inside of the project that isn't a hook.

We should include it, e.g., [as we advertise something similar for the projects plugin transform](https://squidfunk.github.io/mkdocs-material/plugins/projects/#config.projects_config_transform) (to make usage simpler).

> The contents are the same, but it contains 2 requirements.lock.txt and 2 platform.json 🤯 I didn't know a zip file can do that.

That is very, very weird. I'm not sure why this is the case!?

squidfunk:
As an addendum: I wouldn't handle the case where customizations are packaged up that are not used. This would mandate resolving things from `mkdocs.yml`. The current iteration of the plugin is too restrictive, and we should now try opening up those restrictions to see "how bad it gets" when we pack up more stuff. The plugin will also warn about the archive being too large, but we should test whether quality and resilience of reproductions improves with making it less strict.

kamilkrzyskow:
> Additionally, there's no need to remove info from plugins. We shouldn't make mkdocs.yml transformations, as they might introduce new bugs which makes the plugin less resilient. This plugin needs to be maximally resilient.

I didn't mean in the plugin. I just stated the fact that after unpacking a packed zip, before running `mkdocs build` locally, we need to remove the `info` plugin manually, as otherwise it will create yet another zip. So removing hooks or anything else shouldn't be a problem either.

> That is very, very weird. I'm not sure why this is the case!?

I don't know why the Python ZipFile allows for that, or why doesn't 7-zip crash when I open the zip later.
However, it happened because I kept the unpacked requirements.lock.txt and platform.json files from your file inside the directory. As this is a very edgy edge case, I don't think there is need to prevent it.

> Thus, customizations and hooks should never be ignored.

> I wouldn't handle the case where customizations are packaged up that are not used.

Happy days, I don't have to do 4D logic in my head to handle this branching logic.

I finally tested https://github.com/squidfunk/mkdocs-material/discussions/2346 - ([9-community-multi-lang.zip](https://github.com/squidfunk/mkdocs-material/files/14426465/9-community-multi-lang.zip))

And it turned out that we have to exclude site_dir based on the existence of `sitemap.xml` and `sitemap.xml.gz`, because the separate `mkdocs.yml` only saw their own site_dir and not of the other languages.
> Note to self: Perhaps checking `sitemap.xml.gz` is enough 🤔 

---

So I think I'm finished, I only need to fixup the commits and squash them into 2 or 3 properly categorized names.
I also noticed that the plugin class or the on_config function specifically grew quite large, so perhaps I should extract the validation and exclusion logic out of the main class into other files 🤔 
Any thoughts?

squidfunk:
> I didn't mean in the plugin. I just stated the fact that after unpacking a packed zip, before running mkdocs build locally, we need to remove the info plugin manually, as otherwise it will create yet another zip. So removing hooks or anything else shouldn't be a problem either.

Only if you run `mkdocs build` instead of `mkdocs serve` or [changed the settings](https://squidfunk.github.io/mkdocs-material/plugins/info/#config.enabled_on_serve). In my daily work, I find it to be quite convenient that it is usually disabled when serving, which is probably 80% of cases. Thus I understand that you're talking about manually removing hooks when _we_ inspect the reproduction. Yes, that's okay, but we should still warn the user and the archive should only proceed packing if forced, like it currently is implemented.

> I don't know why the Python ZipFile allows for that, or why doesn't 7-zip crash when I open the zip later.
However, it happened because I kept the unpacked requirements.lock.txt and platform.json files from your file inside the directory. As this is a very edgy edge case, I don't think there is need to prevent it.

Agreed.

> And it turned out that we have to exclude site_dir based on the existence of sitemap.xml and sitemap.xml.gz, because the separate mkdocs.yml only saw their own site_dir and not of the other languages.

I'm not sure I entirely understand. So, you essentially want to use `sitemap.xml` as a criterion to determine if something is a site directory or not in order to not open the can of worms that is trying to resolve the site directory? #2346 is quite special because it keeps configuration files somewhere else entirely – different from what we now recommend with the projects plugin (more detailed tutorials need to be written, yes) where users should just keep them in the projects. Thus, I'm not sure #2346 is a case we should be optimizing for. Would that change the solution? Sorry if I'm missing something, I'm on and off this topic as I'm currently working on entirely different layers of this project 😅

> So I think I'm finished, I only need to fixup the commits and squash them into 2 or 3 properly categorized names.
I also noticed that the plugin class or the on_config function specifically grew quite large, so perhaps I should extract the validation and exclusion logic out of the main class into other files 🤔
Any thoughts?

I've started modernizing and professionalizing our Python code base with the latest iteration of the [tags plugin](https://github.com/squidfunk/mkdocs-material-insiders/tree/master/src/plugins/tags) in Insiders. Eventually, I want to migrate all plugins to those patterns, since they are now properly documented with Docstrings and cut much, much smaller. If you find a good abstraction of things in the same style, you can definitely go ahead. We can also discuss the conventions I used, as Python is still like Spanish for me – still learning a lot.

squidfunk:
To add on the last part: if this is too much of an ask, please don't bother. I see it as my obligation to do that, but since you asked for refactoring opportunities that could be taken, I wanted to point out my latest advancements in this area. I will definitely refactor the info plugin at some point, but if you want to go ahead and help with that, you are very much invited.

Again, and I can't stress this enough: thanks for being such an awesome part of our inner community! Your time investments into Material for MkDocs are very valued and much appreciated. Great work!

kamilkrzyskow:
> Only if you run mkdocs build instead of mkdocs serve [...]

🙄I forgot that serve won't run info, so yeah my idea was short-sighted. 

> I'm not sure I entirely understand. So, you essentially want to use sitemap.xml as a criterion to determine if something is a site directory or not in order to not open the can of worms that is trying to resolve the site directory? https://github.com/squidfunk/mkdocs-material/discussions/2346 is quite special because it keeps configuration files somewhere else entirely – different from what we now recommend with the projects plugin (more detailed tutorials need to be written, yes) where users should just keep them in the projects. [...]

Recommending something new and better is fine, not supporting something old and "widely" used isn't. 
In-part I agree that making a special behaviour in the plugin specifically for this setup might be a bit much. However, the only other reasonable solution is to provide the user with the `exclude` option in the plugin to let them control what is excluded.
So I think I will just stick to the current setup, exclude resolved `site_url` paths with dynamic patterns, and additionally check for the sitemap files for any folder which wasn't detected.

> If you find a good abstraction of things in the same style, you can definitely go ahead

The most I can do, is to split the `on_config` into different modules for validation, zip creation, and exclusion logic, so if that will be deleted in the future for a more standardized plugin structure, then I won't bother, as it will be just git blame spam 👌.

I'm also more of a "fan" of flat package structures, and your tags plugin implementation is deeply nested, so I would probably need some diagram to see the whole picture 😅I'm just a simple coder, so deep abstraction structures scare me, reminds me of Java-like business grade structures which are the devil 😆
Jokes aside my skill level isn't up to par for me to split up the info plugin into reasonable micro components.

> I've started modernizing and professionalizing our Python code base with the latest iteration of the [tags plugin](https://github.com/squidfunk/mkdocs-material-insiders/tree/master/src/plugins/tags) in Insiders.

This is very much welcome, once the Pylint, the tests, and some auto formatter is introduced (like `black`) it will be way easier to work in the codebase (as currently I have like 10+ warnings in my IDE because the code is styled like TypeScript 😅). Tough the contribution workflow might be more demanding of the users, like installing all of the dependencies then running them before the commit to clean up styles, or people won't like the idea of having to write tests for the PR to be accepted haha

> Your time investments into Material for MkDocs are very valued and much appreciated.

I appreciate your time investment as well, this back and forth in the PR really helped me stay on track and not spiral into endless edge case dimension :v:

kamilkrzyskow:
I've managed to squash 17 commits into 6.
Here are the final generated result zip archives.

[9.5.12+insiders.4.53.0-final-examples-projects.zip](https://github.com/squidfunk/mkdocs-material/files/14453738/9.5.12%2Binsiders.4.53.0-final-examples-projects.zip)
[9.5.12+insiders.4.53.0-final-mkdocstrings.zip](https://github.com/squidfunk/mkdocs-material/files/14453739/9.5.12%2Binsiders.4.53.0-final-mkdocstrings.zip)
[9.5.12+insiders.4.53.0-final-multi-lang-community.zip](https://github.com/squidfunk/mkdocs-material/files/14453743/9.5.12%2Binsiders.4.53.0-final-multi-lang-community.zip)

kamilkrzyskow:
Sorry about the last force push I decided to only check for `sitemap.xml.gz` as the check is done for almost every directory, so it's better to have 2x less checks.

[9.5.11-final-multi-lang-community-single-sitemap-check.zip](https://github.com/squidfunk/mkdocs-material/files/14453859/9.5.11-final-multi-lang-community-single-sitemap-check.zip)

kamilkrzyskow:
The issue template for bugs asks for a minimal reproduction and links to the page where it says to use the info plugin.
But people create the zip files on their own.
Perhaps the task list entry should also mention the the info plugin?
https://github.com/squidfunk/mkdocs-material/blob/5108432a39f0ce4c9326f653ff3e9fc59085bd3c/.github/ISSUE_TEMPLATE/01-report-a-bug.yml?plain=1#L123

squidfunk:
> Recommending something new and better is fine, not supporting something old and "widely" used isn't.

Can you somehow back up that this is widely used? From what I've seen it isn't. Also, "widely" might be hard to define here. We can keep this approach for now, but if we don't receive any reproductions using it in the coming months, we'll consider removing it again. We have it in the history, so we can go back at any time, but in general, less code to maintain is always something to aim for.

> The issue template for bugs asks for a minimal reproduction and links to the page where it says to use the info plugin. But people create the zip files on their own. Perhaps the task list entry should also mention the the info plugin?
> 
> https://github.com/squidfunk/mkdocs-material/blob/5108432a39f0ce4c9326f653ff3e9fc59085bd3c/.github/ISSUE_TEMPLATE/01-report-a-bug.yml?plain=1#L123

CC @katharinalisalin

katharinalisalin:
Great idea, added in ecac4e7ca.

kamilkrzyskow:
> Can you somehow back up that this is widely used?

I said "widely" based on the fact it's the most popular discussion in the `Show and tell` category and discussion board in general:
https://github.com/squidfunk/mkdocs-material/discussions?discussions_q=sort%3Atop

Privacy Guides uses it, though it's my low hanging fruit to support my claim😆 
https://github.com/privacyguides/privacyguides.org

I also checked FastAPI's docs, and they use a similar structure (closer to the `projects` plugin's structure), but they do it more in-house with custom python script that builds their page in different docs. There each config is still only aware of their own `site_dir`, so the custom `sitemap` handling is useful here too. (I also noticed some other thing, read note at the bottom)

I also checked that `sitemap.xml` and `sitemap.xml.gz` are generated for the most minimal MkDocs build:
```yml
site_name: My Docs
plugins: []
```

Perhaps I should remove the dynamic exclusion pattern generation for `site_dir`, as checking for `sitemap.xml.gz` is just as robust, the `projects` plugin might one day do more `sitemap` trickery, so best to keep both methods.
I didn't check performance at all, since minimal reproduction shouldn't really have that much directories to begin with, so optimizing the speed with limiting the amount of patterns didn't really occur to me as necessary.

Note:
When looking at FastAPI I noticed that they do have hooks in `../` paths, so perhaps I should move the file path validation after the customization validation, and if the `stop_on_validation_error` setting is `False` then add the `hooks` and `custom_dir` paths to the `paths_to_validate` list 🤔 

---

Cool that https://github.com/squidfunk/mkdocs-material/pull/6859 got merged, but the fixes for info plugin aren't merged yet, so people could create more reproductions with missing files 😅

squidfunk:
> Cool that https://github.com/squidfunk/mkdocs-material/pull/6859 got merged, but the fixes for info plugin aren't merged yet, so people could create more reproductions with missing files 😅

I'm afraid I'm going to need a little more time to review this. The changes are quite fundamental and #6859 only included documentation changes. Please give me some more time.

kamilkrzyskow:
> Please give me some more time.

I didn't want to come off as rushing you, I just made the observation that the issue template got modified before the plugin was updated, and the that order of things was not optimal.

Anyways, more time in the PR pipeline allows me to spiral in the edge case dimension a bit longer, so I hope my force pushes don't break your review flow 😅

kamilkrzyskow:
After some more thought I believe that any .dotfile in the root directory should be excluded.
The reproduction doesn't need any `.gitignore` `.vscode` files, so the exclusion part needs to be extended 🤔 
And here the limitations of fnmatch get back at me again, as the wildcard `*` will catch everything, so I would have to set the names for file in root manually...

So I'm inclined to replace the fnmatch patterns with regex patterns which would allow to use `^` or `$` to detect the ends of the path 🤔

squidfunk:
> I didn't want to come off as rushing you, I just made the observation that the issue template got modified before the plugin was updated, and that the order of things was not optimal.
> 
> Anyways, more time in the PR pipeline allows me to spiral in the edge case dimension a bit longer, so I hope my force pushes don't break your review flow 😅

I wasn't sure whether this PR was ready, because you did not re-request a review + it appeared to me that you were still working on it. As we made sure we're aligned with the general direction, and I trust you testing the examples we talked about, my work will be mainly to check it out some of the examples to learn about the new UX and glaze over the code base, maybe do some formatting. Just request a review when you're done and I look into it ☺️

I don't think that the merge of #6859 is out of order, since using the info plugin is mandatory as mentioned in the description of the __Reproduction__ section of our issue template, so the PR just tries to catch users not having read that. 

> So I'm inclined to replace the fnmatch patterns with regex patterns which would allow to use ^ or $ to detect the ends of the path 🤔

I'd say we keep it simple first and not try to overthink this. We already changed so much on a very critical component. When we release this with the next version of Material for MkDocs, all people that update will use the new implementation when reporting bugs, so we should try and keep it as minimal as possible while fixing the known issues and aim to reach stability quickly. If we run into problems, we fix them (as we're doing here now).

squidfunk:
Additionally, please move the PR out of draft state, once it is stable from your perspective ☺️

kamilkrzyskow:
> I wasn't sure whether this PR was ready, because you did not re-request a review + it appeared to me that you were still working on it.

Your assessment turned out to be true, I didn't press the button to re-request the review, but I provided the "final" examples here: https://github.com/squidfunk/mkdocs-material/pull/6826#issuecomment-1972000108
So my idea was, that I finished then and there, but as per usual a force push crept in and another followed.

However, this time I hope it's truly final 💪 The change from `fnmatch` to `regex` allowed for a finer control over the patterns.
I decided to exclude all .dotfiles in the root of the reproduction, this includes any `.gitignore` or other `.config` files.
I want to allow `.github` and `.devcontainer`, as this might be related to the issues people are having etc.
So I had to add separate exclusions for IDE directories like .vscode or .idea.

```bash
DEBUG   -  Excluded pattern '.*\.cache/?': /.cache/
DEBUG   -  Excluded pattern '^/\.vscode/': /.vscode/
DEBUG   -  Excluded pattern '/site/': /site/
DEBUG   -  Excluded pattern '/venv/': /venv/
DEBUG   -  Excluded pattern '^/\.[^/]+$': /.editorconfig
DEBUG   -  Excluded pattern '^/\.[^/]+$': /.gitignore
DEBUG   -  Excluded pattern '^/.*\.zip': /7-projects-examples.zip
DEBUG   -  Excluded pattern '/examples/fonts_site_dir/': /examples/fonts_site_dir/
DEBUG   -  Excluded pattern '.*__pycache__/': /examples/__pycache__/
DEBUG   -  Excluded pattern '/examples/blog-basic/site/': /examples/blog-basic/site/
DEBUG   -  Excluded pattern '.*__pycache__/': /hooks/__pycache__/
```

So in the examples repository, `.vscode`, `.cache`, `.editorconfig`, and `.gitignore` got excluded apart of the other things.
[9.5.12+insiders.4.53.0-regex.zip](https://github.com/squidfunk/mkdocs-material/files/14471838/9.5.12%2Binsiders.4.53.0-regex.zip)

I use the `regex` module included in MkDocs dependencies (?) because the docs claim it's more robust when it comes to handling Unicode than the built-in `re`.

src/plugins/info/plugin.py
```diff
@@ -91,7 +98,7 @@ def on_config(self, config):
         # hack to detect whether the custom_dir setting was used without parsing
         # mkdocs.yml again - we check at which position the directory provided
         # by the theme resides, and if it's not the first one, abort.
-        if config.theme.dirs.index(get_theme_dir(config.theme.name)):
+        if config.theme.custom_dir:
```
squidfunk, line 101: Could you explain why this was changed? I hope I didn't miss an explanation in your comments.

src/plugins/info/plugin.py
```diff
@@ -103,6 +110,57 @@ def on_config(self, config):
...
+        # variable is used.
+        custom_dir = config.theme.custom_dir
+        if custom_dir:
+            custom_dir = _convert_to_abs(
+                custom_dir,
+                abs_prefix = config_file_parent
+            )
+
+        # Support projects plugin
+        projects_plugin = config.plugins.get("material/projects")
```
squidfunk, line 128: This is only for path validation, right? I tested packing up our examples repository and commenting out the projets plugin and packing still works. This could be made a little clearer in the comment, because at first I thought that the info plugin explicitly only supports packing up multi-project folders if the projects plugin is enabled.

src/plugins/info/plugin.py
```diff
@@ -18,20 +18,24 @@
...
 from colorama import Fore, Style
 from importlib.metadata import distributions, version
 from io import BytesIO
 from markdown.extensions.toc import slugify
+from mkdocs.config.defaults import MkDocsConfig
 from mkdocs.plugins import BasePlugin, event_priority
-from mkdocs.structure.files import get_files
-from mkdocs.utils import get_theme_dir
+from mkdocs.utils import get_yaml_loader
+import regex
```
squidfunk, line 38: Please move this to line 26, i.e., in the top import block.

src/plugins/info/plugin.py
```diff
@@ -103,6 +110,57 @@ def on_config(self, config):
             log.error("Please remove 'hooks' setting.")
             self._help_on_customizations_and_exit()
 
+        # Assure that possible relative paths, which will be validated
```
squidfunk, line 113: Please make sure that comments always span to 80 characters. We use 80 characters everywhere.

src/plugins/info/plugin.py
```diff
@@ -103,6 +110,57 @@ def on_config(self, config):
...
+        projects_plugin = config.plugins.get("material/projects")
+        if projects_plugin:
+            abs_projects_dir = _convert_to_abs(
+                projects_plugin.config.projects_dir,
+                abs_prefix = config_file_parent
+            )
+        else:
+            abs_projects_dir = ""
+
+        # Load the current MkDocs config(s) to get access to INHERIT
```
squidfunk, line 137: Could you add some more info in the comments on the different design decisions? We discussed the implementation quite a lot, and our decisions and the limitations those decisions try to workaround should be commented here. You can check other plugins to see how I'm doing this, because when somebody wants to refactor the code, the implications should be clear.

Example:

https://github.com/squidfunk/mkdocs-material/blob/15f15068d958506734d3544108a1e6971d7e511b/src/plugins/blog/plugin.py#L171-L184

src/plugins/info/plugin.py
```diff
@@ -103,6 +110,57 @@ def on_config(self, config):
...
+            custom_dir or "",
+            abs_projects_dir,
+            *[cfg.get("INHERIT", "") for cfg in loaded_configs]
+        ]
+
+        for hook in config.hooks:
+            path = _convert_to_abs(hook, abs_prefix = config_file_parent)
+            paths_to_validate.append(path)
+
+        for path in list(paths_to_validate):
```
squidfunk, line 156: Also please add a comment why this is done.

src/plugins/info/plugin.py
```diff
@@ -112,18 +170,57 @@ def on_config(self, config):
...
-                if os.path.isfile(path):
+            for abs_root, dirnames, filenames in os.walk(os.getcwd()):
+                # Prune the folders in-place to prevent
+                # scanning excluded folders
+                for name in list(dirnames):
+                    path = os.path.join(abs_root, name)
+                    if self._is_excluded(_resolve_pattern(path)):
+                        dirnames.remove(name)
+                        continue
+                    # Multi-language setup from #2346 separates the
```
squidfunk, line 208: Please create [t.ly](https://t.ly/) links to issues, so they can be looked up quickly.

Example:

https://github.com/squidfunk/mkdocs-material/blob/15f15068d958506734d3544108a1e6971d7e511b/src/plugins/blog/plugin.py#L768-L771

src/plugins/info/plugin.py
```diff
@@ -112,18 +170,57 @@ def on_config(self, config):
...
+                    # language config, so each mkdocs.yml file is
+                    # unaware of other site_dir directories. Therefore,
+                    # we add this with the assumption a site_dir contains
+                    # the sitemap file.
+                    sitemap_gz = os.path.join(path, "sitemap.xml.gz")
+                    if os.path.exists(sitemap_gz):
+                        log.debug(f"Excluded site_dir: {path}")
+                        dirnames.remove(name)
+                for name in filenames:
```
squidfunk, line 217: Please add some more inline comments that make it easier to understand what is happening in which order. This might seem pedantic, but will make maintenance down the line much simpler.

src/plugins/info/plugin.py
```diff
@@ -217,12 +320,47 @@ def _help_on_customizations_and_exit(self):
...
+    def _help_on_not_in_cwd(self, bad_paths):
+        print(Fore.RED)
+        print("  The current working (root) directory:\n")
+        print(f"    {os.getcwd()}\n")
+        print("  is not a parent of the following paths:")
+        print(Style.NORMAL)
+        for path in bad_paths:
+            print(f"    {path}")
+        print()
+        print("  To assure that all project files are found")
```
squidfunk, line 344: Please format to 80 character limit.

src/plugins/info/plugin.py
```diff
@@ -237,6 +375,81 @@ def _size(value, factor = 1):
...
+# check the prefix.
+def _convert_to_abs(path: str, abs_prefix: str = None) -> str:
+    if os.path.isabs(path): return path
+    if abs_prefix is None: abs_prefix = os.getcwd()
+    return os.path.normpath(os.path.join(abs_prefix, path))
+
+# Custom YAML loader - required to handle the parent INHERIT config.
+# It converts the INHERIT path to absolute as a side effect.
+# Returns the loaded config, or a list of all loaded configs.
+def _load_yaml(abs_src_path: str):
```
squidfunk, line 389: From what I read here, it is not apparent to me why we need to implement this ourselves and can't just use MkDocs loading functionality. It looks like a duplication of. It is a good idea to explain this in the comment.

Example:

https://github.com/squidfunk/mkdocs-material/blob/15f15068d958506734d3544108a1e6971d7e511b/src/plugins/blog/structure/__init__.py#L59-L75

src/plugins/info/plugin.py
```diff
@@ -103,6 +110,57 @@ def on_config(self, config):
...
+            )
+        else:
+            abs_projects_dir = ""
+
+        # Load the current MkDocs config(s) to get access to INHERIT
+        loaded_configs = _load_yaml(config.config_file_path)
+        if not isinstance(loaded_configs, list):
+            loaded_configs = [loaded_configs]
+
+        # Validate different MkDocs paths to assure that
```
squidfunk, line 142: Comments are a little inconsistent in formatting. Here, a `.` is used to delimit a one-sentence comment, at other places there's no `.`. We follow the style: If it's only one sentence, no `.` should be added. Otherwise, in case of more sentences, add `.`.

squidfunk (changes requested):
From a functional perspective, this is top-notch! I only have primarily formatting- and comment-related remarks that we should address before merging. I hope that it's not too pendantic, but I always try to put myself in my own shoes 3 months from now when I have forgotten what and why I did specific things. Especially when I jump between TypeScript and Python, extensive comments on design and architecture decisions make it much easier to rebuild an overview of the project after some time has passed ☺️

src/plugins/info/plugin.py
kamilkrzyskow: Since version 1.5.0 MkDocs exposes the `Theme.custom_dir` property: https://www.mkdocs.org/about/release-notes/#upgrades-for-plugin-developers

Before that you had to fetch it with some more or less hacky ways like `Theme._vars` or like you with `Theme.dirs`.

Since I used the `theme.custom_dir` value for validation, I guessed that would be the right way to go to have consistency in fetching the value.

src/plugins/info/plugin.py
kamilkrzyskow: Not just path validation, but also to be able to dynamically exclude the `site_dir` in the projects plugin. 
When I first wrote it I didn't want to use the `sitemap.xml` file to detect site directories, but ultimately decided to do that for the multi-language directory.

I don't think that the option to disable projects file gathering when the projects plugin is disabled is worth pursuing.

Instead to minimise the code churn, I'll just remove it and the projects site directories will be excluded just like the community multi-language projects.

Good catch 👌

src/plugins/info/plugin.py
kamilkrzyskow: Am I right to understand you split the import blocks into:
1. Simple `import xyz`
2. `from foo import bar`
3. `from .internal import util`
?

That's a bit unconventional as typically it's:
1. Import built-in
2. Import external installed 
3. Import internal

My `import regex` doesn't meet any criteria, a well spotted mistake, but I wanted to used to be sure.

squidfunk: Yes, currently we use that convention. Please stick to it for now for consistency. We can change it in the future.

src/plugins/info/plugin.py
squidfunk: I'm sorry, but I don't understand what you mean with your last sentence (in store?)

src/plugins/info/plugin.py
squidfunk: Ah, I see! Yes, that makes sense.

src/plugins/info/plugin.py
kamilkrzyskow: Should be "in-store" I guess. I meant that I don't know what are the plans for projects plugin and sitemap manipulation. 
As per the current limitations: https://squidfunk.github.io/mkdocs-material/plugins/projects/#limitations

In the future the projects plugin will somehow manage the sitemap generation, so perhaps there will be an option to disable sitemap creation (?). This is the edge-case dimension all over again 😅 

If you're sure that the current behaviour will be respected and there will always be a `sitemap.xml.gz` file either in every site directory or the global projects site directory, then I could remove the dynamic exclusion part and just keep the exclusion based on the `sitemap.xml.gz` for the build directories.

squidfunk: Ah, I see, thanks for clarifying! Yes, we might change the way how the sitemap is generated to allow for a better consolidated experience, thus it might be possible that nested `site_dir` folders do not contain sitemaps. However, nothing to worry about right now.

src/plugins/info/plugin.py
kamilkrzyskow: I think I elaborated a bit better, but I'm not sure if it's resolved, so let me know if not 👌

squidfunk: Looks good ☺️

squidfunk (approved):

squidfunk:
Thanks again for your awesome work! We'll release it in the coming days!

squidfunk:
I just updated the dependencies in dc808ca90 and had to make [a change to the build process](https://github.com/squidfunk/mkdocs-material/commit/dc808ca90a8fde71c95708addb1b4773ae2acdce#diff-c4099a01f071d5e9de212bcfeeb9fcfbdefabcc26a781bb8e9bee51d41d3a5d4), because the `info.gitignore` file would otherwise be deleted. I'm not sure why it built for you, but for me it would be gone, breaking the plugin 😅

kamilkrzyskow:
I got caught red handed I didn't completely follow the contribution guide. At some point I stopped rebuilding the project, but instead just copied the material/plugins/info directory to the src/plugins directory, because I noticed it doesn't make a difference 😅

Seeing the code responsible for copying the config files and assets perhaps naming the file .gitignore wasn't the best idea.

Especially now after I stopped using fnmatch and started using regex, I think this should be it's own patterns.py file with a list of patterns in python which is later simply imported.

Of course I didn't think of this sooner because it worked, and I didn't stop to think of the build process.

From an IDE perspective the current file is recognised as a .gitignore file while it's not such file 😵‍💫

If you won't do it sooner with the plugin rewrite, I will revisit this again in the future, and fixup the methodology of storing patterns.

Unless I should do it immediately?

squidfunk:
> Especially now after I stopped using fnmatch and started using regex, I think this should be it's own patterns.py file with a list of patterns in python which is later simply imported.

Makes sense! The next release will take 2-3 days at least, so if you can squeeze it in, you can go ahead. If not, no problem, but then we should add a comment with a `@todo` that this should be done in the future.

kamilkrzyskow:
Great, will do it today :v:

PR #6878, opened 2024-03-06
docs: fix pagination_url_format default value

berk-karaal:
The default value of `pagination_url_format` is `page/{page}` as declared [here](https://github.com/squidfunk/mkdocs-material/blob/c8a2142263e5b8542c4546644b73d354850253b4/src/plugins/blog/config.py#L77). Fixed incorrectly set default value in documentation.


squidfunk:
Thanks! Definitely a copy and paste error 😅

PR #6883, opened 2024-03-08
Add missing comma for finnish translation

todeveni:


squidfunk:
Ah, good catch. That slipped by me 😬

PR #6907, opened 2024-03-13
feature: make JSON feed discoverable

Guts:
Since its [version 1.12](https://github.com/Guts/mkdocs-rss-plugin/releases/tag/1.12.0) ([plugin's documentation](https://guts.github.io/mkdocs-rss-plugin/integrations/?h=json#reference-json-feeds-in-html-meta-tags)), the RSS plugin creates also a [JSON Feed](https://www.jsonfeed.org/version/1.1/).

In this PR, I propose to add the same feed discovery mechanism as for RSS in the base template.

### DIFF
--- a/material/templates/partials/languages/en.html
+++ b/material/templates/partials/languages/en.html
@@ -25,6 +25,8 @@
   "footer.next": "Next",
   "footer.previous": "Previous",
   "header": "Header",
+  "json_feed.created": "JSON feed",
+  "json_feed.updated": "JSON feed of updated content",
   "meta.comments": "Comments",
   "meta.source": "Source",
   "nav": "Navigation",
--- a/material/templates/partials/languages/fr.html
+++ b/material/templates/partials/languages/fr.html
@@ -24,6 +24,8 @@
   "footer.next": "Suivant",
   "footer.previous": "Précédent",
   "header": "En-tête",
+  "json_feed.created": "Flux JSON",
+  "json_feed.updated": "Flux JSON du contenu mis à jour",
   "meta.comments": "Commentaires",
   "meta.source": "Source",
   "nav": "Navigation",
--- a/src/templates/base.html
+++ b/src/templates/base.html
@@ -60,7 +60,7 @@
         <link rel="next" href="{{ page.next_page.url | url }}" />
       {% endif %}
 
-      <!-- RSS feed -->
+      <!-- RSS and JSON feed -->
       {% if "rss" in config.plugins %}
         <link
           rel="alternate"
@@ -73,6 +73,18 @@
           type="application/rss+xml"
           title="{{ lang.t('rss.updated') }}"
           href="{{ 'feed_rss_updated.xml' | url }}"
+        />
+        <link 
+          rel="alternate"
+          type="application/feed+json"
+          title="{{ lang.t('json_feed.created') }}"
+          href="{{ 'feed_rss_createdupdated.json' | url }}"
+        />
+        <link 
+          rel="alternate"
+          type="application/feed+json"
+          title="{{ lang.t('json_feed.updated') }}"
+          href="{{ 'feed_rss_updated.json' | url }}"
         />
       {% endif %}
 
--- a/src/templates/base.html
+++ b/src/templates/base.html
@@ -78,7 +78,7 @@
           rel="alternate"
           type="application/feed+json"
           title="{{ lang.t('json_feed.created') }}"
-          href="{{ 'feed_rss_createdupdated.json' | url }}"
+          href="{{ 'feed_rss_created.json' | url }}"
         />
         <link 
           rel="alternate"


squidfunk:
Thanks for the PR! Is that actually a _widely_ supported standard like RSS? It was also not updated in the last 4 years, it seems (scroll down and see 2017-2020), so it looks kind of abandoned. We can't add support for each and every format, only for those that are widely supported and provide a lot of value. Everything else can be added via customization.

kamilkrzyskow:
Looking at https://guts.github.io/mkdocs-rss-plugin/configuration/#json_feed_enabled and https://guts.github.io/mkdocs-rss-plugin/configuration/#rss_feed_enabled this is configurable. 
So it should also check for those values in the template?

Guts:
> Thanks for the PR! Is that actually a widely supported standard like RSS? It was also not updated in the last 4 years, it seems (scroll down and see 2017-2020), so it looks kind of abandoned

I've to say that I don't know. It has been contributed by another dev (I've just finished its work) so I assume that it's still in use. 

> We can't add support for each and every format, only for those that are widely supported and provide a lot of value. Everything else can be added via customization.

Sure, I get the rationale but you do not add any support here, as you don't "support" RSS format.

> Looking at https://guts.github.io/mkdocs-rss-plugin/configuration/#json_feed_enabled and https://guts.github.io/mkdocs-rss-plugin/configuration/#rss_feed_enabled this is configurable.
So it should also check for those values in the template?

I've kept the same behavior with rss output which can also be disabled.

Guts:
Well @squidfunk @kamilkrzyskow, no worries I'll make it on my own side, it's only 2 lines in a base template after all.

squidfunk:
@Guts if it's not widely used, we're currently not considering adding it – in the end, it's more to maintain, even if it's only two lines. Additionally, for us it would be two new translations for 65 languages 😅

kamilkrzyskow:
> I've kept the same behavior with rss output which can also be disabled.

Well, I can see that, but that doesn't mean this is correct, what will happen to the link if the plugin doesn't save a file? It will return a 404, so not ideal IMO. But perhaps just the way the web behaves? I don't know if that is the proper way to do things therefore I asked the question if there should be the conditionals for the file creation.

> it's more to maintain, even if it's only two lines. 

It's not my project, nor do I even come close to developing something for years, so I'm not debating anything here. 
I just want to observe that I feel like there is nothing to maintain here.  The links are in the `<head>` tag and won't be moved elsewhere, maybe another partial (?), and if something happens and you stop supporting the RSS plugin, which supports JSON feeds, you can always remove those 2 lines. Basically your future decision about the development won't be bound by those 2 lines.

> Additionally, for us it would be two new translations for 65 languages

In this instance, as both RSS and JSON are acronyms/abbreviations I feel like this could be _easily_ fixed with a refactor to 
```json
  "content_feed.created": "XYZ feed",
  "content_feed.updated": "XYZ feed of updated content",
```
and then `XYZ` would be later be replaced with the correct short name. Of course given 65 languages there could be an edge case where that doesn't work...

@Guts Maybe you could override the templates, similarly to what I do in https://github.com/squidfunk/mkdocs-material/discussions/5060, either change the Materials templates or detect there is a custom_dir with `main.html` and add a extrahead block for the user 🤔

squidfunk:
> It's not my project, nor do I even come close to developing something for years, so I'm not debating anything here.
I just want to observe that I feel like there is nothing to maintain here. The links are in the <head> tag and won't be moved elsewhere, maybe another partial (?), and if something happens and you stop supporting the RSS plugin, which supports JSON feeds, you can always remove those 2 lines. Basically your future decision about the development won't be bound by those 2 lines.

From my perspective, when I talk about maintaining, it doesn't only mean shoving lines of code around. It means committing to supporting something for the years to come, be it updates, questions, bugs, issues, etc. Material for MkDocs already has a huge code base and I'm currently the main person making changes to it, apart form occasional PRs. If something goes into our code base, we commit to it and individuals and organizations can rely on us doing our best to "make it work". This also means we need to allocate and budget time for it. I'm sorry if this sometimes looks as if I'm a "no-sayer", but if I learned one thing it's that to keep the scope and focus sharp, you have to say no _a lot_. That's at least my/our philosophy, and of course not everybody has to agree with it. We're always happy to discuss potential changes and ideas. However, we also expect our decisions are respected. After all, this project is Open Source. It can be forked and adapted to your liking.

squidfunk:
Additionally, it is _trivial_ to add support for this with [customization](https://squidfunk.github.io/mkdocs-material/customization/#overriding-blocks). When we say no, we always look for an escape hatch for the user, so that the user, with a little more effort, can still achieve the asked thing. Here, it's no more than:

``` jinja
{% block extrahead %}
  <link 
    rel="alternate"
    type="application/feed+json"
    title="JSON feed"
    href="{{ 'feed_rss_created.json' | url }}"
  />
  <link 
    rel="alternate"
    type="application/feed+json"
    title="JSON feed of updated content"
    href="{{ 'feed_rss_updated.json' | url }}"
  />
{% endblock %}
```

PR #6916, opened 2024-03-15
Let browser decide how to properly render text

ryuukk:
Grayscale = blurry text with Skia on linux (probably other OSes too nowadays) on low DPI screen (both chromium/firefox)

This issue with grayscale is much worse with light text on dark background, text becomes unreadable
 

https://github.com/mkdocs/mkdocs/issues/3488

https://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/

### DIFF
--- a/src/templates/assets/stylesheets/main/_typeset.scss
+++ b/src/templates/assets/stylesheets/main/_typeset.scss
@@ -26,8 +26,8 @@
 
 // Enable font-smoothing in Webkit and FF
 body {
-  -webkit-font-smoothing: antialiased;
-  -moz-osx-font-smoothing: grayscale;
+  -webkit-font-smoothing: auto;
+  -moz-osx-font-smoothing: auto;
 
   // Font with fallback for body copy
   --md-text-font-family:


squidfunk:
Thanks for the PR. This is a _very_ fundamental change that impacts all users, which we're not considering just shoving onto our users at the moment. You can easily add this via extra CSS, but we're currently not considering adding this to `master`.

I hope for your understanding.

alexvoss:
@ryuuk, that was uncalled for. @squidfunk pointed out that you can change the behavior via a customization, which seems like a perfectly sensible course of action. As you well know, this has been discussed before and no conclusion was reached as to what the best way would be to fix this in the theme. In determining this, @squidfunk will always have to consider the impact on all users. 

Looking back at previous discussions about the matter, this is not the first time that you show a distinct lack of courtesy. I'd take a moment to consider apologizing. I draw your attention to the [code of conduct](https://github.com/squidfunk/mkdocs-material/blob/2f1b2e950040a89e70a5f193de762388206fb8fa/CODE_OF_CONDUCT.md).

rYuuk:
> @ryuuk, that was uncalled for. @squidfunk pointed out that you can change the behavior via a customization, which seems like a perfectly sensible course of action. As you well know, this has been discussed before and no conclusion was reached as to what the best way would be to fix this in the theme. In determining this, @squidfunk will always have to consider the impact on all users and has to consider what works on a range of browsers and operating systems.
> 
> 
> 
> Looking back at previous discussions about the matter, this is not the first time that you show a distinct lack of courtesy. I'd take a moment to consider apologizing. I draw your attention to the [code of conduct](https://github.com/squidfunk/mkdocs-material/blob/2f1b2e950040a89e70a5f193de762388206fb8fa/CODE_OF_CONDUCT.md).

I think you tagged the wrong person 😅

alexvoss:
>  think you tagged the wrong person

So sorry, not sure how that happened. Apologies again. I've edited the post, hope that has the desired effect.

squidfunk:
@ryuukk if you do not agree with the way we maintain this project, you are free to use something else. This is, after all, a free offering, as it is Open Source and has a permissive license. We try our best to cater to all users, but for this we often need to strike a balance. Additionally, we provide multiple ways how you can customize it, and doing that would involve adding 5 lines (2 to `mkdocs.yml`, 3 to a new CSS file). I think that is a reasonable ask.

This is also [your first warning](https://squidfunk.github.io/mkdocs-material/contributing/#fn:1).

Issue #6917, opened 2024-03-15
Blog: `UTF-8` encoding problems on Windows


avanbremen:
### Bug description

When adding the [built-in blog plugin](https://squidfunk.github.io/mkdocs-material/setup/setting-up-a-blog/#built-in-blog-plugin) and [writing your first post](https://squidfunk.github.io/mkdocs-material/setup/setting-up-a-blog/#writing-your-first-post), spinning up the [live preview server](https://squidfunk.github.io/mkdocs-material/creating-your-site/#previewing-as-you-write) will abort with a `BuildError` (__init__.py [line 73](https://github.com/squidfunk/mkdocs-material/blob/master/src/plugins/blog/structure/__init__.py#L73)):

```
Error reading metadata of post 'blog\posts\hello-world.md' in 'docs':
Expected metadata to be defined but found nothing
```

Including YAML style meta-data in other (not blog) Markdown files will not result in a `BuildError`.

A minimal reproduction project is included.

Tested with [mkdocs-material-9.4.13](https://github.com/squidfunk/mkdocs-material/releases/tag/9.5.13) and [mkdocs-1.5.3](https://github.com/mkdocs/mkdocs/releases/tag/1.5.3).

### Related links

- [Built-in blog plugin](https://squidfunk.github.io/mkdocs-material/setup/setting-up-a-blog/#built-in-blog-plugin)
- [Writing your first post](https://squidfunk.github.io/mkdocs-material/setup/setting-up-a-blog/#writing-your-first-post)

### Reproduction

[material-mkdocs-blog.zip](https://github.com/squidfunk/mkdocs-material/files/14617610/material-mkdocs-blog.zip)

### Steps to reproduce

1. Add [built-in blog plugin](https://squidfunk.github.io/mkdocs-material/setup/setting-up-a-blog/#built-in-blog-plugin)
2. [Write your first post](https://squidfunk.github.io/mkdocs-material/setup/setting-up-a-blog/#writing-your-first-post)
3. Spin up the [live preview server](https://squidfunk.github.io/mkdocs-material/creating-your-site/#previewing-as-you-write)

### Before submitting

- [X] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
[...]

PR #6923, opened 2024-03-17
Fixed UTF-8 with BOM encoding support

kamilkrzyskow:
PR for #6917 

Changed `utf-8` to `utf-8-sig` in all instanced where a file is read, that the user could've created with an external editor. 
Kept `utf-8` for reading files that are generated or already provided by the theme.
I decided to not add comments to each changed line, as supporting BOM isn't a priority, and if someone forgets using `-sig` it won't be the end of the world.

Tested by serving `mkdocs-material` documentation with this Community version of the codebase, and there was no crash. 
Therefore the privacy plugin wasn't tested, but those changed got tested in Insiders.
https://github.com/squidfunk/mkdocs-material-insiders/pull/82

### DIFF
--- a/material/plugins/blog/plugin.py
+++ b/material/plugins/blog/plugin.py
@@ -460,7 +460,7 @@ class BlogPlugin(BasePlugin[BlogConfig]):
             return config.authors
 
         # Open file and parse as YAML
-        with open(file, encoding = "utf-8") as f:
+        with open(file, encoding = "utf-8-sig") as f:
             config.config_file_path = os.path.abspath(file)
             try:
                 config.load_dict(yaml.load(f, SafeLoader) or {})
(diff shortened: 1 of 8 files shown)


squidfunk (approved):

squidfunk:
Awesome, as always, thanks for your work and time ❤️

alexvoss:
You guys are on fire! Updating now so I can holler should anything break.

squidfunk:
@kamilkrzyskow is on fire. I'm only the click-merge-man here ✌️

kamilkrzyskow:
It's nice to have such small PRs, where I can be 99% sure I wouldn't do yet another force push with some second guess change 🤣
Thanks for the fast merge ✌️

PR #6933, opened 2024-03-19
Fix broken reference link

StevenMaude:
To "commercial use" section.

### DIFF
--- a/docs/insiders/faq/sponsoring.md
+++ b/docs/insiders/faq/sponsoring.md
@@ -125,7 +125,7 @@ depending on the number of sites you want to build. For companies using
 Material for MkDocs, we offer three different __commercial__ tiers, from which
 you can choose depending on your requirements.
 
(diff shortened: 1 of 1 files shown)


docs/insiders/faq/sponsoring.md
```diff
@@ -125,7 +125,7 @@ depending on the number of sites you want to build. For companies using
 Material for MkDocs, we offer three different __commercial__ tiers, from which
 you can choose depending on your requirements.
 
-Also, please read what is considered [commercial use].
+Also, please read what is considered [commercial] use.
```
StevenMaude, line 128: An alternative would be to make "commercial use` the linked text, but the syntax is clunkier, so I went with the cleaner version:

```suggestion
Also, please read what is considered [commercial use][commercial].
```

Feel free to use whatever you prefer though :smile:

squidfunk:
Thanks! LGTM.

PR #6937, opened 2024-03-19
Fetch KaTeX and MathJax from UNPKG

vedranmiletic:
This is consistent with how we fetch Mermaid.js and automatically picks the latest minor version.

### DIFF
--- a/docs/plugins/privacy.md
+++ b/docs/plugins/privacy.md
@@ -387,7 +387,7 @@ external assets for different origins:
 plugins:
   - privacy:
       assets_exclude: # (1)!
-        - cdn.jsdelivr.net/npm/mathjax@3/*
+        - unpkg.com/mathjax@3/*
         - giscus.app/*
 ```
 
--- a/docs/reference/math.md
+++ b/docs/reference/math.md
@@ -65,7 +65,7 @@ to your `mkdocs.yml`.
     extra_javascript:
       - javascripts/mathjax.js
       - https://polyfill.io/v3/polyfill.min.js?features=es6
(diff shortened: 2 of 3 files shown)


vedranmiletic:
Never mind, minified version on jsDelivr is unfortunately broken.

vedranmiletic:
I found a way to fix this.

vedranmiletic:
FWIW, I tested it in deployment as well, e.g. you can see it live [on this page with a formula](https://group.miletic.net/hr/nastava/materijali/simulator-ns3-tcp-upravljanje-zagusenjem/#varijante-tcp-tahoe-i-tcp-reno).

squidfunk:
What's the upside? Why is jsdeliver better than Cloudflare?

vedranmiletic:
For me, I like that it automatically picks the latest minor version, no need to modify the config every time KaTeX puts out a new minor (or patch) release.

squidfunk:
Thanks for the explanation. I'd say we switch to unpkg then – same result:

```
https://www.unpkg.com/browse/katex@0/dist/katex.min.js
```

Would you like to adjust the PR?

vedranmiletic:
Gladly. Would you mind if I switch MathJax from jsDelivr to unpkg at the same time?

squidfunk:
We should switch everything to unpkg if possible ✌️

vedranmiletic:
Done and done. 😎

squidfunk:
Alright, let's go!

vedranmiletic:
Thank you!

Issue #6983, opened 2024-03-28
⚠️ Social cards broken due to sudden Google Fonts API change


aronmolnar:
### Context

The social plugin tries to fetch fonts from Google and to unzip them:  
https://github.com/squidfunk/mkdocs-material-insiders/blob/4ff6a572151f684fe2e21527582e1096a88805d0/src/plugins/social/plugin.py#L857

### Bug description

Google no longer provides direct access to the zip file, so building sites fails.

```
  File "/usr/local/lib/python3.12/site-packages/material/plugins/social/plugin.py", line 522, in _render_typography
    path = self._resolve_font(family, style, variant)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/material/plugins/social/plugin.py", line 809, in _resolve_font
    self._fetch_font_from_google_fonts(family)
  File "/usr/local/lib/python3.12/site-packages/material/plugins/social/plugin.py", line 876, in _fetch_font_from_google_fonts
    archive = ZipFile(f)
              ^^^^^^^^^^
  File "/usr/local/lib/python3.12/zipfile/__init__.py", line 1341, in __init__
    self._RealGetContents()
  File "/usr/local/lib/python3.12/zipfile/__init__.py", line 1408, in _RealGetContents
    raise BadZipFile("File is not a zip file")
zipfile.BadZipFile: File is not a zip file
```

Workaround was disabling the social plugin.

### Related links

- [Reporting a bug](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/)

### Reproduction

Using the code referenced above shows that the link provides HTML data.

### Steps to reproduce

```
import requests
family= "Noto Sans"
url = f"https://fonts.google.com/download?family={family}"
res = requests.get(url, stream = True)
for chunk in res.iter_content(chunk_size = 8192):
    print(chunk)

<script data-id="_gd" nonce="lai0FBQz_50-Cv_k8L_woA">window.WIZ_global_data
...snip...
```

### Before submitting

- [X] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
[...]

PR #6987, opened 2024-03-29
fix for #6983

alexvoss:
Google Font download no longer provides zip files, so I added an option to source fonts from a .zip file in a local directory configured with `fonts_dir`, just as in the PR for the Insider Edition. 

The code changes here for the public version are a bit smaller as I accepted a little bit of duplication.


squidfunk:
Superseded by ad7233640f2c679b7f149def57c96e1ec12b9ad9. Currently no plans to add `fonts_dir`. However, well reconsider next time we work on the social plugin's font loading and resolution mechanics.

PR #6988, opened 2024-03-29
Fix unicode, case-sensitive `slugify` setting in the documentation

kuanyili:
Unicode, case-sensitive reference setting causes error below.

```
yaml.constructor.ConstructorError: expected a mapping node, but found scalar
```

Update related sections in the documentation to address this issue.

Newer Python-Markdown has `slugify_unicode` built-in, so this

```yaml
markdown_extensions:
  - toc:
      slugify: !!python/name:markdown.extensions.toc.slugify_unicode
```

also works for case-insensitive use cases.

Fixes: ee1e675da613 ("Update slug reference")

## References

- https://pyyaml.org/wiki/PyYAMLDocumentation#objects
- https://python-markdown.github.io/extensions/toc/#usage
- https://facelessuser.github.io/pymdown-extensions/extensions/tabbed/#options

### DIFF
--- a/docs/setup/extensions/python-markdown.md
+++ b/docs/setup/extensions/python-markdown.md
@@ -257,7 +257,7 @@ The following configuration options are supported:
         ``` yaml
         markdown_extensions:
           - toc:
-              slugify: !!python/object/apply:pymdownx.slugs.slugify
+              slugify: !!python/object/apply:pymdownx.slugs.slugify {}
         ```
 
 <!-- md:option toc.toc_depth -->
--- a/docs/setup/extensions/python-markdown-extensions.md
+++ b/docs/setup/extensions/python-markdown-extensions.md
@@ -686,7 +686,7 @@ The following configuration options are supported:
 
 <!-- md:option pymdownx.tabbed.slugify -->
 
-:   <!-- md:default `toc.slugify` --> This option allows for
+:   <!-- md:default `None` --> This option allows for
     customization of the slug function. For some languages, the default may not
     produce good and readable identifiers – consider using another slug function
     like for example those from [Python Markdown Extensions][Slugs]:
@@ -706,7 +706,7 @@ The following configuration options are supported:
         ``` yaml
         markdown_extensions:
           - pymdownx.tabbed:
-              slugify: !!python/object/apply:pymdownx.slugs.slugify
+              slugify: !!python/object/apply:pymdownx.slugs.slugify {}
         ```
 
 The other configuration options of this extension are not officially supported


squidfunk:
Thanks for the PR. Could you please provide a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/) here? Before we can consider merging a PR that fixes an error, we must first ensure that it actually is an error. This is also why, before creating an issue, we recommend to [discuss changes with us](https://squidfunk.github.io/mkdocs-material/contributing/making-a-pull-request/#preparing-changes-and-draft-pr). (info box) We cannot just change things without understanding their impact. However, in this case, you can attach the minimal reproduction which showcases the problem here. I hope you understand that with a project of this size, we have to have processes and we must ask our users to stick to them. Otherwise it would be unmanageable.

kuanyili:
Ah, totally missed that! My apologies and thanks for the pointer.

Here is the minimal reproduction.

[fix-slugify.zip](https://github.com/squidfunk/mkdocs-material/files/14809329/fix-slugify.zip)

docs/setup/extensions/python-markdown.md
```diff
@@ -237,7 +237,7 @@ The following configuration options are supported:
 
 <!-- md:option toc.slugify -->
 
-:   <!-- md:default `toc.slugify` --> This option allows for
+:   <!-- md:default `markdown.extensions.toc.slugify` --> This option allows for
```
squidfunk, line 240: Please don't use the fully qualified syntax here – there's no need.

squidfunk (changes requested):
Thanks for the PR, cosmetic change.

@facelessuser could I ask you to look over this? I was erroneously assuming that this syntax works, but it seems we always need to provide a mapping? Can this PR be improved from your perspective?

facelessuser:
@squidfunk to be honest, I didn't think to explicitly document the empty parameter in case in the [FAQ](https://facelessuser.github.io/pymdown-extensions/faq/#function-references-in-yaml) 😅.

Because this `slugify_unicode` function returns the required slugify based on parameter input, if you are providing no inputs, then yes, you must use `{}` to specify no parameters.

So this for no parameters:

```yaml
slugify: !!python/object/apply:pymdownx.slugs.slugify {}
```

This to call specific keyword parameters:

```yaml
slugify: !!python/object/apply:pymdownx.slugs.slugify {kwds: {case: lower}}
```

facelessuser:
I've updated the FAQ here: [88a94636a2904c33829037369ca7155f6abc2a62](https://github.com/facelessuser/pymdown-extensions/commit/88a94636a2904c33829037369ca7155f6abc2a62)

squidfunk:
Perfect, thanks! It was a misunderstanding on my side, which led to a wrong code snippet on our documentation. This PR fixes this – thanks for you both!

PR #7054, opened 2024-04-16
fix: was post_slugify instead of categories_slugify in example

alexvoss:
wrong setting name was used in the example for `categories_slugify`.


squidfunk:
LGTM, thanks!

PR #7084, opened 2024-04-24
Remove outdated comment now that GitHub has caught up

jeffwidman:
This is no longer necessary: https://github.com/orgs/community/discussions/6456#discussioncomment-5244057


squidfunk:
Nice! This means we can remove `setup.py`, right?

jeffwidman:
Yes, I hadn't done that as you currently have it setup as a warning file, but if you prefer I can delete it entirely... I know I'd prefer to do that!

squidfunk:
Let's get rid of it, as it has no purpose anymore 🧹

jeffwidman:
💥

squidfunk:
Perfect, thanks for your time!

Issue #6675, opened 2024-01-22
Failed to see any articles under my "Blog" column


Pisces032:
### Context

I'm trying to build my own blog website. I followed the instructions on https://squidfunk.github.io/mkdocs-material/, but my "Blog" page couldn't work.

### Bug description

I had tried my website by "mkdocs serve" without any problems. However, after I uploaded it on GithubPages, I can't see my blogs. You can see details from my repository "P1sc3s007JustMyBlog".
This is my results in "mkdocs serve", which works well. 
<img width="1245" alt="1" src="https://github.com/squidfunk/mkdocs-material/assets/114841520/3128ce5f-0ae7-47b8-87a5-4261807b456c">
This is when is was upload. 
<img width="1245" alt="2" src="https://github.com/squidfunk/mkdocs-material/assets/114841520/3f17968d-3926-4c36-afe9-9746d19bd02e">

### Related links

- [Reporting a bug](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/)
- [My blog](https://github.com/Pisces032/P1sc3s007JustMyBlog)

### Reproduction

[blog根.zip](https://github.com/squidfunk/mkdocs-material/files/14008330/blog.zip)

### Steps to reproduce

I don't know how to solve it?

### Before submitting

- [X] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
- [X] I have attached links to [the documentation](https://squidfunk.github.io/mkdocs-material/), and possibly related [issues](https://github.com/squidfunk/mkdocs-material/issues) and [discussions](https://github.com/squidfunk/mkdocs-material/discussions).
- [X] I assure that I have [removed all customizations](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/#remove-customizations) before submitting this bug report.
- [X] I have attached a __.zip file__ with a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/).

squidfunk:
Thanks for reporting. Everything is working correctly – you have marked your posts with [`draft`](https://squidfunk.github.io/mkdocs-material/plugins/blog/#meta.draft), which are not rendered when the site is built, except for when you set [`draft`](https://squidfunk.github.io/mkdocs-material/plugins/blog/#config.draft) in `mkdocs.yml`:

``` yml
plugins:
  - blog:
      draft: true
```

[...]

PR #7102, opened 2024-04-26
added note that blog posts should not go into nav

alexvoss:
this was in response to
https://github.com/squidfunk/mkdocs-material/issues/6675#issuecomment-1905271171

### DIFF
--- a/docs/setup/setting-up-a-blog.md
+++ b/docs/setup/setting-up-a-blog.md
@@ -33,7 +33,7 @@ plugins:
 ```
 
 If you do not have a navigation (`nav`) definition in your `mkdocs.yml` then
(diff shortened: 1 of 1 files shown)


docs/setup/setting-up-a-blog.md
```diff
@@ -32,6 +32,19 @@ plugins:
   - blog
 ```
 
+If you do not have a navigation (`nav`) definition in your `mkdocs.yml` then
+there is nothing else to do there as the Blog plugin will add navigation
```
squidfunk, line 36: "Blog plugin" -> "blog plugin".

squidfunk (changes requested):

squidfunk:
LGTM.

Issue #4678, opened 2022-11-25
Problem loading assets in mkdocs deployed on github pages


danimrprofe:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the problem doesn't occur with the `mkdocs` or `readthedocs` themes
- [X] ... the problem persists when all overrides are removed, i.e. `custom_dir`, `extra_javascript` and `extra_css`
- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

Hello, don't know if this is the right place to ask for help, but i'm having a little problem and can't find the solution to it. By the way, love mkdocs and the material theme and I would appreciate if you could help me.

Mkdocs works perfect when served locally (localhost), but displays without styles or images when deployed to github.

Thank you in advance, if more information is needed please let me now.

### Expected behaviour

I installed mkdocs, initiated a new project in a repo and configurated the yaml as the guide said. mkdocs serves works perfect as in local server everything displays as expected. Then I deploy with the command with the option gh-deploy on mkdocs but on github doesn't load assets images or styles, only the HTML.

### Actual behaviour

This is where the build shoud deploy the website. I did it from command line via mkdocs command, and it pushed to the new branch of the repository (gh-deploy)

https://danimrprofe.github.io/apuntes/

As you can see no styles ar shown:

![image](https://user-images.githubusercontent.com/29753978/204015759-f0b1dc91-df38-4d13-bcb6-21f773f15f68.png)

This is how the website is displayed (without styles). Seems to be looking at the media in the root of the URL, where they are not. I mean, the HTML looks for:

https://danimrprofe.github.io/assets/images/favicon.png

Instead of where it really is:

https://danimrprofe.github.io/apuntes/assets/images/favicon.png

My pages options in the repo.

![image](https://user-images.githubusercontent.com/29753978/204015591-a9e185d6-fbd1-4207-a378-775cfcd5ea6f.png)

### Steps to reproduce

1. Build with mkdocs
2. Deploy with gh-deploy
3. Check in the github.io page.

### Package versions

- Python: Python 3.8.2
- MkDocs:  mkdocs, version 1.4.2
- Material: Version: 8.5.10

### Configuration

```yaml
site_name: My Docs

theme:
  name: material
  language: es
  features:
[...]

PR #7103, opened 2024-04-26
added note on importance of site_url

alexvoss:
this was in response to
https://github.com/squidfunk/mkdocs-material/issues/4678#issuecomment-1327681673

### DIFF
--- a/docs/creating-your-site.md
+++ b/docs/creating-your-site.md
@@ -44,6 +44,22 @@ theme:
   name: material
 ```
 
+Note that by default, MkDocs will assume that your site is hosted at the root of
+your domain. If that is not the case then you also need to add a `site_url`
+setting:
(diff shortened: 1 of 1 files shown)


docs/creating-your-site.md
```diff
@@ -44,6 +44,22 @@ theme:
   name: material
 ```
 
+Note that by default, MkDocs will assume that your site is hosted at the root of
```
squidfunk, line 47: Great work! We should add that the `site_url` setting is strongly recommended from our perspective, as without it, several features will not work, including instant loading and previews, as well as social cards.

squidfunk (changes requested):

squidfunk (approved):
Thanks, LGTM.

Issue #7125, opened 2024-04-29
Local `.git` directory is embedded in the zip file generated by `info`


zx80:
### Context

When generating a bug report, more things than desired are leaked out.

### Bug description

The zip file generated by the `info` plugin include the local `.git` history.

This **must** be filtered out, because it has potential security implications.

### Related links

- [Info](https://squidfunk.github.io/mkdocs-material/plugins/info/)

### Reproduction

[9.5.19-git-directory-in-zip.zip](https://github.com/squidfunk/mkdocs-material/files/15146025/9.5.19-git-directory-in-zip.zip)

### Steps to reproduce

Open zip generated by info and acknowledge that it contains the `.git` directory.

### Before submitting

- [X] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
- [X] I have attached links to [the documentation](https://squidfunk.github.io/mkdocs-material/), and possibly related [issues](https://github.com/squidfunk/mkdocs-material/issues) and [discussions](https://github.com/squidfunk/mkdocs-material/discussions).
- [X] I assure that I have [removed all customizations](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/#remove-customizations) before submitting this bug report.
- [X] I have attached a __.zip file__ with a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/) using the [built-in info plugin](https://squidfunk.github.io/mkdocs-material/plugins/info/).

squidfunk:
Thanks for reporting. Note that this runs counter to the idea of a minimal reproduction – you should not pack up your project, but create a new one which only includes the parts that show an issue. Other than that, @kamilkrzyskow, should we add `.git` to the list of exclusion patterns?

kamilkrzyskow:
Sure, I believe a workflow where people prefer to delete stuff from their project on a branch to later simply `git restore` if a file needs restoration is worth supporting.

I'm a bit ignorant about what security sensitive info can be stored in the .git cache, perhaps names of branches so more of a privacy matter. Private keys should not be stored there or should they 🤔 

The exclusion pattern should likely also include inner subprojects as they could also have .git directories.
The patterns always are resolved from the beginning so `/[^/]*\.git/` should do the trick.

squidfunk:
[...]

PR #7142, opened 2024-05-01
Add warning for dotfiles in info plugin

kamilkrzyskow:
Attempt at handling #7125 

I admit the PR is a bit "noisy", for such a small feature, but I think this is the best I can do, given that there are 2 separate `for` loops for directory/files paths.
There are 2 warnings, during the processing phase and later in the summary. I want to make sure people see it.
Perhaps the display of the list at the end is unnecessary and not so convenient, as I think it is?

```
Please name your bug report (2-4 words): gmc
WARNING -  The following .dotpath will be included: C:\MyFiles\Gothic Modding\git\gmc\.git
WARNING -  The following .dotpath will be included: C:\MyFiles\Gothic Modding\git\gmc\.github
WARNING -  The following .dotpath will be included: C:\MyFiles\Gothic Modding\git\gmc\.gitignore
Processing: C:\MyFiles\Gothic Modding\git\gmc\.git\objects\pack
```

```
List of potentially sensitive files:
  - C:\MyFiles\Gothic Modding\git\gmc\.git
  - C:\MyFiles\Gothic Modding\git\gmc\.github
  - C:\MyFiles\Gothic Modding\git\gmc\.gitignore
WARNING -  Archive contains potentially sensitive dotfiles.
           Please review and share only necessary data to reproduce the issue.
```

EDIT:
I will try to modify the actual summary file list and change color of the dotfiles. This should remove the need for another list.

### DIFF
--- a/material/plugins/info/plugin.py
+++ b/material/plugins/info/plugin.py
@@ -212,6 +212,9 @@ class InfoPlugin(BasePlugin[InfoConfig]):
                 pattern = _resolve_pattern(project_config.site_dir)
                 self.exclusion_patterns.append(pattern)
 
+        # Track dotpath inclusion to inform about it later
+        contains_dotpath: bool = False
+
         # Create self-contained example from project
         files: list[str] = []
         with ZipFile(archive, "a", ZIP_DEFLATED, False) as f:
@@ -228,6 +231,11 @@ class InfoPlugin(BasePlugin[InfoConfig]):
                     # Exclude the directory and all subdirectories
                     if self._is_excluded(path):
                         dirnames.remove(name)
+                        continue
+
+                    # Warn about .dotdirectories
+                    if _is_dotpath(path, log_warning = True):
+                        contains_dotpath = True
 
                 # Write files to the in-memory archive
                 for name in filenames:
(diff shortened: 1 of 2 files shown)


squidfunk:
PR looks good so far, thanks! This also means that a `.meta.yml` will trigger a warning, right? I guess this is okay, let's see how users will react to it.

kamilkrzyskow:
I removed the second list, and instead changed the color of the dotpaths. This has lead to creating a second list anyway, because the file names are sorted, and the yellow color has a bigger ANSI value, therefore all dotpaths are ordered last. I also removed some conditionals, when clearing the indicator, because it turned out to be unnecessary.

I wanted to print a count of the dotpaths, but the results highlight also the contents of dotdirectories, so the count wouldn't match from what was initially detected, to keep it simple I don't count anything now. 

I think it's cleaner than the initial PR, but let me know what you think :v:

squidfunk (approved):
Thanks! I guess we can give it a spin and see if reports for `.meta.yml` and other non-critical files lead to confusion. The info plugin is something we should try different approaches on until we find the one that works best ☺️

squidfunk:
@kamilkrzyskow you can merge it if it's ready to go 🚀

kamilkrzyskow:
Okey dokey 👌

PR #7150, opened 2024-05-03
Remove erroneous *.js pattern for extra_javascript in schema.json

vorburger:
Since https://github.com/mkdocs/mkdocs/pull/92 for https://github.com/mkdocs/mkdocs/issues/91, an extra_javascript item does not necessarily have to end in *.js; e.g. in https://github.com/enola-dev/enola/pull/669 I have an `extra_javascript: - https://unpkg.com/mustache@latest`, which this flags up as wrong - although it's not (it works great); ergo it's better to remove this constraint.

### DIFF
--- a/docs/schema.json
+++ b/docs/schema.json
@@ -96,9 +96,8 @@
       "markdownDescription": "https://squidfunk.github.io/mkdocs-material/customization/#additional-javascript",
       "type": "array",
       "items": {
-        "title": "Path to JavaScript file",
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. I would not call it erroneous, since the URL you use is not the final resolved URL – it's redirecting to the final URL which contains the `.js` extension. However, I agree that we can remove that restriction.

PR #7152, opened 2024-05-05
Fix JSON parse error in schema.json

mschoettle:
It looks like the recently merged PR #7150  accidentally caused a JSON parse error in the `schema.json`.

I noticed it because I use the YAML extension in vscode and it failed validating my `mkdocs.yml` with the following error:

> Unable to parse content from 'https://squidfunk.github.io/mkdocs-material/schema.json': Parse error at offset 4055.YAML(768)


squidfunk:
Thanks! I think that goes on my pile of shame 😅

PR #7163, opened 2024-05-07
Rewritten Cairo and pngquant installation instructions for Windows

vedranmiletic:
Instead of installing GTK+ and installing from source, suggested
installing MSYS2 and obtaining Cairo built for UCRT64 via its Pacman.

### DIFF
--- a/docs/plugins/requirements/image-processing.md
+++ b/docs/plugins/requirements/image-processing.md
@@ -52,9 +52,17 @@ Material for MkDocs makes use of for generating [social cards] and performing
 
 === ":fontawesome-brands-windows: Windows"
 
-    As stated in the [installation guide], the easiest way to get up and running
-    with the [Cairo Graphics] library on Windows is by installing [GTK+]. You
-    can also download a precompiled [GTK runtime].
+    The easiest way to get up and running with the [Cairo Graphics] library is
+    by installing it via [MSYS2], which is a software distribution and building
+    platform for Windows. Run the following command inside of a MSYS2 shell:
+
+    ```
+    pacman -S mingw-w64-ucrt-x86_64-cairo
+    ```
+
+    MSYS2 provides the Cairo Graphics library in several different environments.
+    The above command uses the [UCRT64] environment, as recommended by the MSYS2
+    developers.
 
 === ":material-linux: Linux"
 
@@ -88,8 +96,8 @@ The following environments come with a preinstalled version of [Cairo Graphics]:
   [Cairo Graphics]: https://www.cairographics.org/
   [Homebrew]: https://brew.sh/
   [installation guide]: https://www.cairographics.org/download/
-  [GTK+]: https://www.gtk.org/docs/installations/windows/
-  [GTK runtime]: https://github.com/tschoonj/GTK-for-Windows-Runtime-Environment-Installer/releases
+  [MSYS2]: https://www.msys2.org/
+  [UCRT64]: https://www.msys2.org/docs/environments/
   [Docker image]: https://hub.docker.com/r/squidfunk/mkdocs-material/
   [GitHub Actions]: ../../publishing-your-site.md#with-github-actions
 
@@ -111,9 +119,13 @@ explains how to install [pngquant] system:
 
 === ":fontawesome-brands-windows: Windows"
 
-    Installing [pngquant] on Windows is a little more involved. The
-    [pngquant-winbuild] repository contains a guide on how to set up an
-    environment for building [pngquant] on Windows.
+    The easiest way to get [pngquant] is by installing it via [MSYS2], which is
+    a software distribution and building platform for Windows. Run the following
+    command inside of a MSYS2 shell:
+
+    ```
+    pacman -S mingw-w64-ucrt-x86_64-pngquant
+    ```
 
 === ":material-linux: Linux"
 
@@ -190,19 +202,19 @@ process is different:
     environmental `PATH` variable. Additionally each library `name` is checked
     in [two variants][find-library-Windows] with the `name` and `name.dll` format.
 
-    The default installation path of [GTK runtime] is:
+    The default binary and shared library path for the [UCRT64] environment of
+    [MSYS2], in which the packages were installed using the above commands, is:
 
     ```powershell
-    C:\Program Files\GTK3-Runtime Win64
+    C:\msys64\ucrt64\bin
     ```
 
-    and the libraries are in the `<INSTALL-DIR>\lib` directory. Use the debug
-    script below to check if the path is included. If it isn't then:
+    Use the debug script below to check if the path is included. If it isn't then:
 
     1. Press ++windows+r++.
     2. Run the `SystemPropertiesAdvanced` applet.
     3. Select "Environmental Variables" at the bottom.
-    4. Add the whole path to the `lib` directory to your `Path` variable.
+    4. Add the whole path to the above directory to your `Path` variable.
     5. Click OK on all open windows to apply changes.
     6. Fully restart any open Terminal windows and their parent hosts like IDEs.
 


squidfunk:
Thanks for the PR! I can't review this, my house has no Windows. @alexvoss?

alexvoss:
I will test it once back home. Got a Win 11 VM and a Win 10 laptop, so some diversity.

docs/plugins/requirements/image-processing.md
```diff
@@ -190,14 +194,13 @@ process is different:
...
-    C:\Program Files\GTK3-Runtime Win64
+    C:\msys64\mingw64\bin
     ```
 
-    and the libraries are in the `<INSTALL-DIR>\lib` directory. Use the debug
-    script below to check if the path is included. If it isn't then:
+    Use the debug script below to check if the path is included. If it isn't then:
 
     1. Press ++windows+r++.
     2. Run the `SystemPropertiesAdvanced` applet.
```
kamilkrzyskow, line 206: Are libs installed via pacman available outside of the MSYS2 terminal? 
Also if they should be, then some path is added to the PATH variable, are those instructions still accurate?

vedranmiletic: Shared libraries are in `\bin` directory, which is noted in the changed lines just above this one; `\lib` directory contains only static libraries, which should not be in `PATH`, as far as I understand.

alexvoss:
Can I make a suggestion - if only so someone can shoot it down? 

We should make WSL, Docker (which is also WSL), and MSYS2 the three officially supported ways of getting the `libcairo` and `pngquant` dependencies on Windows. 

MSYS2 is for all those poor souls who cannot get WSL in their corporate environment, or those who actually like to work in Powershell. Installation is pretty easy to describe for us as we can rely on the installation process of MSYS2 and simply provide the `pacman` instructions. For those folks who installed a Windows version of Pyhton we also will need to document that extra environment variable.

The installation instructions for `pngquant`effectively say to compile from source in a MinGW environment, which would most likely be MSYS2 anyway? So, by telling people to just install with `pacman` we can simplify that a lot and provide all that is needed from our own documentation.

Whatcha think?

vedranmiletic:
> Can I make a suggestion - if only so someone can shoot it down?
> 
> We should make WSL, Docker (which is also WSL), and MSYS2 the three officially supported ways of getting the `libcairo` and `pngquant` dependencies on Windows.
> 
> MSYS2 is for all those poor souls who cannot get WSL in their corporate environment, or those who actually like to work in Powershell. Installation is pretty easy to describe for us as we can rely on the installation process of MSYS2 and simply provide the `pacman` instructions. For those folks who installed a Windows version of Pyhton we also will need to document that extra environment variable.
> 
> The installation instructions for `pngquant`effectively say to compile from source in a MinGW environment, which would most likely be MSYS2 anyway? So, by telling people to just install with `pacman` we can simplify that a lot and provide all that is needed from our own documentation.

Yes. To be fair, MinGW(-w64) is just one of the provided [environments](https://www.msys2.org/docs/environments/) in MSYS2, but it is likely the most commonly used one. We can choose one of the other environments as well if it works well.

Regarding pngquant, there is a package for it in MSYS2 too: https://packages.msys2.org/package/mingw-w64-x86_64-pngquant?repo=mingw64

> Whatcha think?

Sounds good to me.

alexvoss:
> it is likely the most commonly used one

Reading [this StackOverflow post](https://stackoverflow.com/questions/76552264/what-are-msys2-environments-how-do-i-pick-one), it seems the difference is between UCRT64 and MINGW64 is that the latter supports version of Windows older than 10? We could see if UCRT64 works. Since it is the default shell for a new installation, it might be good to go with it?

alexvoss:
I have installed both libraries in UCRT64 and that worked fine. Unless we want to support Windows versions < 10, I would say making UCRT64 the default seems to be the way to go. It is a bit of a pain that pacman needs the precise package name including the specification of the environment, so the prefix would be `mingw-w64-ucrt-x86_64-`, which is a bit of a mouth- erm, handful. There are [solutions for this](https://www.msys2.org/docs/package-naming/), of which the use of the `MINGW_PACKAGE_PREFIX` environment variable seems to me to be the most sensible for our purposes.

vedranmiletic:
> > it is likely the most commonly used one
> 
> Reading [this StackOverflow post](https://stackoverflow.com/questions/76552264/what-are-msys2-environments-how-do-i-pick-one), it seems the difference is between UCRT64 and MINGW64 is that the latter supports version of Windows older than 10? We could see if UCRT64 works. Since it is the default shell for a new installation, it might be good to go with it?

MSYS2 anyhow requires [Windows 10 64-bit or newer](https://www.msys2.org/), so I agree.

> I have installed both libraries in UCRT64 and that worked fine. Unless we want to support Windows versions < 10, I would say making UCRT64 the default seems to be the way to go. It is a bit of a pain that pacman needs the precise package name including the specification of the environment, so the prefix would be `mingw-w64-ucrt-x86_64-`, which is a bit of a mouth- erm, handful. There are [solutions for this](https://www.msys2.org/docs/package-naming/), of which the use of the `MINGW_PACKAGE_PREFIX` environment variable seems to me to be the most sensible for our purposes.

I tested it independently in WinDev VM and confirmed it works as well. Current state of the pull request uses Pacman directly, but I can also rewrite it to use Pacboy.

docs/plugins/requirements/image-processing.md
kamilkrzyskow: So the 4th point in list should point to the `bin` and not `lib` directory, right?

kamilkrzyskow:
> We should make WSL, Docker (which is also WSL), and MSYS2 the three officially supported ways of getting the libcairo and pngquant dependencies on Windows.

I don't know, I'm personally not a fan of Docker or WSL. I once installed it and the virtualization in my system changed and broke all of my VMware/VirtualBox VMs, as in, they wouldn't boot anymore. I didn't have any data on them, so I let it slide, but overall I consider Docker being too "invasive". Perhaps I just have a different ideology, as I prefer installing things manually instead of using reusable images. However, I do realize this is the standard nowadays, for developers.

Now would Docker and WSL be another hurdle or would they actually simplify things, for the generic user, non-developer, which probably installed Python through the Microsoft Store (instead of doing it the "proper" manual way via the installer executable from the Python website)? 🤔 

MkDocs does require running a Terminal command, and therefore by default we require a basic willingness to use the Terminal. However, I believe that most Windows users are used to clicking stuff, like the GTK Runtime installer, or modifying PATH via the window applet.

I feel like MSYS2 is a good middle-ground between pure Windows and WSL as it adds direct options for installing libraries, without the need for a new system in a VM. However, it increases the amount of commands the user has to write to configure their setup, so it's difficult to outright say it will be better for the typical Windows non-developer.

<details>
<summary> Now I understand the backstory of this meme a bit more </summary>



</details>

vedranmiletic:
> > 
> 
> That is also fine with me. I don't think it is here or there whether you add an extra install to then install Cairo and `pngquant` more easily.
> 
> Are you going to work on a new version along these lines? I am traveling tomorrow, so will be offline but happy to look at anything once I am back connected. Will be shlepping the Windows laptop with me.

Honestly, I would prefer 2 commands instead of 3. Users who are familar with MSYS2 will understand the broad idea and know how to use Pacboy themselves. I would keep it as is.

> MkDocs does require running a Terminal command, and therefore by default we require a basic willingness to use the Terminal. However, I believe that most Windows users are used to clicking stuff, like the GTK Runtime installer, or modifying PATH via the window applet.

I see where you are coming from and installers are indeed nice. On the other hand, I think we should recommend maintained packages (like those in MSYS2) over some builds for Windows made by some enthusiasts in 2022 and not kept up to date with upstream software development afterwards  (which is better than not having them at all, of course, and I respect that).

docs/plugins/requirements/image-processing.md
vedranmiletic: Good catch. Let me clear that up.

alexvoss:
> Now would Docker and WSL be another hurdle

@kamilkrzyskow my suggestion was based on the assumption that they are options already in use by many people who are technically-minded but work in a Windows shop. Docker is based on WSL, so should not get in the way of anything happening on the Windows side of things? Correct me if I am mistaken.

MSYS2 is an alternative for those who cannot use WSL or are used to Powershell. It replaces installing GTK for Cairo. Note that the [installation instructions currently linked](https://www.gtk.org/docs/installations/windows/) to mention installing GTK via MSYS2, so not much gained there. The [GTK runtime pointed](https://github.com/tschoonj/GTK-for-Windows-Runtime-Environment-Installer/releases) to is also based on MSYS2, except the result has been bundled up in an installer. Some may well prefer that. 

For pngquant, the effective instructions are to [compile it yourself](https://github.com/jibsen/pngquant-winbuild), presumably using MSYS2, so just using `pacman` for that is a decent step forward.

kamilkrzyskow:
> I think we should recommend maintained packages (like those in MSYS2) over some builds for Windows made by some enthusiasts in 2022 and not kept up to date with upstream software development afterwards (which is better than not having them at all, of course, and I respect that).

True, cairo has a newer version [`1.18.0`](https://cairographics.org/news/) and the runtime build has 1.17.4, so best to keep the software updated for every user to avoid edge cases.

> my suggestion was based on the assumption that they are options already in use by many people who are technically-minded but work in a Windows shop

Your assumption might be correct, I've seen various non-developers (on GitHub issues/discussions) that still use Docker due to the simplicity of running prepared dockerfiles, I can see myself being the outlier with an odd perspective 😅 Depending on the system, enabling WSL with Docker might not be the easiest thing to do (as virtualization might be disabled in the BIOS), so I wouldn't consider WSL as the best candidate for "official" ways to do things.

> Docker is based on WSL, so should not get in the way of anything happening on the Windows side of things? Correct me if I am mistaken.

Yes, as far as I know Docker runs on top of WSL2 with Hyper-V, but I'm not sure what you meant with the "get in the way" or "Windows side of things". Containers are similar to a sandbox environment, the point of them is to isolate software, so I would assume nothing should conflict with them, apart from other virtualization solutions that don't use Hyper-V (this happened in my case).

alexvoss:
> I wouldn't consider WSL as the best candidate for "official" ways to do things

Is this not taken care of with "if you are using WSL"? True, I would not want to write instructions for every BIOS on the plant for enabling virtualization (though this might one day become my own personal circle of hell, who knows). I mentioned Docker because that particular boat has sailed since it is [mentioned in the installation guide](https://squidfunk.github.io/mkdocs-material/getting-started/?h=docker#with-docker). I personally also never understood the allure of Docker for something like this but people have different views on such matters. 

For me, the main question is more about how easy or simple the instructions are once you have chosen and prepared a deployment option. I would assume that someone who is happy to use WSL will also be able to make it run on their machine. Conflicting hypervisors is an unfortunate reality, are not new and not unique to Windows.

vedranmiletic:
Please let me know if I can make further improvements to the wording of the changes.

For what its worth, while preparing a [recent "build your own website with MkDocs" workshop](https://group.miletic.net/hr/prezentacije/2024-04-24-radionica-izradite-svoj-web-u-4-sata/), I found out there is now a [PowerToys utilty for configuring Environment Variables](https://learn.microsoft.com/en-us/windows/powertoys/#environment-variables). It could be a modern replacement for the current one that will eventually make it to official Windows via a future update. While it looks far nicer than the Windows built-in one, I would for the time being stick with the one that everyone has installed.

squidfunk:
Once @alexvoss finds some time to review this, I'm happy to merge it ☺️ As said, not a Windows user, so currently hard to test for me. I really need to get my hands on a notebook for testing at some point.

kamilkrzyskow (approved):

kamilkrzyskow:
Wording is fine by me, also Alex has tested the MSYS2 installation.
I think the issue got stale, because you mentioned >but I can also rewrite it to use Pacboy
and Alex sort of gave an okay for that 🤔 

> I would for the time being stick with the one that everyone has installed.

I'm for sticking to the old Windows XP applet approach until the end of life for Windows 10 aka 2028 🤓

squidfunk:
Perfect, thanks! LGTM. I didn't mention Pacboy (no idea what that is), @vedranmiletic did. 

> I'm for sticking to the old Windows XP applet approach until the end of life for Windows 10 aka 2028 🤓

Sounds sensible.

alexvoss:
@vedranmiletic @kamilkrzyskow are you both using the Python from MSYS2? The one case that is not covered, was far as I can see, is where people install Python from python.org before they come across MSYS2. In that case they will use the recently introduced [CAIROCFFI_DLL_DIRECTORIES environment variable](https://doc.courtbouillon.org/cairocffi/stable/overview.html#importing) or need to make sure they are always using the version from MSYS2, which would likely be the case within an MSYS2 environment but not in a Powershell.

kamilkrzyskow:
Uh oh, did I jump the gun again 👀 💦 

> I didn't mention Pacboy

Yes, I was answering to Vedran, but didn't adjust my message after you posted yours Martin, so the ordering did not work out 😅 

> are you both using the Python from MSYS2?

I'm using Python installed from python.org, but I'm also not using MSYS2/Powershell directly. I'm using Git for Windows, which added Bash, and Git Bash is basically a Terminal like `mintty` together with `MINGW64`, so basically the same as MSYS2, but I don't have `pacman` 😆Windows PATH works as expected, so if the current approach does add a directory to PATH like the previous runtime installation did, then everything should continue to work ✌️

alexvoss:
@kamilkrzyskow, am I reading your message right that in the end you are using a MingW64 version of Python? My experience with Python 3.12 from python.org is that I *do* need to set that extra environment variable. When I use Python from a MingW64 environment (MSYS2 in my case), I do not need this. My understanding is that this is because a MingW64 Python is compiled as if for a Unix-type environment and so does not contain the security patch that keeps the Windows version from searching in the PATH. This understanding seems consistent with the cairocffi documentation.

kamilkrzyskow:
I'm not sure, as it doesn't say GCC/MINGW64, which is what I would expect:
```py
sys.version
3.12.0 (tags/v3.12.0:0fb18b0, Oct  2 2023, 13:03:39) [MSC v.1935 64 bit (AMD64)]
```
I can test MSYS2 on my machine if it's necessary to get to the bottom of this 🤔

alexvoss:
For comparison, this is what I get in Powershell:

```
PS C:\src\tmp> gcm python

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     python.exe                                         0.0.0.0    C:\Users\Alex Voss\AppData\Local\Microsoft\WindowsApps\python.exe
PS C:\src\tmp> python
Python 3.12.3 (tags/v3.12.3:f6650f9, Apr  9 2024, 14:05:25) [MSC v.1938 64 bit (AMD64)] on win32
...
```

Could it be that the difference is that Python is installed under my account?

alexvoss:
So, I suspected that the fact I had you have the DLL in the System path as opposed to the user path was the crucial difference but it did not make any difference on my system. The odd thing is that we are trying to find out *why it works on your system* rather than why it does not on mine. Gotta go and get my daughter from school. I suggest we leave it be until I find new information or someone else gets in touch who can help.

vedranmiletic:
> @vedranmiletic @kamilkrzyskow are you both using the Python from MSYS2? The one case that is not covered, was far as I can see, is where people install Python from python.org before they come across MSYS2. In that case they will use the recently introduced [CAIROCFFI_DLL_DIRECTORIES environment variable](https://doc.courtbouillon.org/cairocffi/stable/overview.html#importing) or need to make sure they are always using the version from MSYS2, which would likely be the case within an MSYS2 environment but not in a Powershell.

From my experience with Python 3.11 installed in MSYS2, it just works without that variable. On the workshop we used Python 3.12, but did not build social cards, so I can't say. (That's my primary driver here: I want to show how good the social card look and how easy they are to build when we have that workshop again in a few months).

PR #7178, opened 2024-05-13
Update count in `philosophy.md`

baseplate-admin:
Clarify that more than 40,000 projects are using mkdocs-material now


squidfunk:
Thanks! Sure, why not ✌️

PR #7203, opened 2024-05-21
Added --force-reinstall to insiders upgrade instructions

alexvoss:
As discussed, added `--force-reinstall` to the upgrade instructions for people who need the latest development version. This avoids the problem that `pip` might decide there is nothing to do based on comparing version numbers.


alexvoss:
Oh, and: I felt it was warranted to give an explanation as to why the flag is there.

squidfunk:
LGTM, thanks!

Issue #1801, opened 2020-07-07
🚀 The future of Material for MkDocs


squidfunk:
___tl;dr__: Material for MkDocs will switch to a [Sponsorware](https://github.com/sponsorware/docs) model. New features will first be released exclusively to Sponsors. When a certain Sponsoring goal is reached, these features will be released to the public and be generally available!_

<a href="https://github.com/sponsors/squidfunk">
<img alt="Sponsor this project" src="https://user-images.githubusercontent.com/932156/86788364-601e0a00-c066-11ea-804b-5faef7d5401a.png" width="100" />
</a>

---

Material for MkDocs has come a long way. Version 5 brought many improvements and a completely new architecture which makes extension and customization much easier. We haven't even begun to scratch the surface. After the dust settled and the majority of bugs introduced by the latest major release were fixed, I want to talk about the future.

### Where we are

Thanks to all of you who participated in the [user survey](https://twitter.com/squidfunk/status/1259832077668810752), we have a pretty good idea of what can be improved and what features are missing or could be better. Additionally, I have tons of ideas on how to improve this project, especially for large (and enterprise) documentation projects. With Material for MkDocs being [more popular than ever before](https://porter.io/github.com/squidfunk/mkdocs-material), the number of issues and support requests is increasing at a steady pace.

__As the original author, I'm doing 99% of maintenance and design work__. I want to ensure that issues get fixed quickly and that new useful features make their way into `master`, but I also have to somehow pay for my living. I've invested months and months of my spare time into this project. __I'm super excited to see the first sponsors joining__, supporting this project, but the collected amount is currently very far from the value of the time I'm putting in this project. With so many big companies using it for their entire front-facing documentation, I'm hoping for more support, to ensure the future of this project!

### Where we're going

This brings me to the announcement I want to make: Material for MkDocs switches to a [Sponsorware](https://github.com/sponsorware/docs) model! This means:

- [x] __New features will first be made available exclusively to sponsors__
- [x] __Sponsors will be able to propose, discuss and vote, which issues and/or features we'll tackle next__

Of course, some things won't change:

- [x] __Material for MkDocs will remain free and Open Source, forever__
- [x] __Bugfixes will be addressed as quickly as possible__

I'm currently reshaping the Sponsorship tiers to offer something for agency and companies, but from $10 a month, you will be the first to enjoy new features that will increase the value and uniqueness of your documentation project.

### What's up next

The following roadmap shows, at which monthly goals those features will be available to the public:

[...]

PR #7210, opened 2024-05-24
New Insiders section

katharinalisalin:
Created a new and detailed Insiders section.


alonsomoya:
I actually kind of miss the old view on what features are already in insider and the brilliant sponsorware approach taken with funding goals introduced in [🚀 The future of Material for MkDocs](https://github.com/squidfunk/mkdocs-material/issues/1801)
It really helped having a view on exciting features coming next upstream, and an actual motivation to become insider

squidfunk:
Thanks for your feedback! We're still sticking to the funding goal based approach – nothing changed. We just rearranged the information. It's now spread across multiple pages because the one single Insiders page became too big:

- [Our story + why sponsorware](https://squidfunk.github.io/mkdocs-material/insiders/why-sponsor-us/)
- [Features and funding goals](https://squidfunk.github.io/mkdocs-material/insiders/benefits/)

We tried to answer all of the questions that we received again and again, and this made this step necessary.

Issue #7226, opened 2024-05-31
Version switcher fails to stay on the same page when `canonical_version` is set in mike  


kevinh-hcl:
### Context

We are using `mkdocs-material` together with `mike` for our documentation. 

To improve SEO we set the `canonical_version` in `mike`. This sets the `rel="canonical"` `link` on every page and also leads to the `sitemap.xml` in each version using the canonical links (with `latest`). This seems to be the expected way of handling canonical URLs according to [Google Search Central](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls) and others.

### Bug description

When switching from a sub page of any version to a version that was built **with** the `canonical_version` you get redirected to the homepage of the version. The expected behaviour would be to stay on the same sub page, but in a different version.

From what I can see this is probably due to `mkdocs-material` using the sitemap of the version to determine if a sub page is available in that version. Since the sitemap always uses the `canonical_version` instead of the actual version, the page is not found and it falls back to the homepage. 

### Related links

**Related documentation**
- [Reporting a bug](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/)
- [Setting up versioning - Stay on the same page when switching versions](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/#stay-on-the-same-page-when-switching-versions)

**Pointer from a previous discussion**
- [kamilkrzyskow seems to have the right idea in this discussion about the `latest` in all `sitemap.xml`](https://github.com/squidfunk/mkdocs-material/discussions/6809#discussioncomment-8551776)

**Related `mike` documentation and code**
- [mike: configuration options](https://github.com/jimporter/mike?tab=readme-ov-file#configuration)
- [mike: building `site_url` from `canonical_version`](https://github.com/jimporter/mike/blob/master/mike/mkdocs_plugin.py#L46-L49)

**Reproduction in a fork from `mkdocs-material-example-versioning`**
- [Setting `canonical_version`](https://github.com/kevinh-hcl/mkdocs-material-example-versioning/commit/37d9188bea97c801623d041368f604c60546f376)
- [Github Pages Version `0.5` (with `canonical_version`).](https://kevinh-hcl.github.io/mkdocs-material-example-versioning/0.5/bar/) You can switch **FROM** this versions to others just fine
[...]

PR #7227, opened 2024-05-31
initial idea for canonical version logic

kevinh-hcl:
Initial quick draft of an idea regarding https://github.com/squidfunk/mkdocs-material/issues/7226. This is a first draft, so feel free to give any feedback on improvements or misuse of any of the mechanisms of mkdocs or the theme.


squidfunk:
Thanks for the PR. Could I ask you to simplify the control flow and scope the code added to a check for `canonical_version`? I'm not sure if this breaks anything, because versioning with mike is so hard to test, so I'd rather opt for an approach where we change behavior based on the presence of this flag without mixing with other code. It's currently too hard to follow with `checkPath` etc.

squidfunk:
Superseded by  #7559.

PR #7244, opened 2024-06-04
Add Hummingbot as sponsor =)

fengtality:
Hey Martin, I created this PR to add Hummingbot back as a bronze sponsor, since I noticed we weren't on the website. The image links in the PR need to be updated.

I've included our logo below. 
blacklogo_with_text_new


squidfunk:
Of course, happy to do so! Note that we should use your logo on white background, because GitHub can be used in dark mode (which many people do, including me), so your logo would look like this:

Bildschirm­foto 2024-06-06 um 10 19 01

I suggest that we use the one we already used a while ago:

test

Is this okay for you?

fengtality:
sure, that one works well - thanks!

squidfunk:
Great! Superseded by 0c54a880d.

PR #7303, opened 2024-06-28
fixed broken link to discussions

alexvoss:
Noticed that the link to the discussions on GitHub was broken, contained only the query parameter


squidfunk:
Thanks for the PR. However, I'm pretty sure this is correct:

https://github.com/squidfunk/mkdocs-material/blob/c8d013cd152fe18d193d646b1413af3c4bea2297/docs/contributing/reporting-a-bug.md?plain=1#L84-L86

It talks about the documentation site, not about discussions. The reporter should search the documentation for relevant sections to be included in a bug report. Thus, I don't think we should change this link.

alexvoss:
Oh, I am sorry, I confused the two sections.

PR #7337, opened 2024-07-08
dev environment: Add `npm run fix` to fix (some) esLint errors

ilyagr:
This is the most helpful command I have figured out so far.

Let me know if there is a better way to fix formatting errors, e.g. an auto-formatting tool you use. Currently, I do "Format Document" in VS Code that results in a different style of formatting (semicolons everywhere), and then I run `npm run fix` to change it to this project's style.

(In case you are curious, I'm trying to refactor the version selector code slightly so that we can have some unit tests, and thus more confidence when improving the "stay on the same page when switching versions" behavior. There is a draft in #7338 )


ilyagr:
(In case you are curious, I'm trying to refactor the version selector code slightly so that we can have some unit tests, and thus more confidence when improving the "stay on the same page when switching versions" behavior. I will hopefully share the PR for that soon.)

squidfunk:
Thanks for the PR! I guess, why not. I'm actually using the VS Code integration by now that automatically fixes on save.

ilyagr:
Thanks! It's much easier for me to use this than to figure out how to have VS Code configured differently for different projects.

squidfunk:
You're welcome. However, as mentioned in another comment, please _always_ first discuss fundamental changes with us. You're introducing a lot of very opinionated changes and we're not planning to merge all of those. After all, we need to make sure that we can still guarantee to maintain the project.

PR #7370, opened 2024-07-16
Fixed blog readtime calculation to ignore non-content text

### LINKED ISSUE #7367: Blog readtime includes inline SVG text content
### Context

I'm inlining SVG images to be able to use Material for MkDocs' CSS variables for more consistent image styles with the rest of the page.

---

Slightly off topic but just for completeness: Unlike in the below reproduction (as I want to keep it minimal), I actually wrap the inline SVG in

```html
<div>
  <template shadowrootmode="open">
    <svg ...>
       ...
    </svg>
  </template>
</div>
```

to avoid leakage of the CSS definitions into the document and potentially conflict with other inline SVGs' CSS definitions.

### Bug description

The blog readtime plugin extracts _any_ text data from the generated HTML page content including, e.g., CSS definitions in inline SVG images. This leads to a bad estimate of the read time, especially when, e.g., there are many CSS definitions (i.e. much "text").

A fix might involve skipping some tags such as `<svg>`, `<style>`, and `<script>` when gathering text data. I'd be happy to contribute a fix if you agree with the bug report and when we've converged on a solution proposal.

### Related links

- [Reporting a bug](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/)
- [Basic blogs](https://squidfunk.github.io/mkdocs-material/tutorials/blogs/basic/)

### Reproduction

[9.5.29-blog-readtime-inline-svg.zip](https://github.com/user-attachments/files/16245715/9.5.29-blog-readtime-inline-svg.zip)

I've edited `venv/lib/python3.12/site-packages/material/plugins/blog/readtime/__init__.py` to demonstrate the current behavior of the HTML parser:

```diff
     # Extract words from text and compute readtime in seconds
+    print("DEBUG: ", parser.text)
     words = len(re.split(r"\W+", "".join(parser.text)))
     seconds = ceil(words / words_per_minute * 60)
```

### Steps to reproduce

1. Unzip the reproduction.
2. Run `mkdocs serve`.
3. Observe the line

    ```
    DEBUG:  ['Readtime includes inline SVG text', '\n', '\n  ', '\n    ', '\n      ', '\n        <![CDATA[\n        .fill-red {\n          fill: red;\n        }\n        ]]>\n      ', '\n    ', '\n    ', '\n  ', '\n  ', 'Red SVG rectangle', '\n']
    ```

    in the terminal which shows the list of text data extracted by the HTML parser of the readtime plugin.

### Browser

_No response_

### Before submitting

- [X] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
- [X] I have attached links to [the documentation](https://squidfunk.github.io/mkdocs-material/), and possibly related [issues](https://github.com/squidfunk/mkdocs-material/issues) and [discussions](https://github.com/squidfunk/mkdocs-material/discussions).
- [X] I assure that I have [removed all customizations](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/#remove-customizations) before submitting this bug report.
- [X] I have attached a __.zip file__ with a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/) using the [built-in info plugin](https://squidfunk.github.io/mkdocs-material/plugins/info/).

sisp:
I've fixed the blog readtime calculation to ignore non-content text, i.e. any text nodes that are inside the following tags are omitted from calculating the readtime:

* `<object>`
* `<script>`
* `<style>`
* `<svg>`

The implementation is inspired by the [search plugin parser](https://github.com/squidfunk/mkdocs-material/blob/4f8081c268d31bf74d546e600cadd0cff7dc89e8/src/plugins/search/plugin.py#L371-L579).

Note that I've duplicated the `void` set from the search plugin. It might be cleaner to deduplicate it, but I wasn't sure where to best move it to import from. If you'd prefer deduplication, I'd appreciate your guidance on where to factor `void` out, @squidfunk.

Fixes #7367.

### DIFF
--- a/material/plugins/blog/readtime/parser.py
+++ b/material/plugins/blog/readtime/parser.py
@@ -20,6 +20,10 @@
 
 from html.parser import HTMLParser
 
+# TODO: Refactor the `void` set into a common module and import it from there
+# and not from the search plugin.
+from material.plugins.search.plugin import void
+
 # -----------------------------------------------------------------------------
 # Classes
 # -----------------------------------------------------------------------------
@@ -31,15 +35,40 @@ class ReadtimeParser(HTMLParser):
     def __init__(self):
         super().__init__(convert_charrefs = True)
 
+        # Tags to skip
+        self.skip = set([
+            "object",                  # Objects
+            "script",                  # Scripts
+            "style",                   # Styles
+            "svg"                      # SVGs
+        ])
+
+        # Current context
+        self.context = []
+
         # Keep track of text and images
         self.text   = []
         self.images = 0
 
-    # Collect images
+    # Called at the start of every HTML tag
     def handle_starttag(self, tag, attrs):
+        # Collect images
         if tag == "img":
             self.images += 1
 
-    # Collect text
+        # Ignore self-closing tags
+        if tag not in void:
+            # Add tag to context
+            self.context.append(tag)
+
+    # Called for the text contents of each tag
     def handle_data(self, data):
-        self.text.append(data)
+        # Collect text if not inside skip context
+        if not self.skip.intersection(self.context):
+            self.text.append(data)
+
+    # Called at the end of every HTML tag
+    def handle_endtag(self, tag):
+        if self.context and self.context[-1] == tag:
+            # Remove tag from context
+            self.context.pop()
--- a/src/plugins/blog/readtime/parser.py
+++ b/src/plugins/blog/readtime/parser.py
@@ -20,6 +20,10 @@
 
 from html.parser import HTMLParser
 
+# TODO: Refactor the `void` set into a common module and import it from there
+# and not from the search plugin.
+from material.plugins.search.plugin import void
+
 # -----------------------------------------------------------------------------
 # Classes
 # -----------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


squidfunk:
> Note that I've duplicated the void set from the search plugin. It might be cleaner to deduplicate it, but I wasn't sure where to best move it to import from. If you'd prefer deduplication, I'd appreciate your guidance on where to factor void out, @squidfunk.

There's no need to integrate the void tags at all, as for them, the data handler is not called. Void tags are tags without children, i.e., leaf components. Please remove the void handling logic, as it does not influence the result.

squidfunk:
Ah wait, no, we might need them to maintain the `context`. Can you please just import it from the search plugin for now and make a note in the code that we might refactor it at some point? Please don't do the refactoring now, we're heavily working behind the curtains right now and there will be fundamental changes to Material for MkDocs, which is why we currently try to keep things as stable and unchanging as possible.

sisp:
I've checked: The `_endtag` is not called for void tags. Now, the `void` set is imported from the search plugin as you suggested.

squidfunk:
Okay great, thanks for investigating. Note that this is an HTML4 parser, not HTML5, which we chose for speed and package size (as no further dependency is needed), so as I remember, this is why we need to check for void tags by ourselves. We'll check in the future if we can swap this out for an approach that directly operates on the HTML AST, which would be much more efficient, but for now, I think we can keep it.

squidfunk:
From what I can see, this looks great, ready to merge! Thanks for your time!

sisp:
Perfect, thanks for the helpful review and feedback! :pray:

PR #7373, opened 2024-07-17
Documentation of sane_lists extension (#7327)

joapuiib:
This PR adds documentation corresponding to the [Sane Lists extension](https://python-markdown.github.io/extensions/sane_lists/) included by default in Python-Markdown, but not mentioned in the documentation.

See: #7327

I've set the minim version for using this extension at `0.1.0`, but I'm really not sure if it's the correct one.

Also, I've enabled the extension in `mydocs.yml`:
```yml
markdown_extensions:
  - sane_lists
```

This extension modifies the default Markdown behavior (See [Sane Lists extension syntax](https://python-markdown.github.io/extensions/sane_lists/#syntax)). It may have changed the output in some other pages.

### DIFF
--- a/docs/reference/lists.md
+++ b/docs/reference/lists.md
@@ -11,13 +11,14 @@ are supported through extensions.
 
 ## Configuration
 
-This configuration enables the use of definition lists and tasks lists, which
+This configuration enables the use of definition lists, tasks lists and sane lists, which
 are both not part of the standard Markdown syntax. Add the following lines to
 `mkdocs.yml`:
 
 ``` yaml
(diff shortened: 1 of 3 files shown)


squidfunk:
Thanks for the PR! However note that we cannot possibly list all Markdown extensions. We're currently not considering merging this PR, because there _is_ a lists implementation, albeit it might not be the best one. However, listing it in our documentation means we provide official support for it, which we don't.

Before introducing new extensions to our documentation it's always a good idea to discuss this ☺️

PR #7420, opened 2024-08-04
docs: add documentation on how to integrate mkdocs-drawio with navigation.instant feature

tuunit:
# Motivation

Add documentation for how to integrate mkdocs-drawio with material instant loading feature.

Solution for issue: https://github.com/tuunit/mkdocs-drawio/issues/11

# Description
Solution proposed and tested is inspired by the documentation for the math libraries integration.

https://squidfunk.github.io/mkdocs-material/reference/math/

# Test

Local serve of doc changes:

image

### DIFF
--- a/docs/reference/drawio.md
+++ b/docs/reference/drawio.md
@@ -1,0 +1,60 @@
+---
+icon: material/shape-outline
+---
+
+# Drawio
+
+[Drawio] is a popular tool and library for creating diagrams and charts
+in browsers. This documentation site provides information on how to integrate
+the [mkdocs-drawio] MkDocs plugin with the [instant-loading] feature of Material.
+
+  [Drawio]: https://www.drawio.com/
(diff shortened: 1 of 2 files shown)


squidfunk:
Thanks for the PR! However, we can't possibly add documentation for all third party plugins. If we add documentation for a plugin, we commit to providing first-class support for it, which, given that there are 280+ plugins, is quite impossible for us. Thus, this should be added to the documentation of draw.io, which we consider a downstream project that integrates with Material for MkDocs instant navigation functionality.

I hope my reasoning makes sense, if not, don't hesitate for clarification ☺️

tuunit:
@squidfunk I was already expecting something of the sort. So now worries. I know from experience that maintaining a large project means saying no to a lot of suggestions and prioritizing maintainability of the tool itself but the docs as well.

Any thanks for the fast reply and nicely written response.

squidfunk:
Thanks for understanding the pecularities of maintaining such a large project ☺️

PR #7466, opened 2024-08-20
Add `edit_uri_template` to docs/schema.json

FredZinelli:
Adding `edit_uri_template` to the main `docs/schema.json` file.

---

I initially planned to add the `docs/schema/theme.json` changes here, hence the name of the branch, but I realized that, if I add the name of my theme to your schema, it will appear in the suggestions of all your users and you probably won't want that.
This would mean to make the following change:

```yaml
{
  "$schema": "https://json-schema.org/draft-07/schema",
  "title": "Theme configuration",
  "markdownDescription": "https://squidfunk.github.io/mkdocs-material/",
  "type": "object",
  "properties": {
    "name": {
      "title": "Theme name",
      "markdownDescription": "https://www.mkdocs.org/user-guide/configuration/#name",
      "oneOf": [
        {
          "const": "material"
        },
        {
          "const": "pyodide-mkdocs-theme"      # <<< Here
        },
        {
          "type": "null"
        }
      ],
      "default": "material"
    },
...
```

Another possibility could be to accept strings, maybe?
Or maybe keeping the default, but using a pattern to validate both without the user seeing the name of mine?

Or just no changes, I can manage the red line if you're not happy with any of these suggestions.

### DIFF
--- a/docs/schema.json
+++ b/docs/schema.json
@@ -44,6 +44,11 @@
       "markdownDescription": "https://www.mkdocs.org/user-guide/configuration/#edit_uri",
       "type": "string"
     },
+    "edit_uri_template": {
+      "title": "More flexible variant of edit_uri",
+      "markdownDescription": "https://www.mkdocs.org/user-guide/configuration/#edit_uri_template",
+      "type": "string"
+    },
     "copyright": {
       "title": "Copyright, used in footer",
       "markdownDescription": "https://squidfunk.github.io/mkdocs-material/setup/setting-up-the-footer/#copyright-notice",


squidfunk:
Thanks for the PR. LGTM!

> I initially planned to add the docs/schema/theme.json changes here, hence the name of the branch, but I realized that, if I add the name of my theme to your schema, it will appear in the suggestions of all your users and you probably won't want that.

Yes, I'm afraid we won't consider this for merging, not because we do not want other themes to be represented, but because we would start managing your (extended) theme settings in our repository, which we can't commit to 😅 A better idea is to build on top of our JSON metadata (possibly by linking our schemas, as we do with third-party schemas).

FredZinelli:
This sure makes sense.

I did some other tries on mys side:

```yaml

    "name": {
      "title": "Theme name",
      "markdownDescription": "https://www.mkdocs.org/user-guide/configuration/#name",
      "oneOf": [
        {
          "type": "string"
        },
        {
          "type": "null"
        }
      ],
      "default": "material"
    },
  ...
```

The suggestion stays there for the user, but it would be possible to extend the theme without getting the red line.
What do you think?

squidfunk:
Sorry, but we're not considering this. The reason is that this opens our scheme to other themes, which we cannot commit to maintain, as mentioned before. Note that the JSON schema we provided is specifically tailored to our project. I understand that you're building a derivative theme, and want to use the same schema, but this is not something we actively support.

squidfunk:
To expand on that: this project is used by tens of thousands of projects. We must be very careful and cautious of what we commit to and what surface we offer. I hope you understand this from a maintainer's point of view.

FredZinelli:
> I hope you understand this from a maintainer's point of view.

Definitely, don't worry! :)

PR #7467, opened 2024-08-20
Update material builtin plugins with patterns

FredZinelli:
WIP

The `patternProperties` idea was definitely better!

I'm not really fond of the pattern for the strings, though. Maybe two `const` declarations would be better?
I think I should have either used two `const`, or maybe added a default value (if possible. It's my first time building json schema, as you probably guessed...) with the patterns, because for now, the user will loose the suggestions.

Your opinion?

### DIFF
--- a/docs/schema/plugins/blog.json
+++ b/docs/schema/plugins/blog.json
@@ -4,12 +4,13 @@
   "oneOf": [
     {
       "markdownDescription": "https://squidfunk.github.io/mkdocs-material/plugins/blog/",
-      "const": "blog"
+      "type": "string",
+      "pattern": "^(material/)?blog$"
     },
     {
       "type": "object",
-      "properties": {
-        "blog": {
+      "patternProperties": {
+        "^(material/)?blog$": {
           "markdownDescription": "https://squidfunk.github.io/mkdocs-material/plugins/blog/",
           "type": "object",
           "properties": {
--- a/docs/schema/plugins/group.json
+++ b/docs/schema/plugins/group.json
@@ -4,8 +4,8 @@
   "oneOf": [
     {
       "type": "object",
-      "properties": {
-        "group": {
+      "patternProperties": {
+        "^(material/)?group$": {
           "markdownDescription": "https://squidfunk.github.io/mkdocs-material/plugins/group/",
           "type": "object",
           "properties": {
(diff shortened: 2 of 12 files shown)


FredZinelli:
So, in fact... It's the use of `patternProperties` that removes the plugin from the suggestions list. Maybe not the way to go, in that case?

squidfunk:
Thanks for the PR. I'm still on the fence whether we want to support the material/* syntax here, since it is not necessary, and now those suggestions are popping up in the IDE. It might lead to more questions on our side (we have many non-/low-tech users), as there are now two ways to achieve the same thing.

However, you can change your theme's entrypoint and list the Material plugins with your prefix, which doesn't mandate a change on our side, specifically here:

https://github.com/squidfunk/mkdocs-material/blob/47f7f631043b3d862a7dd88dc9f742b0e143c47b/pyproject.toml#L77C1-L85C60

Something like the following should work:

``` toml
[project.entry-points."mkdocs.plugins"]
"foobar/blog" = "material.plugins.blog.plugin:BlogPlugin"
...
```

If your theme's called `foobar`, just re-export all plugins with `foobar`. Closing, as this can be resolved on your side.

squidfunk:
By the way, the big upside is, that users can use a configuration for `material` with your theme without changes.

FredZinelli:
Yes, on second thoughts, it's probably better like this.

I'll check these entry points things. Thx for the suggestion.

:+1:

squidfunk:
There are a few mentions in the templates that you'd need to override.

- [`material/search`](https://github.com/search?q=repo%3Asquidfunk%2Fmkdocs-material+material%2F+language%3AHTML+material%search&type=code)
- [`material/tags`](https://github.com/search?q=repo%3Asquidfunk%2Fmkdocs-material+material%2F+language%3AHTML+material%tags&type=code)

Replace them with your theme name, and it should work. I'm sorry that this is so complicated, but to be honest, MkDocs doesn't make it particularly simple for themes to ship plugins.

FredZinelli:
Wow, so obvious (and not, at the same time... XD ). Works great, thanks!

I'll proceed based on this and provide some updated schema on my side. 
I'll also fork again to remove the plugin related to my theme: that wouldn't make any sense anymore to have it in your schema while the plugin cannot go without my theme, which will propose its own schema overrides...

Thanks a lot for your help!

squidfunk:
Glad I could help and we could conserve the steps for anyone trying to achieve the same 🚀

PR #7468, opened 2024-08-20
Remove pyodide_macros plugin's schema

FredZinelli:
As discussed [here](https://github.com/squidfunk/mkdocs-material/pull/7467#issuecomment-2299427455), this schema will be introduced in the theme's own schemas "overrides" and since the plugin cannot be used outside of my theme anyway, so this schema has no point in your repo.

(Keeping the other plugin which has some use within other context)

Thanks again for your time and help!


squidfunk:
LGTM!

PR #7469, opened 2024-08-21
Fix issue with author images not filling container height

uus-dev:
Without this some author avatars became squeezed if the avatar image didn't fill the container height.

| Before | After |
|--------|--------|
| image | image |


squidfunk:
Thanks for the PR. Please provide a minimal reproduction. The problem with `height: 100%` is that it will stretch non-rectangular images. We should use `height: auto` here, but that's already defined for those images. Please provide a reproduction with exact steps how to observe the issue, because I'm not seeing it.

squidfunk:
Closing as stale.

PR #7475, opened 2024-08-23
Update Mermaid to 11.*

manselmi:
Bump the major version of Mermaid from 10 to 11.

Supported diagram types (flowcharts, sequence diagrams, state diagrams, class diagrams and entity-relationship diagrams) all seem to render the same as the ones [here](https://squidfunk.github.io/mkdocs-material/reference/diagrams/), in both dark and light modes.


src/templates/assets/javascripts/components/content/mermaid/index.css
```diff
@@ -68,7 +68,7 @@ marker {
 
 /* Flowchart edge label */
 .edgeLabel,
-.edgeLabel rect {
+.edgeLabel p {
```
manselmi, line 71: I made this change in order for the edge labels in [this flowchart](https://squidfunk.github.io/mkdocs-material/reference/diagrams/#using-flowcharts) to have their background look the same as the diagram backgroud (in both dark and light modes).  I know little about CSS and HTML - I'm not sure if this change breaks anything outside of my limited testing.

squidfunk: Yes, I think we might need this change, or edge labels are not readable in dark mode. It looks like they changed the `rect` to `p`, so this change should be okay. We'll find out as we go whether we need more changes.

squidfunk:
Thanks for the PR! LGTM.

PR #7486, opened 2024-08-29
Drop mkdocs-material-extensions from requirements

davide125:
This dependency isn't used anywhere anymore afaict


squidfunk:
Thanks for the PR. Yes, it is deprecated, but we must keep it for compatibility until we release the next major version. Otherwise, we'll break a lot of existing builds. Closing as unmergable.

PR #7489, opened 2024-08-30
Setup GitHub action

stollero:


stollero:
Sorry, should happen inside my fork! My bad!

PR #7525, opened 2024-09-14

jeremy-feng:
This commit replaces several outdated links in the documentation with updated URLs to ensure users have access to the latest and correct resources.


squidfunk:
Thanks! Good catch 🤟

PR #7527, opened 2024-09-15
link to subsection on adding an excerpt from blog plugin page

boegel:
I just spent way too much time to figure out how to create a blog post excerpt after finding a hint towards that feature at https://squidfunk.github.io/mkdocs-material/plugins/blog/#config.post_excerpt.

If there would have been a link in that section to the subsection on adding an excerpt at https://squidfunk.github.io/mkdocs-material/setup/setting-up-a-blog/#adding-an-excerpt, that would have helped me a lot...


squidfunk:
Thanks, sure, why not ☺️

PR #7572, opened 2024-09-29
Added gif and webp to the regex of the images patterns in theme schema

DaviAMSilva:
I believe the schema should allow the `.gif` and `.webp` formats for the images in the Logo and Favicon since they are widely used formats in the web.

For the Logo both of them should work just fine since they are standardized formats and commonly used for this purpose.

Aditionally, for the Favicon, according to [Wikipedia](https://en.wikipedia.org/wiki/Favicon#Image_file_format_support) and its sources, all major browsers have support to the `.gif` format, while Firefox and Opera even allows it to be animated.


squidfunk:
Using `webp` as favicon might not be the best idea, but sure, we can add it to the list of accepted extensions.

PR #7573, opened 2024-09-30
Updated some twitter links to X

Angxddeep:
Updated some Twitter.com links to X.com. Which just save a redirect,😉

### DIFF
--- a/material/overrides/main.html
+++ b/material/overrides/main.html
@@ -14,7 +14,7 @@
     <strong>Fosstodon</strong>
   </a>
   and
-  <a href="https://twitter.com/squidfunk">
+  <a href="https://x.com/squidfunk">
     <span class="twemoji twitter">
       {% include ".icons/fontawesome/brands/twitter.svg" %}
     </span>
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1178,7 +1178,7 @@ mkdocs-material-8.5.11+insiders-4.27.0 (2022-12-20)
 
 mkdocs-material-8.5.11 (2022-11-30)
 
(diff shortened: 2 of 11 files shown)


material/overrides/main.html
```diff
@@ -14,7 +14,7 @@
     <strong>Fosstodon</strong>
   </a>
   and
-  <a href="https://twitter.com/squidfunk">
+  <a href="https://x.com/squidfunk">
     <span class="twemoji twitter">
       {% include ".icons/fontawesome/brands/twitter.svg" %}
```
MGaetan89, line 19: Shouldn't the icon also be changed?
Same in `src/overrides/main.html`

Angxddeep: Oh, I was thinking of making multiple PRs for changing icons and links and text separately. Will it be OK if this is a massive PR as it will change almost every file TBH

MGaetan89: I'm not linked with `mkdocs-material`, I was just asking a question 🙂
@squidfunk will know better

Angxddeep: OK, I will do whichever way he is comfortable with

Angxddeep: I guess the Twemoji project is now dead, or something last commit was a year ago and their current URL doesn't work, so I will mark this conversation as resolved as nothing can be done about it now. :)

squidfunk: There is `x-twitter.svg`, but honestly, it doesn't matter as long as FontAwesome has `twitter.svg`.

squidfunk:
Thanks! LGTM.

PR #7581, opened 2024-10-03
Docs: add missing vscode tag config for slugify

yanyongyu:
slugify uses a [custom yaml tag](https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown-extensions/#+pymdownx.tabbed.slugify) to customize the slug function. This pr adds the missing config for VSCode in the docs.


squidfunk:
Thanks! I wasn't aware that you can specify the type after the tag, nice! Confirmed that it works now.

PR #7582, opened 2024-10-03
Add missing `anchors` field in top-level of validation config

yanyongyu:
The `anchors` field can also be set at the top-level of `validation`.

Ref: [mkdocs 1.6 config schema](https://json.schemastore.org/mkdocs-1.6.json)

Related to #7158


squidfunk:
Another great catch, thanks!

Issue #7545, opened 2024-09-20
Having two blog posts with matching title and date causes inconsistent navigation


KonstantinNosov:
### Context

Blog post uri is built from `title` and `date` while blog index rendering iterates posts so duplicated post is rendered twice on blog index while in navigation there is only on uri. 

From my perspective  having blog posts with matching title and date is valuable. Real world scenario - notifying users on hotfixes with blog and having two hotfixes within same day (it is cleaner to provide details in title and have  at least "hotfix 2.17.2" rather than "hotfix" in title but sometimes it is better to inform users fast rather than do it in proper way).

So I'd say it worth handling such cases with changing uri template from `[date]/[title]` to `[date]/[title]-[number]` if there are more than one post with given title an date e.g. for the attcahed zip uris may be `/blog/2024/09/20/bugfix-release-1/` and `/blog/2024/09/20/bugfix-release-2/`

### Bug description

If there are two `blog` posts with same title and date, blog index shows two items but both links point to the same post

### Related links

- [Reporting a bug](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/)
- [Blog](https://squidfunk.github.io/mkdocs-material/plugins/blog/#configuration)

### Reproduction

[blog issue.zip](https://github.com/user-attachments/files/17073327/blog.issue.zip)

### Steps to reproduce

1. enable `blog` plugin with `mkdocs.yml`
2. put two files with same title and date to blog/posts :

blogs/posts/post1.md
```
---
date:
  created: 2024-09-20
tags:
- notes
title: Bugfix Release
---


Bugfix release 2.17.1 is deployed
```

blogs/posts/post2.md
```
---
date:
  created: 2024-09-20
tags:
- notes
title: Bugfix Release
---


Bugfix release 2.17.2 is deployed
```

3. Launch `mkdocs serve`

### Actual
index page has two records both pointing post  from `blogs/posts/post2.md`

### Expected
one of:
- two blog records on blog index with correct urls for each post 
- one record on index and warning in log stating that one post is lost

### Browser

Chrome

### Before submitting

- [X] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
[...]

PR #7583, opened 2024-10-04
BlogPlugin: handling of generated blog posts

KonstantinNosov:
As it has been discussed in https://github.com/squidfunk/mkdocs-material/issues/7545 I've modified `blog` plugin code to handle generated posts. 

Now if file `Post` instance is being created from generated content, it reads content with `file.string_content` property (Implementation from master branch used to crash in this case since tried to load non existing file from file system).

Error messages are also changed to reference path and plugin that has generated the file. 

For case where file is loaded form filesystem the behavior and messages are kept as is to ensure backward compatibility.


squidfunk:
Thanks! Great work! I'm currently tight on time, but I'll try to get back to this ☺️

KonstantinNosov:
a minimum code generating blog post. Should fail without my changes and work with them

```python

POST_CONTENT="""---
date:
  created: {date}
tags:
  - generated
title: Generated Blog Post
categories:
  - test
---

This post is generated


"""


class BlogPostGeneratingPlugin(BasePlugin):
    def on_files(self, files: Files, /, *, config: MkDocsConfig) -> Files | None:
        file = File.generated(config=config, src_uri="blog/posts/generated.md", content=POST_CONTENT.format(date=datetime.now()))
        files.append(file)
        return files
```

KonstantinNosov:
@squidfunk is there a chance for this PR to be accepted? I have a hack to workaround the issue in my [project](https://docs.exordis.com/documentation/partial-documentation/#mkdocs-material-blogs) and either need to invest time to cover it with tests and improve code if it is permanent or just wait till it is fixed in mkdocs-material and remove...

squidfunk:
Thanks again for your time investment and efforts here. We currently have a feature freeze due to the [foundational work](https://squidfunk.github.io/mkdocs-material/blog/2024/08/19/how-were-transforming-material-for-mkdocs/) that's happening right now. Since this PR improves deeper customizations and integrations rather than common use cases, we're not considering it for merge in the light of our current efforts.

I'm sorry for any inconveniences caused. We'll come back to it after the next announcement.

KonstantinNosov:
No worries. It is important to keep focus.

wtch-ops (commented):
KonstantinNosov:blog-no-fs

squidfunk:

PR #7602, opened 2024-10-08
Fix: equation compatible with KateX and GitHub

FrancoisCapon:
Hi,

With my students we discovered that the former equation is not compatible with GitHub and KateX

**GitHub**

The following macros are not allowed: operatorname

**Katex**
```
auto-render.min.js:1 
 KaTeX auto-render: Failed to parse `
\operatorname{ker} f=\{g\in G:f(g)=e_{H}\}{\mbox{.}}
` with  ParseError: KaTeX parse error: Undefined control sequence: \mbox at position 45: …G:f(g)=e_{H}\}{\̲m̲b̲o̲x̲{.}}
````

### DIFF
--- a/docs/reference/math.md
+++ b/docs/reference/math.md
@@ -138,14 +138,14 @@ lines:
 
 ``` latex title="block syntax"
 $$
-\operatorname{ker} f=\{g\in G:f(g)=e_{H}\}{\mbox{.}}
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. So you're proposing another example that works with both, MathJax and KaTeX, do I understand that correctly? Otherwise, I'm not sure why the entire formula was changed 😅

squidfunk:
Alright, let's use this one then.

Issue #7600, opened 2024-10-07
Built-in `privacy` plugin does not download external assets of embedded .html file


mrsmrynk:
### Context

As discussed in #7596

I use an inline frame (`<iframe>`) to embed .html files in my documentation.
The .html files are generated by [folium](https://python-visualization.github.io/folium/latest/) and download .js files from various CDNs.
The built-in `privacy` plugin is used for self-hosting of these external assets.

### Bug description

The built-in `privacy` plugin does not download the external assets of the embedded .html file.

I would expect it to:
- download the external assets to the `site` directory
- replace all references with links to the downloaded copies

### Related links

- [Built-in `privacy` plugin](https://squidfunk.github.io/mkdocs-material/plugins/privacy/)

### Reproduction

[9.5.39-privacy-with-embedded-html.zip](https://github.com/user-attachments/files/17284246/9.5.39-privacy-with-embedded-html.zip)

### Steps to reproduce

1. run `mkdocs serve`
2. check the `.cache/plugin/privacy` directory - the external assets of the embedded .html file are not downloaded
3. check the `docs/maps/bounding_box.html` file - the references are not replaced with links to the downloaded copies

### Before submitting

- [X] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
- [X] I have attached links to [the documentation](https://squidfunk.github.io/mkdocs-material/), and possibly related [issues](https://github.com/squidfunk/mkdocs-material/issues) and [discussions](https://github.com/squidfunk/mkdocs-material/discussions).
- [X] I assure that I have [removed all customizations](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/#remove-customizations) before submitting this bug report.
- [X] I have attached a __.zip file__ with a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/) using the [built-in info plugin](https://squidfunk.github.io/mkdocs-material/plugins/info/).

squidfunk:
Thanks for reporting. While this might not be obvious, for MkDocs to consider your HTML file as something to process, it must be listed under `extra_templates`. If you add the following to your `mkdocs.yml`:

``` yaml
extra_templates:
  - maps/bounding_box.html
```

The privacy plugin will process it and download all assets:

```
[...]

PR #7604, opened 2024-10-08
Add HTML reference to Limitations in privacy plugin docs

mrsmrynk:
As discussed in #7600

I've added the HTML reference to the **Limitations** section, as it is too niche to fit into the **How it works** section without causing clutter, in my opinion.
Tried to be as concise as possible ☺️


squidfunk:
Thanks! LGTM.

Issue #2388, opened 2021-03-04
The logo link points to site_url if it's set; this can sometimes conflict with mike


jimporter:
This is a bit of a weird issue, so I'll do my best to explain it. I recently landed a fix for https://github.com/jimporter/mike/issues/40 which, among other things, allows you to set a `canonical_version` in the `mike` MkDocs plugin; this is there to allow for SEO so that the `<link rel="canonical">` tag always points to the `latest` version alias. That way, when search engines index your site, they'll link to the `/latest/` version, rather than showing results for some random old doc version that was current when they indexed the site.

However, this causes an odd behavior with the Material theme. Suppose I have 2 versions: `1.0` and `2.0` (aliased with `latest`), with `canonical_version` set to `latest`. If I go to the 1.0 docs and click the top left logo, I'll be taken to the homepage for the 2.0 docs because the `site_url` is `https://example.com/latest/`. I think that's the wrong behavior, and a user would expect to stay on the 1.0 docs. This is only an issue when `canonical_version` is explicitly set though; when it's unset, things should work fine.

Note: since https://squidfunk.github.io/mkdocs-material-example-versioning/ was built with mike 0.5, the canonical URL is set incorrectly there; that'll be fixed in mike 1.0 via [this commit](https://github.com/jimporter/mike/commit/bf4f54f9e8d5a2ff70bf7d071e659e725e85866a).

squidfunk:
Thanks for reporting. I agree, that the user should stay on the current version– only the version selector should change the version, not internal navigation. The `site_url` is determined here:

https://github.com/squidfunk/mkdocs-material/blob/73602b23e158c42021d41a31e254d8448f924aa6/src/partials/header.html#L23-L27

And then used here:

https://github.com/squidfunk/mkdocs-material/blob/73602b23e158c42021d41a31e254d8448f924aa6/src/partials/header.html#L37-L44

Do you have an idea what we need to change to support `mike`?

jimporter:
I think what we'd need to do is *only* use `config.site_url` for setting the canonical URL. Then we'd change line 24 to something like this:

```
{% set base_url = d(nav.homepage.url, true) | url %}
```

[...]

Issue #7619, opened 2024-10-15
The tooltip style of the logo is lost in instant loading mode


Xiaokang2022:
### Context

Python: `3.13.0`
OS: `Windows11 23H2`

### Bug description

With the Instant Loading mode, the style of the logo's tooltip is only correct the first time, and will lose the style after clicking on another page.

### Related links

- [Reporting a bug](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/)
- https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/?h=instant#instant-loading

### Reproduction

[9.5.40-logo-bug.zip](https://github.com/user-attachments/files/17371841/9.5.40-logo-bug.zip)

### Steps to reproduce

1. Refresh the current page
2. Check that the style of the logo's tooltip is correct
3. Click on a different page or click on the page again and check again
4. Styles are missing

### Browser

Edge

### Before submitting

- [X] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
- [X] I have attached links to [the documentation](https://squidfunk.github.io/mkdocs-material/), and possibly related [issues](https://github.com/squidfunk/mkdocs-material/issues) and [discussions](https://github.com/squidfunk/mkdocs-material/discussions).
- [X] I assure that I have [removed all customizations](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/#remove-customizations) before submitting this bug report.
- [X] I have attached a __.zip file__ with a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/) using the [built-in info plugin](https://squidfunk.github.io/mkdocs-material/plugins/info/).

squidfunk:
Fixed in 53cb6a625. We must remount the header component on instant navigation to account for tooltips that are not hidden in nested components in the header (like repository or search). This commit moves the header to the `$content` observable chain, which will now update correctly.

squidfunk:
Released as part of [9.5.41](https://squidfunk.github.io/mkdocs-material/changelog/#9.5.41).

PR #7620, opened 2024-10-15
Fix a bug where the tooltip style of the logo was missing in instant loading mode

Xiaokang2022:
Issue: #7619 

I've tried to fix this, and while I'm not sure if my fix makes sense, it works. If there is a problem, please point it out, thanks.


squidfunk:
Thanks for the PR and digging into this problem! Unfortunately, we must update the logo component on instant navigation, which we added in https://github.com/squidfunk/mkdocs-material/commit/859173ef88f08a5ff226f08b99f9d00880498aa5 to fix https://github.com/squidfunk/mkdocs-material/issues/2388. Without this fix, the logo will not correctly be updated when using versioning.

Closing this PR as not mergable. We need to look for another solution.

PR #7621, opened 2024-10-15
Fix the issue of mismatched regional flag

myyerrol:
Hello squidfunk, I think mkdocs-material is a very clean and beautiful mkdocs theme that helped me quickly build an open source project website, thank you very much about that! But in the process of setting the language of the website, I found that the flag of **Taiwan** is incorrect. The **one-China principle** is an international consensus, Taiwan, as a part of China, should be set up as the **China** flag, I hope you can help correct it, thank you!!!


myyerrol:
OK, I understand your meaning, thanks for your reply!

PR #7622, opened 2024-10-15
Add github-callouts as callouts markdown extension

mschoettle:
The `markdown-callouts` extension also supports GitHub-style alerts.
Instead of `callouts` it is enabled via `github-callouts`.

See here: https://github.com/oprypin/markdown-callouts/releases/tag/v0.4.0

### DIFF
--- a/docs/schema/extensions/markdown.json
+++ b/docs/schema/extensions/markdown.json
@@ -32,7 +32,8 @@
       "markdownDescription": "https://github.com/oprypin/markdown-callouts",
       "enum": [
         "markdown.extensions.callouts",
-        "callouts"
+        "callouts",
+        "github-callouts"
       ]
     },
     {


squidfunk:
Thanks for the PR! We do not officially support this Markdown extension, which is why we won't consider adding it to our documentation. Please note that MkDocs is an ecosystem of many loosely coupled projects. I'm afraid with our limited time and small team, we can't support everything. Thanks for your understanding!

squidfunk:
Ah wait, no, sorry, I've mistaken that for a docs change. Sorry. Of course we can add it to the schema 😅

mschoettle:
Thanks! I should have probably added “to schema” in the title to make it less ambiguous. 😃

squidfunk:
No, no fault on your side. I should've read better – I was in a hurry 😅 Shouldn't tackle issues and PRs then.

PR #7628, opened 2024-10-18
Remove mkdocs-material-extensions from requirements.txt

clintonsteiner:
Per mkdocs-material-extensions and the repos code, this is no longer a required dependency and the logic has been put into this repo

Searching the repo, there were no usages of materialx

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,7 +22,6 @@
 jinja2~=3.0
 markdown~=3.2
 mkdocs~=1.6
-mkdocs-material-extensions~=1.3
 pygments~=2.16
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. We have deprecated it, but to remove it, we need to release a new major version. Thus, we can't remove it right now, as there's currently no major version queued for release.

squidfunk:
If we remove it now, and some users still use it, it's a breaking change to their configuration, which will break their CI workflows. It's currently only a warning. The best practice is to deprecate it and remove it on the next major release, which might include breaking changes. We won't issue a major release only for this tiny change.

PR #7650, opened 2024-10-29
Handle external images for SVGs in privacy plugin

nejch:
Some users (like us) might have quirks around serving SVG images/logos, so we include them from remote locations, e.g. for licensing reasons etc.

Since we can't include them locally but some users would like to have offline builds, this PR adds support for loading remote images from SVGs in the privacy plugin. WDYT? See also some references below :bow:  

* SVG `<image>` reference: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/image
* example site with external SVG: https://opensource.siemens.com/
* regex match against real example: https://regex101.com/r/3Dp1TP/1

:hammer_and_wrench: with :heart: at Siemens

### DIFF
--- a/material/plugins/privacy/plugin.py
+++ b/material/plugins/privacy/plugin.py
@@ -313,12 +313,19 @@ class PrivacyPlugin(BasePlugin[PrivacyConfig]):
                     file = self._queue(url, config)
                     el.set("src", resolve(file))
 
+            # Handle external image in SVG
+            if el.tag == "image":
+                url = urlparse(el.get("href"))
+                if not self._is_excluded(url, initiator):
+                    file = self._queue(url, config)
+                    el.set("href", resolve(file))
+
             # Return element as string
             return self._print(el)
 
         # Find and replace all external asset URLs in current page
         return re.sub(
-            r"<(?:(?:a|link)[^>]+href|(?:script|img)[^>]+src)=['\"]?http[^>]+>",
+            r"<(?:(?:a|link|image)[^>]+href|(?:script|img)[^>]+src)=['\"]?http[^>]+>",
             replace, output, flags = re.I | re.M
         )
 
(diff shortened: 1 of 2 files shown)


squidfunk:
Thanks for the PR! Could you please provide a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/), so we can understand what this PR is trying to solve exactly? Also, if no trouble, could you please include the other cases we already support, so we can estimate whether this is something we're happy to support? As mentioned, always a good idea to [talk to us first](https://squidfunk.github.io/mkdocs-material/contributing/making-a-pull-request/) to omit duplicate work, but I think this is something worth supporting and should go into master, once we know that it aligns with our goals 😅

nejch:
> Thanks for the PR! Could you please provide a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/), so we can understand what this PR is trying to solve exactly? Also, if no trouble, could you please include the other cases we already support, so we can estimate whether this is something we're happy to support? As mentioned, always a good idea to [talk to us first](https://squidfunk.github.io/mkdocs-material/contributing/making-a-pull-request/) to omit duplicate work, but I think this is something worth supporting and should go into master, once we know that it aligns with our goals 😅

Ah sorry got a little ahead of myself, will provide a little repro project for both!

squidfunk:
No worries! At some point in the future we'll have tests, that'll make it easier 😅

nejch:
@squidfunk here's a mini repro: [9.5.42-external-svg-images-excluded-in-privacy-plugin.zip](https://github.com/user-attachments/files/17571192/9.5.42-external-svg-images-excluded-in-privacy-plugin.zip)

You can see with the current version the background is still loaded from the external source, which won't work offline.

I just used https://www.w3schools.com/graphics/tryit.asp?filename=trysvg_image2 but I think you'll get the idea :slightly_smiling_face: I'll get to the other issue in a bit!

squidfunk:
This one LGTM, thanks for the PR again!

squidfunk:
Released as part of [9.5.43](https://squidfunk.github.io/mkdocs-material/changelog/#9.5.43).

PR #7651, opened 2024-10-29
Handle quoted external CSS URLs in privacy plugin

nejch:
This one is very similar to https://github.com/squidfunk/mkdocs-material/pull/7650, but keeping it separate as it covers different files. I was wondering why our external woff/woff2 files weren't being downloaded and noticed this issue.

* CSS map regex test: https://regex101.com/r/UG2Bv0/1
* JS map regex test: https://regex101.com/r/zPRrTp/1

The back-reference helps get the exact URL and to avoid more false positives on invalid quoting, though there are still some - but still better than just expecting quotes on either side IMO. Also, this is probably caught by CSS linters and build system anyway.

I had to introduce named groups to handle this, hence the slight change in logic as `findall` doesn't support named groups. But IMO this might also help in the future if more edge cases show up.

🛠️ with ❤️ at Siemens

### DIFF
--- a/material/plugins/privacy/plugin.py
+++ b/material/plugins/privacy/plugin.py
@@ -64,8 +64,8 @@ class PrivacyPlugin(BasePlugin[PrivacyConfig]):
         # Initialize collections of external assets
         self.assets = Files([])
         self.assets_expr_map = {
-            ".css": r"url\((\s*http?[^)]+)\)",
-            ".js": r"[\"'](http[^\"']+\.(?:css|js(?:on)?))[\"']",
+            ".css": r"url\(\s*([\"']?)(?P<url>http?[^)'\"]+)\1\s*\)",
+            ".js": r"[\"'](?P<url>http[^\"']+\.(?:css|js(?:on)?))[\"']",
             **self.config.assets_expr_map
         }
 
@@ -271,7 +271,8 @@ class PrivacyPlugin(BasePlugin[PrivacyConfig]):
         # Find and extract all external asset URLs
         expr = re.compile(self.assets_expr_map[extension], flags = re.I | re.M)
         with open(initiator.abs_src_path, encoding = "utf-8-sig") as f:
-            return [urlparse(url) for url in re.findall(expr, f.read())]
+            results = re.finditer(expr, f.read())
+            return [urlparse(result.group("url")) for result in results]
 
     # Parse template or page HTML and find all external links that need to be
     # replaced. Many of the assets should already be downloaded earlier, i.e.,
--- a/src/plugins/privacy/plugin.py
+++ b/src/plugins/privacy/plugin.py
@@ -64,8 +64,8 @@ class PrivacyPlugin(BasePlugin[PrivacyConfig]):
         # Initialize collections of external assets
         self.assets = Files([])
         self.assets_expr_map = {
-            ".css": r"url\((\s*http?[^)]+)\)",
-            ".js": r"[\"'](http[^\"']+\.(?:css|js(?:on)?))[\"']",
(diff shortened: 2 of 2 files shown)


squidfunk:
Same as https://github.com/squidfunk/mkdocs-material/pull/7650#issuecomment-2444523329, need a little more input.

squidfunk:
Looking at the CSS regex, in your example, cases that should not match still do:

Bildschirm­foto 2024-10-30 um 10 15 42

Moving the `?` quantifier into the capturing group will allow the back reference to accept an empty match:

``` regex
url\(([\"']?)(?P<url>\s*http?[^)'\"]+)\1\)
```

Bildschirm­foto 2024-10-30 um 10 15 54

nejch:
> Looking at the CSS regex, in your example, cases that should not match still do:
> ...
> Moving the `?` quantifier into the capturing group will allow the back reference to accept an empty match:
> ...
> ```regexp
> url\(([\"']?)(?P<url>\s*http?[^)'\"]+)\1\)
> ```

Ah nice catch, missed that one! Will adapt the PR.

Here's a reproduction zip with all 3 quote styles:

[9.5.42-css-quoted-urls-missing-from-privacy-plugin.zip](https://github.com/user-attachments/files/17571711/9.5.42-css-quoted-urls-missing-from-privacy-plugin.zip)

With the current version, only the unquoted external URL gets fetched.

Let me know if these 2 repro zips are clear enough @squidfunk :bow:

squidfunk:
Thanks! I believe we need to allow spaces before and after the URL, as according to syntax level 3, specifically [how URL tokens are parsed](https://www.w3.org/TR/css-syntax-3/#consume-url-token). There are two ways how URLs are handled: if they contain a string, they're just considered to be normal function tokens, which definitely demands for whitespace after and before. If they contain a verbatim URL, I believe we need to support whitespace as well.

``` regex
url\(\s*([\"']?)(?P<url>http?[^)'\"]+)\1\s*\)
```

This now works correctly with the following strings, albeit it consumes the trailing whitespace on the verbatim version, which should not be a problem though:

``` css
/* correct */
url("https://example.com/images/myImg.jpg");
url('https://example.com/images/myImg.jpg');
url(  'https://example.com/images/myImg.jpg'  );
url(  "https://example.com/images/myImg.jpg"  );
url(https://example.com/images/myImg.jpg);
url(   https://example.com/images/myImg.jpg   );


/* mismatching */
url('https://example.com/images/myImg.jpg);
url("https://example.com/images/myImg.jpg);
url('https://example.com/images/myImg.jpg");
url(https://example.com/images/myImg.jpg');
url(https://example.com/images/myImg.jpg");

/* non-http links */
url("data:image/jpg;base64,iRxVB0…");
url(myImg.jpg);
url(#IDofSVGpath);
```

PS: How can I share on regex101? I'm too stupid to find the share button 😅

nejch:
Perfect, thanks! Added the new pattern.

Hehe, Ctrl+S should save it and give you a modal with the link: https://regex101.com/r/LVJJfK/1https://regex101.com/r/LVJJfK/1. Not sure if it requires starting over when receiving a link though.

I also just checked and as you say at least urlparse seems to be happy with trailing whitespace:

```python
>>> from urllib.parse import urlparse
>>> urlparse("https://example.com/images/myImg.jpg   ")
ParseResult(scheme='https', netloc='example.com', path='/images/myImg.jpg   ', params='', query='', fragment='')

```

squidfunk:
Perfect, thanks for investigating! I think this is safe to merge then 🤟

squidfunk:
Released as part of [9.5.43](https://squidfunk.github.io/mkdocs-material/changelog/#9.5.43).

PR #7699, opened 2024-11-14
Add safety net for failing github action

Nicolasara:
If someone adds the code block in the customization docs  

https://squidfunk.github.io/mkdocs-material/customization/#setup-and-theme-structure

And if they use the cli command provided their github action will fail.

https://github.com/Nicolasara/IIT-Docs/actions/runs/11831606702/job/32967030933

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -50,6 +50,7 @@ contents:
               restore-keys: |
                 mkdocs-material-
           - run: pip install mkdocs-material # (4)!
+          - run: mkdir overrides # (5)!
           - run: mkdocs gh-deploy --force
     ```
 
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. I'm not sure I understand. So this is for the case where somebody sets the `custom_dir` setting, or just blindly copies the code block from the customization guide, and then runs `mkdocs build`? `overrides` is only a recommendation for the name that hosts the overridden partials, it is not static. I understand that you ran intro troubles, but the error message that is printed is absolutely clear what's wrong:

```
ERROR   -  Config value 'theme': The path set in custom_dir ('/home/runner/work/IIT-Docs/IIT-Docs/overrides') does not exist.
```

Thus, it's an easy fix, and we don't need to add more noise, keeping it minimal.

Nicolasara:
Sounds good, I know it's an easy fix but I was just thinking about people who might be less technical and are just following along.

Nicolasara:
Regardless, great theme, thanks!

PR #7704, opened 2024-11-15
Add MapLibre to Open Source Projects list

louwers:
We're happy users of Material for MkDocs. :)

MapLibre GL JS also uses it https://maplibre.org/maplibre-gl-js/docs/

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,7 @@ For detailed installation instructions, configuration options, and a demo, visit
 [Knative](https://knative.dev/docs/),
 [Kubernetes](https://kops.sigs.k8s.io/),
 [kSQL](https://docs.ksqldb.io/),
+[MapLibre](https://maplibre.org/maplibre-style-spec/),
 [NetBox](https://netboxlabs.com/docs/netbox/en/stable/),
 [Nokogiri](https://nokogiri.org/),
 [OpenFaaS](https://docs.openfaas.com/),


squidfunk:
Thanks! MapLibre has 6.7k stars, so I think this qualifies as a project to add to the list.

squidfunk:
However, the link you posted is not https://maplibre.org/maplibre-gl-js/docs/ – the project you linked only has very few stars. We should use MapLibre instead.

louwers:
@squidfunk The MapLibre Style Spec is used by both MapLibre GL JS and MapLibre Native. That's why I shared that particular docs site.

MapLibre GL JS has 6.7k stars and almost 370k+ monthly downloads on npm. MapLibre Native has 1k GitHub Stars, but is used in many apps and has many millions of daily users.

squidfunk:
I see. So we won't add projects that don't have at least a few thousand stars on their back. The main idea for this section o the REAME is for users to quickly be able to find a project that they know, so it has to be popular. For this reason, we're currently not considering adding this project. I hope for your understanding. If we would add less poplar projects, the section wouldn't have much value. For instance, we just added [OpenSSL](https://github.com/squidfunk/mkdocs-material/commit/c4edbbca0095ac20184441f82a9e379028c7c54e).

louwers:
We set up a different repo for the style spec documentation, so the star count of the repo is a bit misleading.

Happy to add the documentation of MapLibre GL JS to the list instead.

PR #7718, opened 2024-11-19
Add Caption extension to JSON schema

myl7:
_Caption_ Python Markdown extension is introduced in the doc and implemented, but not defined in the JSON schema.
Users who follows the [doc](https://squidfunk.github.io/mkdocs-material/creating-your-site/#minimal-configuration) to add the schema in VSCode would find a warning on the value `pymdownx.blocks.caption`.
This PR adds the extension to the schema along with its options.

See also:
- https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown-extensions/#caption
- https://facelessuser.github.io/pymdown-extensions/extensions/blocks/plugins/caption/

### DIFF
--- a/docs/schema/extensions/pymdownx.json
+++ b/docs/schema/extensions/pymdownx.json
@@ -68,6 +68,73 @@
     {
       "oneOf": [
         {
+          "type": "object",
+          "properties": {
+            "pymdownx.blocks.caption": {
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! This was a recent addition, so it was forgotten to update the schema.

PR #7733, opened 2024-11-22
Include preload hints in privacy plugin replacements

### LINKED ISSUE #7732: Privacy plugin: Include preloads in URL replacements
### Context

Follow-up to but slightly different from https://github.com/squidfunk/mkdocs-material/pull/7650.

### Description

We have some `preload`s in our child theme for custom fonts defined in our `mkdocs_theme.yml`.  Currently, these are ignored by the replace mechanism in the privacy plugin.

Although this is just an optimization and leaving original links shouldn't be much of an issue in my opinion, some of our users want to get rid of external references altogether when building for offline deployments.

Looking at https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload, preloading local files still seems to be a valid use for this, so it's not entirely irrelevant. The change would be trivial, e.g.:

```diff
-                    if rel == "stylesheet" or rel == "icon":
+                    if rel in ("icon", "preload", "stylesheet"):
```

This is not really important, since `preconnect` rels are already also ignored, but it's something our users have reported so just wanted to discuss. I'm happy to open a little PR to extend this if you agree!

### Related links

- https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload

### Use Cases

Offline builds where the user wants to remove links to external resources or ensure no requests are attempted at all.

### Visuals

_No response_

### Before submitting

- [X] I have included relevant links to [the documentation](https://squidfunk.github.io/mkdocs-material/), related [issues](https://github.com/squidfunk/mkdocs-material/issues), and [discussions](https://github.com/squidfunk/mkdocs-material/discussions) to underline the need for my idea.

nejch:
Closes #7732 

Did a bit of ABC sorting while at it in case there's more later :)

### DIFF
--- a/material/plugins/privacy/plugin.py
+++ b/material/plugins/privacy/plugin.py
@@ -302,8 +302,8 @@ class PrivacyPlugin(BasePlugin[PrivacyConfig]):
                     if rel == "preconnect":
                         return ""
 
-                    # Replace external style sheet or favicon
-                    if rel == "stylesheet" or rel == "icon":
+                    # Replace external favicon, preload hint or style sheet
+                    if rel in ("icon", "preload", "stylesheet"):
                         file = self._queue(url, config)
                         el.set("href", resolve(file))
 
--- a/src/plugins/privacy/plugin.py
+++ b/src/plugins/privacy/plugin.py
@@ -302,8 +302,8 @@ class PrivacyPlugin(BasePlugin[PrivacyConfig]):
                     if rel == "preconnect":
                         return ""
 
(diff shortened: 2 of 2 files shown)


squidfunk:
Thanks! Note that the changes also need to be made in the `material` repository – thus you need to [build the theme](https://squidfunk.github.io/mkdocs-material/customization/#development-mode).

squidfunk:
Yeah, we definitely need to remove the copying of the Python code, but it was easier at that time 😅

PR #7752, opened 2024-11-26
Update getting-started.md - additional installation info

MaximilianKohler:
When installing mkdocs on a new PC I was running into errors due to missing this additional set of instructions.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -174,4 +174,8 @@ Next, install the theme and its dependencies with:
 pip install -e mkdocs-material
 ```
 
+### Additional steps
+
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. The plugins are opt-in, they must be enabled specifically, and the configuration and installation guides of the plugins mention this. Thus, this information is already contained in the documentation, IMHO, at a better location than the getting started guide, because this guide is mainly for beginners. Specifically:

[Social plugin reference](https://squidfunk.github.io/mkdocs-material/plugins/social/#configuration)

> However, in order to generate social card images, it's necessary to install the dependencies for [image processing](https://squidfunk.github.io/mkdocs-material/plugins/requirements/image-processing/), if they're not already available on your system. The linked guide includes instructions for several operating systems and mentions some alternative environments.

[Social plugin configuration](https://squidfunk.github.io/mkdocs-material/setup/setting-up-social-cards/#configuration)

> The built-in social plugin automatically generate a custom preview image for each page. Install all [dependencies for image processing](https://squidfunk.github.io/mkdocs-material/plugins/requirements/image-processing/) and add the following lines to mkdocs.yml:

We're considering this to be enough.

PR #7767, opened 2024-12-03
Dockerfile: improve build speed and image size

FrancoisCapon:
Hello,

Thanks for your feedback on my previous PR on Dockerfile.

Here's another one :
- that uses **uv** to speed up "pip installs" (**total gain of  ~30%**)
```
@home
docker system prune -f && build with no cache:
[+] Building 33.1s (22/22) FINISHED 
[+] Building 51.6s (15/15) FINISHED

@github action
#7767 Build Docker image 40s
#7675 Build Docker image 57s
```
- that go **below the 150MB** size for the image
```
Total Image size: 149 MB         
Potential wasted space: 0 B
Image efficiency score: 100 % 
```

Perhaps, you know the documentation framework used :wink: 
https://docs.astral.sh/uv/ 

By the way, no feedback on my idea of [setting palettes for pages and sections](https://github.com/squidfunk/mkdocs-material/discussions/7736)?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,13 +19,11 @@
 # IN THE SOFTWARE.
 
 FROM python:3.11-alpine3.19 AS build
-
 # Build-time flags
 ARG WITH_PLUGINS=true
 
 # Environment variables
 ENV PACKAGES=/usr/local/lib/python3.11/site-packages
-ENV PYTHONDONTWRITEBYTECODE=1
 
 # Set build directory
 WORKDIR /tmp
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. However, we're not considering merging this right now. The reason is that we currently do not want to introduce any new dependencies, because we're working from first principles in the background, hence we have a feature freeze. I understand this is an optimization, but it bears the risk of more maintenance down the road.

I hope for your understanding. Thanks!

PR #7796, opened 2024-12-15
docs(offline-plugin): Update limitations

niclasheinz:
In many discussions (e.g #7793) I have seen that users have problems with the offline plugin. Therefore, I have edited the limitations section so that users can see in the documentation that the use_directory_urls setting is disabled.

### DIFF
--- a/docs/plugins/offline.md
+++ b/docs/plugins/offline.md
@@ -147,6 +147,8 @@ file system:
 - [Versioning]
 - [Comment systems]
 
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. This is already mentioned in the [How it works](https://squidfunk.github.io/mkdocs-material/plugins/offline/#how-it-works) section, which is a better fit than the limitations, as disabling the setting is not a limitation. Closing as unactionable.

PR #7808, opened 2024-12-18
Docs(Publish your Page): Update Gitlab CI

niclasheinz:
Hi @squidfunk 

I have edited the Gitlab CI example. Normally, you will find a Gitlab pipeline configuration like this.

Greetings
Niclas

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -168,6 +168,9 @@ contents:
 === "Material for MkDocs"
 
     ``` yaml
+    stages:
+      - deploy # (1)!
+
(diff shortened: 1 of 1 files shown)


niclasheinz:
The use of ‘stages’ in a Gitlab pipeline is more elegant than without. Of course, the pipeline can still be executed. However, the more complex the pipeline becomes, the more error-prone it is. In addition, the argument `stage` is included in the `job` itself [here](https://squidfunk.github.io/mkdocs-material/publishing-your-site/#__codelineno-3-2).

PR #7812, opened 2024-12-18
Add `"relative_to_docs"` value to `absolute_links` enum

Bzly:


docs/schema/validation.json
```diff
@@ -43,7 +43,7 @@
     },
     "absolute_links": {
       "title": "warning level when absolute links are used",
-      "$ref": "#/$defs/warning_levels"
+      "enum": ["warn", "info", "ignore", "relative_to_docs"]
```
squidfunk, line 46: Thanks for the PR. Could we please keep the definition reference, and add the `relative_to_docs` as another option, possibly by using `oneOf` or similar? [This would also be in line with how MkDocs implements it](https://github.com/mkdocs/mkdocs/blob/7e4892ab2dd4a52efd95a9de407eee63310e0780/mkdocs/config/defaults.py#L28-L32).

squidfunk (changes requested):

docs/schema/validation.json
Bzly: Thanks; my search of the JSON schema docs was not thorough enough to find `oneOf`. 

I've pushed a new commit.

squidfunk:
Perfect, LGTM!

squidfunk (approved):

alisa-dermen:
When is this fix going to be released?

squidfunk:
It became part of our schema the moment it was merged – nothing to release:

https://github.com/squidfunk/mkdocs-material/blob/45f60a66903105d7a2fcb2a1ad64c3b347723f42/docs/schema/validation.json#L44-L50

PR #7816, opened 2024-12-19
small improvements and a correction

niclasheinz:
Changes:

- Fixes line highlighting in [codeblock](https://squidfunk.github.io/mkdocs-material/tutorials/blogs/basic/#pinning)
- Update trusted by users count in [philosophy](https://squidfunk.github.io/mkdocs-material/philosophy/), because it's outdated
- Add a small thing to "Works on all devices"  in [philosophy](https://squidfunk.github.io/mkdocs-material/philosophy/)

### DIFF
--- a/docs/philosophy.md
+++ b/docs/philosophy.md
@@ -30,6 +30,6 @@ discusses the [conventions] used in this documentation.
   documentation with touch devices, keyboard, and screen readers. Semantic
   markup ensures that your documentation works for everyone.
 
-- __Open Source__: Trust 55,000+ users – choose a mature and well-funded
+- __Open Source__: Trust 50,000+ users – choose a mature and well-funded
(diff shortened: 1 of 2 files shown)


squidfunk:
Thanks! Actually, the texts are not synced with the latest versons on the README. Same goes for the landing page texts in Insiders. Also, we're calculating with 50,000 users, since it's 55,000 proiects + a lot of projects inside companies, so 50k is a conservative estimate. Could you maybe also synchronize the texts on the landing page and the philosophy page as well? You can take the README as the source of truth. We really need to deduplicate those at some point.

niclasheinz:
I'll get it done later today.

niclasheinz:
I've adopt the text in philosophy from README

squidfunk:
Perfect, thank you!

Issue #7822, opened 2024-12-22
Recommend adding a link to the community-supported documentation for publishing to ReadTheDocs


tleonhardt:
### Description

It would be nice to include a link to the official `ReadTheDocs` [documentation](https://docs.readthedocs.io/en/stable/intro/mkdocs.html) to the bottom of the [Publishing your site](https://squidfunk.github.io/mkdocs-material/publishing-your-site/#other) documentation for MkDocs Material along the other community-supported documentation.

### Related links

- [ReadTheDocs guide for MkDocs Material](https://docs.readthedocs.io/en/stable/intro/mkdocs.html)
- [Other section of MkDocs Material Publishing docs](https://squidfunk.github.io/mkdocs-material/publishing-your-site/#other)
- [Source file that needs to change](https://github.com/squidfunk/mkdocs-material/blob/master/docs/publishing-your-site.md#other)

### Proposed change

Edit the - [publishing-your-site.md](https://github.com/squidfunk/mkdocs-material/blob/master/docs/publishing-your-site.md#other) file to contain one additional bullet at the bottom with a link to the `ReadTheDocs` documentation for publishing MkDocs Material docs there.

### Before submitting

- [X] I have read and followed the [documentation issue guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-docs-issue/).
- [X] I have attached the links to the affected sections of [the documentation](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-docs-issue/#related-links)

niclasheinz:
Hi @tleonhardt 

Thanks for your suggestions. I'll check the links and make a PR.

squidfunk:
Thanks for suggesting! I'm aware of this guide. If you closely read the publishing guide that you linked, you'll realize that it does not include a proper configuration of Material for MkDocs. In fact, the information in this guide is wrong:

> Minimal configuration required to build an existing Material for MkDocs project on Read the Docs looks like this, specifying a python toolchain on Ubuntu, defining the location of the installation requirements, and using the built-in [mkdocs](https://docs.readthedocs.io/en/stable/config-file/v2.html#mkdocs) command:
>
> ``` yaml
>  version: 2
> 
>  build:
>    os: ubuntu-24.04
>    tools:
>      python: "3"
> 
>  python:
>    install:
>      - requirements: requirements.txt
> 
>  mkdocs:
>    configuration: mkdocs.yml
> ```

[...]

PR #7835, opened 2024-12-23
Fix CF publishing link

ericholscher:
This was previously 404'ing.

Refs https://github.com/squidfunk/mkdocs-material/issues/7822#issuecomment-2560357943


squidfunk:
Thanks for the PR! I seem to have missed that one.

PR #7844, opened 2024-12-26
deps(mkdocs-git-committers-plugin-2) Bump to v2.4.1

niclasheinz:
I've tested it locally. It works seamlessly.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,7 +59,7 @@ recommended = [
   "mkdocs-rss-plugin~=1.6"
 ]
 git = [
-  "mkdocs-git-committers-plugin-2~=1.1",
+  "mkdocs-git-committers-plugin-2~=2.4.1",
   "mkdocs-git-revision-date-localized-plugin~=1.2,>=1.2.4"
 ]
 imaging = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,7 +59,7 @@ recommended = [
   "mkdocs-rss-plugin~=1.6"
 ]
 git = [
-  "mkdocs-git-committers-plugin-2~=2.4.1",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -59,7 +59,7 @@ recommended = [
   "mkdocs-rss-plugin~=1.6"
 ]
 git = [
-  "mkdocs-git-committers-plugin-2~=1.1",
+  "mkdocs-git-committers-plugin-2~=2.4.1",
```
squidfunk, line 62: Please limit this to `~=2.4`, not `~=2.4.1` – those have different meanings:

1. `~= 2.4`: means `>=2.4,<3`
2. `~=2.4.1`: means `>=2.4.1,<2.4`

We want 1., as we're only guarding against major version upgrades.

squidfunk (changes requested):

pyproject.toml
squidfunk: Thinking about it, I'm not sure whether we need to change anything here, as 1.1 still works from our side, since [2.0 only changed the plugin's internals and/or configuration](https://github.com/ojacques/mkdocs-git-committers-plugin-2/releases), without necessary adaptions of our integration. We can relax the requirement to include everything above `>=1.1` up to `<3`. Thus, I think we should set it to `>=1.1,<3`.

squidfunk: To sum up: we should always check what the absolute minimum requirements are and keep it at that as long as we can. This makes it easier for everyone downstream. My philosophy always is: as lax as possible, while guarding against major releases that might include breaking changes.

niclasheinz: Yes, that makes sense. I will change it

squidfunk: Perfect, thanks!

PR #7845, opened 2024-12-26
deps(mkdocs-git-revision-date-localized-plugin): Bump to v1.3.0

niclasheinz:
I've tested it locally. It works seamlessly

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ recommended = [
 ]
 git = [
   "mkdocs-git-committers-plugin-2~=1.1",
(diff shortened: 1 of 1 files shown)


squidfunk:
So this PR is not necessary to merge, because ~=1.2 includes the current version 1.3. Raising the requirement would mean that there might be plugins or installations breaking because of the new higher requirement. Thus, we can leave things as they are for now. The current version range includes everything between `>=1.2.4,<2`.

niclasheinz:
Good catch

PR #7876, opened 2025-01-06
Doc: navigation - tags in nav, use as list

bartdorlandt:
Had to make this a list for me to use tags as described.


squidfunk:
Yes, that's definitely incorrect. Thanks!

PR #7878, opened 2025-01-06
Update icons-emojis.md

bartdorlandt:
Fixing the pumping heart in div


bartdorlandt:
It might require more changes:
https://github.com/squidfunk/mkdocs-material/blob/caeb33e95536f528252465c4f656e7e4474c8218/src/overrides/assets/stylesheets/custom/_typeset.scss#L86

Here it is actually defined as .mdx-heart, and not as .heart as the example.
So perhaps some clarifying text is needed, or more files require changing.

bartdorlandt:
I messed up. Please ignore.

PR #7929, opened 2025-01-22
feat: add ability to control mermaid icon packs

kirillmakhonin-brt:
# About

This pull request adds an ability to control mermaid JS, particulary:

1. Load additional [icon packs](https://mermaid.js.org/config/icons.html)
2. Control version of Mermaid JS library

# Checklist

- [x] Documentation
- [x] Example
- [x] JSON Schema

### DIFF
--- a/docs/reference/diagrams.md
+++ b/docs/reference/diagrams.md
@@ -16,7 +16,7 @@ popular and flexible solution for drawing diagrams.
 <!-- md:version 8.2.0 -->
 
 This configuration enables native support for [Mermaid.js] diagrams. Material
-for MkDocs will automatically initialize the JavaScript runtime when a page 
+for MkDocs will automatically initialize the JavaScript runtime when a page
 includes a `mermaid` code block:
 
 ``` yaml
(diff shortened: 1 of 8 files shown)


squidfunk:
Thanks for the PR. We are currently in feature freeze, because we're putting all of our efforts into very important [foundational work](https://squidfunk.github.io/mkdocs-material/blog/2024/08/19/how-were-transforming-material-for-mkdocs/). Once that's done, we might consider adding this functionality.

shahida615:
shahida

egandro:
+1 for this feature

nice icons are missing and can only be added via a library

squidfunk:
Please use reactions on the original post. [Do not +1, as it only adds noise](https://squidfunk.github.io/mkdocs-material/contributing/#before-commenting).

robinlioret:
I was missing this feature, thanks a lot!

I have two suggestions:
- Parameter names should have consistent cases: every parameters are in snake_case. Wouldn't be better practice to keep it that way?
- I find very confusing the way logo groups are configured. `iconnify:logo@1` or `url:aws:...`  does not provide a clear view on how to refer to the group in the diagrams, nor how to configure it.

Regarding this last point, my idea would be to explicitly make the user specify the name and the type of the target. Here's an idea on what the configuration could look like:

```yaml
theme:
  mermaid:
    icon_packs:
      logos:
        type: iconify
        path: logos@1
      aws:
        type: url
        url:  https://raw.githubusercontent.com/awslabs/aws-icons-for-plantuml/v19.0
``` 

Using a dict instead of a list:
- ensures the uniqueness of the logo group name
- ensures the clarity of the group name and how to use it in the documentation's diagrams
- allow the user to use a custom name that fit is needs (enterprise conventions or else)

Explicit type clarify the documentation, Mkdocs Material could have a small paragraph dedicated to each type and how to configure it.

kirillmakhonin-brt:
@squidfunk could you please tell if this project is still in the feature freeze? I'm ready to update PR once you can confirm that you're ready to accept changes. This feature is popular within our company and it looks like community also has an interest, so we would like to help.

squidfunk:
Thanks again for the PR and the time you invested. We added the ability to load a custom version of Mermaid [3 weeks ago](https://github.com/squidfunk/mkdocs-material/issues/8312#issuecomment-3146419998) (also [documented here](https://squidfunk.github.io/mkdocs-material/reference/diagrams/#customization)), which will allow you to customize the  Mermaid.js integration as you need. We won't deepen our integration, since the API has proven to be too unstable in the past, which is why customization is the way to go ☺️

PR #7959, opened 2025-02-03
Attempt to fix outdated banner

squidfunk:
Should fix https://github.com/squidfunk/mkdocs-material/discussions/7947#discussioncomment-12039767.


squidfunk:
Reported to fix the issue in https://github.com/squidfunk/mkdocs-material/discussions/7947#discussioncomment-12041769.

PR #7961, opened 2025-02-03
Change absolute localhost URL to relative path

Paul2708:
This PR simply changes an absolute localhost URL to a relative path found in the latest blog post.


squidfunk:
Thanks!

PR #7969, opened 2025-02-05
Fix embedded mermaid css to fix class diagram arrow heads

willemkokke:
Originally used id based selecters, which doesn't work as the ids are made unique. Replacing them with class based selectors fixes this.

This original css tries to target this id, but as you can see it's prepended with `__mermaid_3_` path marker id

The new css directly targets the marker: marker selected

Using 'waybackmachine bisect' I can see it got introduced between:

- Issue not present: [Nov 21 2023](https://web.archive.org/web/20231121182633/https://squidfunk.github.io/mkdocs-material/reference/diagrams/#using-class-diagrams)

- Issue Appears: [Nov 27 2023](https://web.archive.org/web/20231127170019/https://squidfunk.github.io/mkdocs-material/reference/diagrams/#using-class-diagrams)

which leads to [these commits](https://github.com/squidfunk/mkdocs-material/commits/master/?since=2023-11-21&until=2023-11-27), the most likely culprit is the upgrade to mermaid 10.6.1 

Before: 
before

After:
 after

Tested on latest safari and chrome on macOS, latest edge and chrome on windows 11.

### DIFF
--- a/src/templates/assets/javascripts/components/content/mermaid/index.css
+++ b/src/templates/assets/javascripts/components/content/mermaid/index.css
@@ -171,19 +171,15 @@ g.classGroup text {
 }
 
 /* Class extension, composition and dependency marker */
-defs #classDiagram-extensionStart,
-defs #classDiagram-extensionEnd,
-defs #classDiagram-compositionStart,
-defs #classDiagram-compositionEnd,
-defs #classDiagram-dependencyStart,
-defs #classDiagram-dependencyEnd {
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR, and for isolating the update that broke it. LGTM!

squidfunk:
Released as part of [9.6.3](https://squidfunk.github.io/mkdocs-material/changelog/#9.6.3).

PR #7971, opened 2025-02-06
Fix faulty css in example

DeNelo:
The example showing how to add custom colors was missing a selector.

### DIFF
--- a/docs/setup/changing-the-colors.md
+++ b/docs/setup/changing-the-colors.md
@@ -340,7 +340,7 @@ add:
 === ":octicons-file-code-16: `docs/stylesheets/extra.css`"
 
     ``` css
-    :root {
+    :root  > * {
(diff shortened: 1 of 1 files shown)


iBug:
Doesn't seem necessary to me. The theme is already using `:root` selector so any custom settings also using the same selector should override the theme's variables, assuming custom CSS is always loaded *after* theme's defaults.

squidfunk:
Unfortunately, it's necessary due to issues with specificity. I haven't found another way to restructure the CSS to support CSS variable customization and different color schemes at the same time, but if there is, I'm all yours!

iBug:
Ah, got it. The point here is, elements have `[data-md-color-primary=whatever]` which has the same specificity as `:root`, and consequently (as a "direct match" selector) overrides those inherited from `:root`.

PR #7977, opened 2025-02-06
Add admonition for publishing guide

niclasheinz:
I made this PR because in some discussions people use the inbuilt preview server as a ‘production’ server. To prevent this abuse, I have clarified in this change that the image is only for previewing the website or for building it. 

But somehow I'm still not happy with it. Do you have any ideas on how to make this better @alexvoss?

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -4,11 +4,6 @@ The great thing about hosting project documentation in a `git` repository is
 the ability to deploy it automatically when new changes are pushed. MkDocs
 makes this ridiculously simple.
 
-???+ warning
-
-    Please note that the container image with which mkdocs-material is delivered is not an image for
-    publishing the documentation in a production environment. The container image supports the creation
-    of the documentation and has a built-in live preview server.
 
 
 ## GitHub Pages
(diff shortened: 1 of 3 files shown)


docs/publishing-your-site.md
```diff
@@ -4,6 +4,13 @@ The great thing about hosting project documentation in a `git` repository is
 the ability to deploy it automatically when new changes are pushed. MkDocs
 makes this ridiculously simple.
 
+???+ warning
+
+    Please note that the container image with which mkdocs-material is delivered is not an image for
```
squidfunk, line 9: IMHO, this is misleading, it sounds like you should not build the documentation, which is what the image is for. Also, I'm not sure whether this page is the right choice to add this admonition, it might be rather confusing.

squidfunk: Maybe rather under the Docker installation instructions in the getting started guide? Here's a draft:

> Please note that the Docker container is optimized for local previewing, not for deployments. The reason for this is that the web server that MkDocs provides for the live preview functionality is not production-grade, and might have security issues.

There might be some note on the MkDocs docs that mentions this, but I'm having a hard time finding it. There ar, however, a lot of issues where the maintainers of MkDocs have pointed this out to users trying to just run `mkdocs serve` in production.

niclasheinz:
Thanks for the review @squidfunk. I moved it to `Getting Started` at the end of the chapter  `with docker`.

squidfunk:
Great, LGTM! Thanks for investing the time ☺️

Issue #7976, opened 2025-02-06
Group plugin breaks if environmen value is "1", must be "true" to work


iBug:
### Bug description

Group plugin fails if the value is `1` (as shown in documentation). For it to work, the env var must be `true` instead.

```console
$ CI=1 mkdocs build
ERROR   -  Config value 'plugins': Plugin 'material/group' option 'enabled': Expected type: <class 'bool'> but received: <class 'int'>
```

### Related links

- [Reporting a bug](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/)
- [Built-in group plugin](https://squidfunk.github.io/mkdocs-material/plugins/group/)

### Reproduction

[9.6.2-group-enabled-type.zip](https://github.com/user-attachments/files/18694503/9.6.2-group-enabled-type.zip)

In case it matters, I'm on Python 3.12 (Ubuntu 24.04).

### Steps to reproduce

1. Create a `mkdocs.yml` with the following content:

    ```yaml
    site_name: test

    theme:
      name: material
    
    plugins:
      - info
      - group:
          enabled: !ENV CI
          plugins:
            - search
    ```

    The provided ZIP contains exactly this minimal reproduction.

2. Run `CI=1 mkdocs build`

### Before submitting

- [x] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
- [x] I have attached links to [the documentation](https://squidfunk.github.io/mkdocs-material/), and possibly related [issues](https://github.com/squidfunk/mkdocs-material/issues) and [discussions](https://github.com/squidfunk/mkdocs-material/discussions).
- [x] I assure that I have [removed all customizations](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/#remove-customizations) before submitting this bug report.
- [x] I have attached a __.zip file__ with a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/) using the [built-in info plugin](https://squidfunk.github.io/mkdocs-material/plugins/info/).

squidfunk:
[...]

PR #7980, opened 2025-02-07
doc(insiders/getting-started): Fix wrong env value type (int → bool)

iBug:
See #7976 for context. I was misled by this page.

### DIFF
--- a/docs/insiders/getting-started.md
+++ b/docs/insiders/getting-started.md
@@ -153,14 +153,14 @@ plugins:
   - search
   - social
 
-  # CI=1 mkdocs build
+  # CI=true mkdocs build
   - group:
       enabled: !ENV CI
       plugins:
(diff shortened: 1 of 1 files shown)


squidfunk:
Perfect, thanks. I'm astonished this wasn't a problem before, as this is a GitHub Actions build that uses Ubuntu.

iBug:
> as this is a GitHub Actions build that uses Ubuntu.

As you have probably realized, GitHub Actions sets `CI` to `true`, not `1`:

> image

squidfunk:
My point was, that we had it like this for ages, and millions of GitHub Actions builds seem to have succeeded without it being a problem that we used `CI=1` in our scripts. I'm just astonished that this surfaced now for the first time, and that your build was not working, where it seems to be working for everybody else.

PR #7989, opened 2025-02-10
doc(README): Use comma as thousand separator

iBug:
Same reasoning as #3688, just this time in README.


squidfunk:
Thanks, LGTM.

PR #7991, opened 2025-02-10
Github action pip cache: fix path, add year to cache key, resolves #7990

### LINKED ISSUE #7990: Improve caching in GitHub Actions workflow
### Description

I found that the Github action given to publish the site fails with error message `Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.` in the post run hook for `actions/cache@v4`. Currently the path specified to be cached is `.cache`. However, apparently the home directory is not the place the action is executed and thus the relative path fails to resolve. Further, it might be enough to cache the pip subdirectory. Hence, the path should be changed to include the home folder  `~/.cache/pip`.

I have tested the suggested change with the [BIMserver doc](https://github.com/bauinformatik/BIMserver-doc) and it can also be confirmed by looking at the examples in the [Github Actions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows) and the [Cache action](https://github.com/actions/cache/blob/main/examples.md#python---pip) documentation.

### Related links

- [Publish a site with Github actions](https://squidfunk.github.io/mkdocs-material/publishing-your-site/#with-github-actions)

### Proposed change

see PR #7991

### Before submitting

- [x] I have read and followed the [documentation issue guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-docs-issue/).
- [x] I have attached the links to the affected sections of [the documentation](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-docs-issue/#related-links)

hlg:


squidfunk:
Thanks for the PR. However, this cannot be merge due to the reasons stated in https://github.com/squidfunk/mkdocs-material/issues/7990#issuecomment-2649510292.

PR #8018, opened 2025-02-19
fix: large icons in TOC

alexvoss:
the icons in md:sponsors got rendered as very large icons as the typeset plugin is turned on in the production build


squidfunk:
KTHX! LGTM.

PR #8032, opened 2025-02-25
Support fallback from `regex` package to standard `re` module

### LINKED ISSUE #8031: Please provide a fallback from `regex` to the standard `re` module
### Context

I am maintaining Python packages for Gentoo Linux. [mkdocs-autorefs](https://pypi.org/project/mkdocs-autorefs/) package supports PyPy3, and it recently gained a dependency on mkdocs-material.

### Description

Some of the mkdocs-material plugins currently depend on the [regex](https://pypi.org/project/regex/) package. This package relies on CPython implementation details and has significant technical debt, and support for other Python implementations such as PyPy has been rejected upstream.

### Related links

- https://pypi.org/project/regex/#pypy
- https://github.com/mrabarnett/mrab-regex/issues/374#issuecomment-955863790

### Use Cases

It will enable full functionality of the plugin when using alternate Python implementations such as PyPy.

### Visuals

_No response_

### Before submitting

- [x] I have read and followed the [change request guidelines](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/).
- [x] I have verified that [my idea is a change request and not a bug report](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/#its-not-a-bug-its-a-feature).
- [x] I have ensured that, to the best of my knowledge, [my idea will benefit the entire community](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/#benefit-for-the-community).
- [x] I have included relevant links to [the documentation](https://squidfunk.github.io/mkdocs-material/), related [issues](https://github.com/squidfunk/mkdocs-material/issues), and [discussions](https://github.com/squidfunk/mkdocs-material/discussions) to underline the need for my idea.

mgorny:
Support using the standard Python library `re` module if the `regex` package is not available.  This improves the compatibility of plugins with environments where `regex` is not or cannot be installed, in particular when using PyPy.

The `regex` package is heavily relying on implementation details of CPython and has a significant technical debt, that is preventing the author from making the code more portable.  At the same time, the standard library `re` module provides most of the functionality used by mkdocs-material.

The `regex` package continues being used if it is installed in the environment, and the change should therefore remain backwards compatible.  When it is not installed, the `info` and `search` plugins will use the standard library `re` module, therefore it becomes possible to use them without having `regex` installed.  However, the Chinese search support via `jieba` requires `regex`, and therefore will be disabled if the fallback is used — much like it is currently disabled if `jieba` is not installed.

I could not find any formal declaration of `regex` or `jieba` requirements.  However, I did update the documentation to hint on the necessity of having both `jieba` and `regex` installed for the Chinese search to work.  I have also updated the old blog post — if that is undesirable, I can revert it.

Fixes #8031

### DIFF
--- a/docs/blog/posts/chinese-search-support.md
+++ b/docs/blog/posts/chinese-search-support.md
@@ -41,12 +41,12 @@ search plugin in a few minutes._
 ## Configuration
 
 Chinese language support for Material for MkDocs is provided by [jieba], an
-excellent Chinese text segmentation library. If [jieba] is installed, the
-built-in search plugin automatically detects Chinese characters and runs them
-through the segmenter. You can install [jieba] with:
+excellent Chinese text segmentation library. If [jieba] and [regex] packages
+are installed, the built-in search plugin automatically detects Chinese
+characters and runs them through the segmenter. You can install them with:
 
 ```
-pip install jieba
+pip install jieba regex
 ```
 
 The next step is only required if you specified the [`separator`][separator]
@@ -77,6 +77,7 @@ proficient in Chinese (yet?). If you find a bug or think something can be
 improved, please [open an issue].
 
   [jieba]: https://pypi.org/project/jieba/
+  [regex]: https://pypi.org/project/regex/
   [zero-width whitespace]: https://en.wikipedia.org/wiki/Zero-width_space
   [separator]: ../../plugins/search.md#config.separator
   [q=支持]: ?q=支持
--- a/docs/plugins/search.md
+++ b/docs/plugins/search.md
@@ -301,13 +301,15 @@ The following pipeline functions can be used:
 ### Segmentation
 
 The plugin supports text segmentation of Chinese via [jieba], a popular
-Chinese text segmentation library. Other languages like Japanese and Korean are
+Chinese text segmentation library, combined with extended regular expression
+support provided by [regex]. Other languages like Japanese and Korean are
 currently segmented on the client side, but we're considering to move this
 functionality into the plugin in the future.
 
 The following settings are available for segmentation:
 
   [jieba]: https://pypi.org/project/jieba/
+  [regex]: https://pypi.org/project/regex/
 
 ---
 
--- a/material/plugins/info/plugin.py
+++ b/material/plugins/info/plugin.py
@@ -23,7 +23,6 @@ import json
 import logging
 import os
 import platform
-import regex
 import requests
 import site
 import sys
@@ -40,6 +39,11 @@ from zipfile import ZipFile, ZIP_DEFLATED
 
 from .config import InfoConfig
 from .patterns import get_exclusion_patterns
+
+try:
+    import regex as re
+except ImportError:
+    import re
 
 # -----------------------------------------------------------------------------
 # Classes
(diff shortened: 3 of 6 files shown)


squidfunk:
> I'm not sure how deep Material delves into Regex features, but if they are mainly tapping into Unicode properties, something like the above package could likely provide sufficient functionality as a fallback. But if it delves deeper than that, then maybe not.

@facelessuser that's actually the only thing we're using it for:

https://github.com/squidfunk/mkdocs-material/blob/396c493f4774fa0bafeada5d0b555d02193557e8/src/plugins/search/plugin.py#L288

There's also a use in the info plugin, but I'm sure this can be replaced with `re`:

https://github.com/squidfunk/mkdocs-material/blob/396c493f4774fa0bafeada5d0b555d02193557e8/src/plugins/info/plugin.py#L413

Thus, it sounds like a better idea to scrap `regex` altogether and replace it with `bre` and `re`, ensuring that Chinese search still works. The functionality we need is "is there at least one Chinese character in the string", which is the current (albeit hacky) way of checking whether we need to segment the string. Please send a new PR for this, if it makes sense.

---

The current PR looks good so far, but requires a major release, given that Chinese users will now to explicitly have to install a new dependency. Thus, it will break their builds. Since this change request is primarily motivated to ease maintenance on a downstream project, and not something that we need to fix for _our_ users, we consider it low priority, and would release it together with the next major update. There's currently no ETA for a major release.

The approach mentioned above could be released as a patch release, if it's a fully functional replacement.

facelessuser:
@squidfunk As mentioned, `bre` will use `script_extensions` instead of `script` by default, so if absolute parity with what was used before is desired, or if you just want to reduce the chance of "something changing", then `expr = bre.compile(r"(\p{script: Han}+)", re.UNICODE)` would be identical. If that doesn't matter, then `expr = bre.compile(r"(\p{IsHan}+)", re.UNICODE)` is fine. I think regex is always using the latest Unicode, `bre` will use the same Unicode version as the Python release it is installed in.

>  Please send a new PR for this, if it makes sense.

I assume this is directed towards me?

squidfunk:
@facelessuser thanks for the clarification.

> I assume this is directed towards me?

It was directed towards the OP, but any help is appreciated. Maybe since you're the creator of `bre`, it's probably easiest for you, so if you can spare the time, it would be greatly appreciated ☺️ We could consider this a patch release then, right?

squidfunk:
@facelessuser I'm taking core of the info plugin right now.

squidfunk:
Replaced `regex` with `re` for the info plugin in 37162a57c. @kamilkrzyskow you originally used `regex` in your efforts to improve packing up the archive, but I don't think it's necessary, since only file paths are filtered against regular regex patterns that `re` should all know how to resolve:

https://github.com/squidfunk/mkdocs-material/blob/2d3982489da24b410b2258518957f808be3d57e9/material/plugins/info/patterns.py#L16-L26

Thus, I considered this change safe. If it isn't, please let me know ☺️

facelessuser:
@squidfunk, yeah, so I had some questions. Do any of these plugins take patterns they parse from the user? What I'm asking is basically which ones need to be Unicode property aware? Obviously, the Han line in question needs such behavior, but are there areas where user may have had access to Unicode properties and may still desire such? Or is it all internal.

facelessuser:
Screenshot 2025-02-25 at 8 22 49 PM

squidfunk:
It's all internal. As said, we only use it in the one place to detect whether user-supplied text (from Markdown) contains Chinese characters. That's it.

facelessuser:
Okay, so I'll just update everything to use Re only except that line that checks for Chinese chars.

facelessuser:
PR is up, I haven't had a chance to validate it.

facelessuser:
Seems to be working, take a look when you get a chance.

PR #8037, opened 2025-02-27
Lower backref version to support Py3.8

facelessuser:


squidfunk:
LGTM, discussion via https://github.com/squidfunk/mkdocs-material/pull/8034#issuecomment-2686548399.

Issue #8052, opened 2025-03-01
Sanitize aria-label on toc-items


reteps:
https://github.com/squidfunk/mkdocs-material/blob/922cb7cea295310c63ae623d350114e0df608734/src/templates/partials/toc-item.html#L33

Needs

```
| escape
```

Example breakage:

```
## "LTI 1.3" tab in your PrairieLearn course instance
```

renders to

```
<nav class="md-nav" aria-label="" LTI 1.3" tab in your PrairieLearn course instance">
```
which  breaks my HTML.

squidfunk:
Thanks for reporting. However, please stick to our [bug reporting process](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/). We're [closing as incomplete](https://squidfunk.github.io/mkdocs-material/contributing/#incomplete-issues-and-duplicates).

PR #8053, opened 2025-03-01
escape aria-labels to prevent breaking HTML markup

reteps:
- Bug fix for #8052 
- The TOC titles are not sanitized, causing issues with characters such as double quotes when rendering the HTML. I verified this by adding a `print(toc_tokens)` in `get_toc` of mkdocs.

Let me know if you require a formal reproduction (something I didn't provide in #8052).

### DIFF
--- a/src/templates/partials/toc-item.html
+++ b/src/templates/partials/toc-item.html
@@ -30,7 +30,7 @@
 
   <!-- Table of contents list -->
   {% if toc_item.children %}
-    <nav class="md-nav" aria-label="{{ toc_item.title | striptags }}">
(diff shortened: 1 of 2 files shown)


squidfunk:
> Let me know if you require a formal reproduction (something I didn't provide in https://github.com/squidfunk/mkdocs-material/issues/8052).

Before we can consider this PR, we need a proper issue, yes.

reteps:
Added a reproduction in #8054.

squidfunk:
Thanks, should be safe via https://github.com/squidfunk/mkdocs-material/issues/8054#issuecomment-2692637153.

PR #8055, opened 2025-03-02
getting started: mention also podman

niclasheinz:
With this PR, the documentation will also mention how to get started with podman.


squidfunk:
Thanks for the PR! However, please note that we can't commit to support each and every development runtime or environment, or package manager that's out there 😅 For instance, we also don't mention pipx, uv, etc. I have no experience with podman, but once we add it to our documentation, questions will start coming in which we can't guarantee to provide support for. Thus, I'm afraid we need to turn this down for economical reasons. I'm sorry.

PR #8058, opened 2025-03-03
Improve caching in GitLab Pipeline

niclasheinz:
Based on #7990 I have improved the caching in GitLab pipelines. Now the `.cache` folder is stored in the cache.

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -181,11 +181,16 @@ contents:
       script:
         - pip install mkdocs-material
         - mkdocs build --site-dir public
+      cache:
+        key: ${CI_COMMIT_REF_SLUG}
+        paths:
+          - .cache/
       artifacts:
         paths:
           - public
       rules:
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! I'm looping in @kamilkrzyskow here, maybe he can comment if he has the time. Historically, correct caching has been a tricky topic, so we need to get this right. I currently do not have the time to properly review this, but maybe we can reach a consensus in a discussion here, so we can be sure that it improves the situation. Prior art:

- https://github.com/squidfunk/mkdocs-material/pull/5463
- https://github.com/squidfunk/mkdocs-material/pull/5460
- https://github.com/squidfunk/mkdocs-material/pull/5448

squidfunk:
Ah wait, I've mistaken this for _our_ workflow 🤦‍♂️ Never mind, sorry for the noise. I guess it's good to merge.

squidfunk:
@kamilkrzyskow Thanks! Note that I've mistaken this for our workflow, which we need to revisit at some point, but no hurry.

PR #8070, opened 2025-03-08
Fixed favicon path

### LINKED ISSUE #8066: Favicon not displayed in Google Search Results
### Context

_No response_

### Bug description

While doing google search, the results pertaining to websites built with Mkdocs Material do not seem to have a favicon. **This problem is also happening with documentation pages hosted on https://squidfunk.github.io/**

Screenshots highlighting the issue with squidfunk.github.io as an example:

Image

Image

Incidentally, another person has created a clone of an older mkdocs-material (https://mrkeo.github.io/getting-started/) and their page links have the favicon in google search results (2nd screenshot above), so this seems to be a regression.

### Related links

- [Reporting a bug](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/)
-

### Reproduction

Added screenshots to the bug description. Issue is generic and also happens for https://squidfunk.github.io

Issue is happening on both desktop and mobile browsers.

### Steps to reproduce

1. Create a website with Mkdocs Material.
2. Do a google search so that content from the website matches with the search query.
3. Observe that the results do not have a favicon.

### Browser

_No response_

### Before submitting

- [x] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
- [x] I have attached links to [the documentation](https://squidfunk.github.io/mkdocs-material/), and possibly related [issues](https://github.com/squidfunk/mkdocs-material/issues) and [discussions](https://github.com/squidfunk/mkdocs-material/discussions).
- [x] I assure that I have [removed all customizations](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/#remove-customizations) before submitting this bug report.
- [x] I have attached a __.zip file__ with a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/) using the [built-in info plugin](https://squidfunk.github.io/mkdocs-material/plugins/info/).

squidfunk:
This PR ensures that the favicon path is absolute if `site_url` is set, and attempt to fix #8066.

### DIFF
--- a/material/templates/base.html
+++ b/material/templates/base.html
@@ -31,7 +31,12 @@
         <link rel="alternate" type="application/rss+xml" title="{{ lang.t('rss.created') }}" href="{{ 'feed_rss_created.xml' | url }}">
         <link rel="alternate" type="application/rss+xml" title="{{ lang.t('rss.updated') }}" href="{{ 'feed_rss_updated.xml' | url }}">
       {% endif %}
-      <link rel="icon" href="{{ config.theme.favicon | url }}">
+      {% if config.site_url %}
+        {% set favicon = config.site_url + config.theme.favicon %}
+        <link rel="icon" href="{{ favicon | url }}">
+      {% else %}
+        <link rel="icon" href="{{ config.theme.favicon | url }}">
+      {% endif %}
       <meta name="generator" content="mkdocs-{{ mkdocs_version }}, mkdocs-material-9.6.7">
     {% endblock %}
     {% block htmltitle %}
--- a/src/templates/base.html
+++ b/src/templates/base.html
@@ -77,7 +77,12 @@
       {% endif %}
 
       <!-- Favicon -->
-      <link rel="icon" href="{{ config.theme.favicon | url }}" />
+      {% if config.site_url %}
+        {% set favicon = config.site_url + config.theme.favicon %}
+        <link rel="icon" href="{{ favicon | url }}" />
+      {% else %}
+        <link rel="icon" href="{{ config.theme.favicon | url }}" />
+      {% endif %}
 
       <!-- Generator banner -->
       <meta


kamilkrzyskow:
This was the initial idea, taken from Google docs and W3Schools, however as mentioned later this site:
> - https://matteobaccan.github.io/owner/
  `<link rel="icon" type="image/x-icon" href="https://matteobaccan.github.io/owner/favicon.png" />`
  
Has been using the `site_url` and favicon combo for [12 years](https://github.com/matteobaccan/owner/blob/ebb604be8044dc63a70e17a6e7edab3fde112c98/owner-site/site/_includes/top.html#L16).

Therefore it's unlikely that it would help with the detection in general 😞 

> It looks like currently the root domain favicon or lack of it will be prioritized over everything 😞 

> So the current solution is to add the favicon on the root domain or add a redirection, which works on GMC, or does it...?

> GMC could also be using a cached icon from the time it did work and the redirection doesn't help on Google, just appears to help for now.

Apart of that the change LGTM. The docs could mention that using an user/organization level GitHub Pages https://pages.github.com/ additionally to the repository project page could help the most.

kamilkrzyskow (approved):

squidfunk:
Thanks! Let's wait for 2 weeks and then check our search results. If it did not help, we can close this PR.

niclasheinz:
Maybe I'm too stupid to understand the real problem here, but there is a [“default” favicon](https://squidfunk.github.io/mkdocs-material/setup/changing-the-logo-and-icons/#favicon) set in the docs that no longer exists in the repository. In the docs there is a link to the [default image ](https://github.com/squidfunk/mkdocs-material/blob/master/material/assets/images/favicon.png), which no longer exists. In my opinion, the correct favicon is [here](https://github.com/squidfunk/mkdocs-material/blob/master/material/templates/assets/images/favicon.png). Plus, I can't see any favicon in the [docs/assets ](https://github.com/squidfunk/mkdocs-material/tree/master/docs/assets/images) folder. 
Maybe this is just a problem with the documentation

kamilkrzyskow:
The documentation issue is separate, but you're right that the old link wasn't updated after the switch to using the `templates` directory.

In https://github.com/squidfunk/mkdocs-material/commit/56b96b4d799174d48d36613d70ffac9b964171cc the theme data was moved, but for the end-user the way it's accessed stayed the same, as the `templates` directory became the root for the theme.

> Plus, I can't see any favicon in the [docs/assets](https://github.com/squidfunk/mkdocs-material/tree/master/docs/assets/images) folder.

You won't find it there because it's exposed as part of the theme, which you linked above 
> the correct favicon is [here](https://github.com/squidfunk/mkdocs-material/blob/master/material/templates/assets/images/favicon.png).

The theme root works like customization:
- https://squidfunk.github.io/mkdocs-material/customization/#extending-the-theme

The files inside are merged with the final site build, the favicon is provided for all users by default.
You can see the favicon here, because the creators only configured the `logo`: 
- https://hewlettpackard.github.io/squest/

PR #8090, opened 2025-03-14
Prefer incremental pushes over force pushes for gh-deploy

Goooler:
It would be helpful for users to view changes between the commits on gh-deploy, and the "incremental" should be faster than "force".

### DIFF
--- a/docs/blog/posts/git-sparse-checkout.md
+++ b/docs/blog/posts/git-sparse-checkout.md
@@ -90,7 +90,7 @@ jobs:
         with:
           python-version: 3.x
       - run: pip install mkdocs-material
-      - run: mkdocs gh-deploy --force
+      - run: mkdocs gh-deploy
 ```
 
 ## Conclusion
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -35,6 +35,8 @@ contents:
         runs-on: ubuntu-latest
         steps:
           - uses: actions/checkout@v4
+            with:
+              fetch-depth: 0 # (3)!
           - name: Configure Git Credentials
             run: |
               git config user.name github-actions[bot]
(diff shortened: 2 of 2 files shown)


squidfunk:
Thanks for the PR. Those changes are pretty controversial. Could you please elaborate in more detail why you think they are necessary? We can't just merge a PR like this without knowing exactly what the implications are and what it aims to improve 😅 Thank you.

Goooler:
I migrated my project site from VUE to MkDocs (in in https://github.com/GradleUp/shadow/pull/1302) and copied the deployment steps from the documentation. Everything works well, but the `gh-deploy` branch was force-pushed using `--force`. This might be risky for users who are unaware of it, as they may want to preserve their `gh-deploy` commits. As you mentioned, this is a controversial topic. I don’t have a strong opinion on whether to merge this.

Goooler:
Thanks for your explanation!

kamilkrzyskow:
This is resolved, but I would like to add my 2 cents.

The `--force` flag is required to overwrite the commit history **if there is a conflict**, otherwise it would crash and there would be no deployment, no user would want to fix HTML source conflicts. It won't overwrite the history if there is no conflict and has no impact on performance, as `mkdocs gh-deploy` always does a full site build.

The `actions/checkout` action checks out (👀tongue twister) the branch which invoked the workflow (e.g. `main`), adding `fetch-depth: 0` maximizes the depth of loading the history, however it does so for the aforementioned branch it has no impact on `mkdocs gh-deploy`

Also adding `fetch-depth: 0` could greatly degrade performance in repositories with long histories and a lot of binary file changes.
- https://squidfunk.github.io/mkdocs-material/blog/2023/09/22/using-git-sparse-checkout-for-faster-documentation-builds/

PR #8093, opened 2025-03-14
Updated twemoji cdn for icon search index

kamilkrzyskow:
Errata to #8089

Before the merge I did run `mkdocs serve` locally, and saw that the new icons in the updated icon index do not show up, however I somehow assumed they're loaded from the deployed material .icons directory, so would show up after merging.

My assumption was of course wrong... as they're loaded from the CDN, which got updated over time, and this is the repository which Python Markdown Extensions use over `jsdelivr`, but I kept the raw github cdn here, because the website is hosted on github.io:
- https://github.com/facelessuser/pymdown-extensions/blob/cffc5847a0473f5d49278e05494d85530855b7b3/pymdownx/emoji.py#L39

Sorry about the PR churn, as per usual 😒


squidfunk:
No worries, LGTM!

PR #8099, opened 2025-03-16
Update "Creating your site" article to include -u option for "docker run" in Unix environment

iamdbychkov:
Hello!

Right now if you follow the instructions, CLI will create a file structure owned by root user and with write permission only for the owner of the file:

```shell
$ sudo docker run --rm -it -v ${PWD}:/docs squidfunk/mkdocs-material new .
INFO    -  Writing config file: ./mkdocs.yml
INFO    -  Writing initial docs: ./docs/index.md

$ ls -larth
total 16K
drwxrwxrwt 13 root     root     4.0K Mar 16 14:07 ..
-rw-r--r--  1 root     root       19 Mar 16 14:07 mkdocs.yml
drwxr-xr-x  2 root     root     4.0K Mar 16 14:07 docs
drwxr-xr-x  3 dbychkov dbychkov 4.0K Mar 16 14:07 .
```

which adds the necessity to change file's ownership as it is cannot be edited right away.

This can be avoided by using -u option in docker run command:

```shell
$ sudo docker run --rm -it -v ${PWD}:/docs -u $(id -u):$(id -g) squidfunk/mkdocs-material new .
INFO    -  Writing config file: ./mkdocs.yml
INFO    -  Writing initial docs: ./docs/index.md

$ ls -larth
total 16K
drwxrwxrwt 13 root     root     4.0K Mar 16 14:11 ..
-rw-r--r--  1 dbychkov dbychkov   19 Mar 16 14:11 mkdocs.yml
drwxr-xr-x  2 dbychkov dbychkov 4.0K Mar 16 14:11 docs
drwxr-xr-x  3 dbychkov dbychkov 4.0K Mar 16 14:11 .
```

Hence the proposed change

### DIFF
--- a/docs/creating-your-site.md
+++ b/docs/creating-your-site.md
@@ -10,7 +10,13 @@ mkdocs new .
 
 Alternatively, if you're running Material for MkDocs from within Docker, use:
 
-=== "Unix, Powershell"
+=== "Unix"
+
+    ```
+    docker run --rm -it -v ${PWD}:/docs -u $(id -u):$(id -g) squidfunk/mkdocs-material new .
+    ```
+
+=== "Powershell"
 
     ```
     docker run --rm -it -v ${PWD}:/docs squidfunk/mkdocs-material new .


niclasheinz:
@iamdbychkov thanks for your contribution

My thoughts on this are:

Your approach does not follow the instructions in the guide. The reason you need to use `sudo` is because your current user is not part of the `docker` group. If you add your user to the `docker` group, you should be able to run Docker commands without `sudo`, which would prevent the file ownership issue you had.

More information can you find [here](https://docs.docker.com/engine/install/linux-postinstall/)

iamdbychkov:
Hi! [niclasheinz](https://github.com/niclasheinz) 

Unfortunately adding user to docker group only allows you to not use `sudo` to run docker's commands, container's user will remain `root` and files will be created from root user

niclasheinz:
@iamdbychkov 

Yes, you are right. You need `sudo` rights, which can compromise security.

iamdbychkov:
no worries :) Thank you for your work!

PR #8105, opened 2025-03-21
Fixed pin for `mkdocs-git-revision-date-localized-plugin` plugin

zyv:
Hi,

Unfortunately the `~=` pins are a minefield and do not work the way you think. In #7845 you said:

>The current version range includes everything between `>=1.2.4,<2`.

However, `~=1.2` actually means `<=1.3.0`, so the current highest allowed plugin version is `1.3.0`. The latest version is at `1.4.5`, and it has a feature I need (git hashes), but I can't get it due to an overly restrictive pin.

Could you please accept the pin update? Thanks a lot!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ recommended = [
 ]
 git = [
   "mkdocs-git-committers-plugin-2>=1.1,<3",
-  "mkdocs-git-revision-date-localized-plugin~=1.2,>=1.2.4"
+  "mkdocs-git-revision-date-localized-plugin>=1.2.4,<2"
 ]
 imaging = [
   "pillow~=10.2",


squidfunk:
Thanks for the PR. Unless I'm missing something, that's not what PEP 440 says. See the section on [compatible releases](https://peps.python.org/pep-0440/#compatible-release):

> A compatible release clause consists of the compatible release operator ~= and a version identifier. It matches any candidate version that is expected to be compatible with the specified version.
>
> The specified version identifier must be in the standard format described in [Version scheme](https://peps.python.org/pep-0440/#version-scheme). Local version identifiers are NOT permitted in this version specifier.
>
> For a given release identifier V.N, the compatible release clause is approximately equivalent to the pair of comparison clauses:

Specifically:

```
~= 2.2
>= 2.2, == 2.*

~= 1.4.5
>= 1.4.5, == 1.4.*
```

Thus:

```
~= 1.2
>= 1.2, == 1.*
```

This should include 1.3.0.

squidfunk:
I can confirm that this works, as I just upgraded the plugin:

```
(venv) squidfunk@squidfunk mkdocs-material/community git:master $ pip install --upgrade mkdocs-git-revision-date-localized-plugin
Requirement already satisfied: mkdocs-git-revision-date-localized-plugin in ./venv/lib/python3.11/site-packages (1.2.2)
Collecting mkdocs-git-revision-date-localized-plugin
  Downloading mkdocs_git_revision_date_localized_plugin-1.4.5-py3-none-any.whl (25 kB)
...
```

zyv:
Thanks for the quick reply! Okay, it looks like you are right according to the docs:

```
~= 1.2
>= 1.2, == 1.*
```

So 1.4.5 should be installable. However, this only confirms my statement that `~=` is a minefield.

I wonder why I can't upgrade? I use `uv`, not `pip`, and `uv lock --upgrade` doesn't do it. Will have to investigate further :(

I guess the PR can be rejected, though it might be a good idea to accept it anyway, since it's already a second one.

zyv:
> I wonder why I can't upgrade? I use `uv`, not `pip`, and `uv lock --upgrade` doesn't do it. Will have to investigate further :(

Well, `uv sync --all-extras` did it. Sorry for the noise... As I said, you might want to change the pin, if only to avoid more misdirected PRs.

PR #8112, opened 2025-03-25
Make templates compatible with minijinja (Rust)

squidfunk:
This PR improves compatibility with [minijinja](https://github.com/mitsuhiko/minijinja), a template engine we're currently exploring. Additionally, it replaces several instances of Python function invocations with idiomatic use of template filters. All instances where variables have been mutated inside templates have been replaced. Most changes have been made in partials, and only a few in blocks, and all of them are fully backwards compatible, so no changes to your overrides are necessary.

Note that this PR does not replace the [jinja](https://jinja.palletsprojects.com/en/stable/) template engine with [minijinja](https://github.com/mitsuhiko/minijinja). However, our templates are now 99% compatible with minijinja, which means we can explore alternative jinja-compatible implementations. Additionally, immutability and removal of almost all Python function invocations means much more idiomatic templating ☺️

---

🙋‍♀️ __Want to help testing?__ Just check out this branch, and build your documentation with it – it only takes a minute:

```
pip install git+https://github.com/squidfunk/mkdocs-material.git@refactor/jinja-compat
mkdocs build
```

There should be no difference in output whatsoever. If there is, please let us know!

### DIFF
--- a/material/templates/base.html
+++ b/material/templates/base.html
@@ -58,8 +58,8 @@
     {% endblock %}
     {% block fonts %}
       {% if config.theme.font != false %}
-        {% set text = config.theme.font.get("text", "Roboto") %}
-        {% set code = config.theme.font.get("code", "Roboto Mono") %}
+        {% set text = config.theme.font.text | d("Roboto", true) %}
+        {% set code = config.theme.font.code | d("Roboto Mono", true) %}
         <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
         <link rel="stylesheet" href="https://fonts.googleapis.com/css?family={{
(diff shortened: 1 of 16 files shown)


squidfunk:
We've tested the changes on several big projects, and consider them safe to be merged. All changes are backward compatible, which means that no overrides need to be adapted. Only Jinja syntax has changed, but template semantics have stayed the same. The only thing that is different in behavior is that the JSON passed to the application will now include keys that are `null`, as minijinja does not allow to progressively construct dicts, as all values are immutable. Yet, this does not change behavior, it only adds a few more bytes while giving us the upside of almost 99% compatibility.

PR #8120, opened 2025-03-29
Update code-blocks.md

mrxsierra:
Added tip for code block annotation usage


squidfunk:
Thanks for the PR. However, this is clearly already explained with an example in the usage guide of the same document under the [adding annotations](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#adding-annotations) section. There's not need for duplication, and the configuration section should only contain configuration.

mrxsierra:
Sorry for trouble man! "I should scroll more" before making PR. Thanks for your response.

PR #8124, opened 2025-03-31
Update python base image to alpine3.21

rkoehn:
Updating the docker base image to alpine 3.21, as python dropped support for alpine 3.19.

See https://github.com/docker-library/python/commit/3d7b328b66525fe2e82af7063af10c176b6ee8cd


squidfunk:
Thanks, LGTM.

PR #8150, opened 2025-04-06

StarfallProjects:
No longer offering Digital Ocean or Vercel guides: just did a round of checks and updates. DO was never one I recommended highly, and the Vercel guide seems to need a complete re-write, so removing for now. I've refreshed the Netlify guide and that one is now nice and up to date at least . . .


squidfunk:
Thanks for the PR! Sad to see the go, but totally understandable. Could you please also updata the list of links?

https://github.com/squidfunk/mkdocs-material/blob/a86ec49225eae9e0e2f3c6e9cf4ae3542c96879a/docs/publishing-your-site.md?plain=1#L256-L261

StarfallProjects:
@squidfunk Done. I'll try and bring Vercel back at some stage, but figured it was better to remove until I could do it properly (it was VERY wrong at this point) I also have guides for Ionos and Statichost on my list, if they seem good when I try them out.

squidfunk:
Thanks! LGTM.

PR #8155, opened 2025-04-09
Add mention of custom domain when publishing site using GitHub Actions

mschoettle:
I ran into the issue that the `CNAME` file that GitHub creates when specifying the custom domain in the repo settings gets overwritten on the next deploy.

This PR adds a sentence and refers to the MkDocs docs that explain the addition of the `CNAME` file.

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -138,6 +138,8 @@ Page is set to `gh-pages`.
 
 Your documentation should shortly appear at `<username>.github.io/<repository>`.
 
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. Can we generalize this into "To publish your site on a custom domain, please refer to the MkDocs documentation"? This delegates the entire details to MkDocs, so if more things that are necessary to set up for custom domains appear in MkDocs' documentation, we don't need to change our guide. Thanks!

mschoettle:
Yes, that's a good idea. I like this better.

Issue #8153, opened 2025-04-09
Mermaid ER diagrams cardinality symbols are hard to see


Jazzinghen:
### Context

I am trying to add a couple of Mermaid ER diagrams into my documentation, however the relational cardinality symbols are very hard or impossible to see and if the theme is forced they show on top of the ER diagram node, instead of being half hidden behind each node.

### Bug description

ER diagrams are not rendered properly in MKDocs using Material theme, this can be seen from this project's documentation:

![Image](https://github.com/user-attachments/assets/5b1511ee-dcde-467d-b816-04aee51cdc8a)

While on the official Mermaid documentation the same diagram is rendered like this:

![Image](https://github.com/user-attachments/assets/46962840-1b77-430f-a55c-ae15649f456a)

In my case, this is how my diagram is rendered:

![Image](https://github.com/user-attachments/assets/a0afd01f-a085-4836-bad1-2b46d8de3044)

While on the official online editor it's shown like this:

![Image](https://github.com/user-attachments/assets/62832171-66e6-49cf-abf6-9c857b66442c)

If the theme is forced to be `dark` then the issue is a bit different:

![Image](https://github.com/user-attachments/assets/bbe555fe-f93e-45ce-8e84-9fad8ccfc7b7)

### Related links

- [MKDocs Material ER Diagrams](https://squidfunk.github.io/mkdocs-material/reference/diagrams/?h=mermaid#using-entity-relationship-diagrams)
- [Mermaid ER Diagrams](https://mermaid.js.org/syntax/entityRelationshipDiagram.html#entity-relationship-diagrams)

### Reproduction

[9.6.11-mermaid-er-broken-rendering.zip](https://github.com/user-attachments/files/19663450/9.6.11-mermaid-er-broken-rendering.zip)

### Steps to reproduce

1. Create a MKDocs Material documentation project
2. Set it up to use Mermaid as [per documentation](https://squidfunk.github.io/mkdocs-material/reference/diagrams/?h=mermaid#configuration)
3. [Optional] Set some theme options
4. Create a page with a Mermaid ER diagram using relationship cardinalities

### Browser

Firefox

### Before submitting

- [x] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
[...]

PR #8166, opened 2025-04-16
Fix rendering of mermaid ERD

amaline:
Fix the rendering of mermaid entity diagrams by removing the base .node configuration only for entity diagrams and fixing the relationship lines, marker stroke color, as well as the node labels.

This is for issue #8153


squidfunk:
Superseded by #8211. Thanks for digging into it! However, this PR is incomplete and includes changes that are too opinionated (e.g., setting node color to black regardless of color theme). Let's continue in #8211.

PR #8169, opened 2025-04-17
Remove useless section for customizing colors

pgiraud:
Setting the `primary` (or `accent`) colors to `custom` isn't required. The [example](https://github.com/mkdocs-material/examples/tree/master/examples/custom-colors) doesn't do it.

We can remove it.


squidfunk:
Thanks for reporting. We'll keep this setting, because if we remove it, and users still have `theme.palette.primary` set, the custom overrides don't take effect. For this reason, it is not useless, but a safe guard for us to reduce the number of reported issues. Thus, we'll keep it, but in general you're right, it is not necessary. It doesn't hurt either.

PR #8170, opened 2025-04-17
Reflect GitHub releases in changelog

Goooler:
image

### DIFF
--- a/docs/changelog/index.md
+++ b/docs/changelog/index.md
@@ -2,18 +2,18 @@
 
 ## Material for MkDocs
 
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. What information is in the GitHub release, so that it is valuable to link to? It's the exact same content, so there's not upside adding those links, IMHO.

Goooler:
It's handy to view the commit heads and keep the changelog closer to https://keepachangelog.com/en/1.1.0/

squidfunk:
Thanks for explaining! We're currently not considering this change, we're happy how it is ☺️

PR #8217, opened 2025-05-14
Add rel=edit link to edit button link

capjamesg:
This PR adds the rel="edit" attribute to the mkdocs material edit link.

rel=edit lets a page indicate that the linked resource can be used to edit the page. It is defined at https://microformats.org/wiki/rel-edit. This can then be parsed by tools like the Universal Edit Button and custom bookmarklets to open the edit page corresponding with a website.


squidfunk:
Thanks for the PR! However, the `rel` attribute is placed on the view button, not on the edit button. Could you change this to the edit button please?

squidfunk:
Thanks!

Issue #4277, opened 2022-08-24
Feedback widget is not loaded when navigating to another page with navigation.instant feature


vedranmiletic:
### Contribution guidelines

- [X] I've read the [contribution guidelines](https://github.com/squidfunk/mkdocs-material/blob/master/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] ... the problem doesn't occur with the `mkdocs` or `readthedocs` themes
- [X] ... the problem persists when all overrides are removed, i.e. `custom_dir`, `extra_javascript` and `extra_css`
- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

Feedback widget is not loaded when navigating to another page with `navigation.instant` feature. This works as expected without `navigation.instant` feature.

### Expected behaviour

It should be loaded normally.

### Actual behaviour

It is not loaded.

### Steps to reproduce

1. Have at least two pages.
2. Load the first one.
3. Navigate to the second one.
4. Observe the missing feedback widget.

### Package versions

- Python: 3.10.6
- MkDocs: 1.3.1
- Material: 8.4.1

### Configuration

```yaml
site_name: My Docs
theme:
  name: material
  features:
    - navigation.instant
extra:
  analytics:
    provider: google
    property: G-XXXXXXXXXX
    feedback:
      title: Was this page helpful?
      ratings:
        - icon: material/emoticon-happy-outline
          name: This page was helpful
          data: 1
          note: >-
            Thanks for your feedback!
        - icon: material/emoticon-sad-outline
          name: This page could be improved
          data: 0
          note: >- 
            Thanks for your feedback! Help us improve this page by
            using our <a href="..." target="_blank" rel="noopener">feedback form</a>.
```

### System information

- Operating system: Garuda Linux
- Browser: Brave

squidfunk:
Thanks for reporting. Clear oversight on my side, I forgot to test it with instant loading. Fixed in 46d06c1c1.

squidfunk:
Released as part of 8.4.2.

vedranmiletic:
Thank you!

PR #8275, opened 2025-06-23
Update documentation code snippet to ensure feedback widget displays after instant navigation

staadecker:
Previously, the feedback widget wouldn't display after instant navigation (see issue #4277). Although this was fixed, the documentation's example for implementing a [custom feedback handler](https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-analytics/#custom-site-feedback) still results in this issue. Please double check my PR as I blindly replicated the fix from #4277. This worked for me but javascript is not my forte.


squidfunk:
Thanks for the PR. Yes, that's definitely oversight, and the fix looks correct.

Issue #8284, opened 2025-06-27
[Bug]: Content tabs component rendering exception on narrow screen


SuperManito:
### Context

![Image](https://github.com/user-attachments/assets/381e7555-f0f0-4ecb-b875-2cd37c1d7d41)

![Image](https://github.com/user-attachments/assets/13697340-39b1-484e-ba89-a0162aeade81)

fix method

```css
.tabbed-set {
    overflow: hidden !important;
}
```

### Bug description

see picture

### Related links

- [Reporting a bug](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/)
-

see picture

### Reproduction

see picture

### Steps to reproduce

see picture

### Browser

Chrome

### Before submitting

- [x] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
- [x] I have attached links to [the documentation](https://squidfunk.github.io/mkdocs-material/), and possibly related [issues](https://github.com/squidfunk/mkdocs-material/issues) and [discussions](https://github.com/squidfunk/mkdocs-material/discussions).
- [x] I assure that I have [removed all customizations](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/#remove-customizations) before submitting this bug report.
- [x] I have attached a __.zip file__ with a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/) using the [built-in info plugin](https://squidfunk.github.io/mkdocs-material/plugins/info/).

squidfunk:
Thanks for reporting. The stretching of the content tabs to the entire screen is by design. Other than that I don't understand the problem. If you only give us "see picture" and no minimal reproduction, that's all we can do here.

SuperManito:
> Thanks for reporting. The stretching of the content tabs to the entire screen is by design. Other than that I don't understand the problem. If you only give us "see picture" and no minimal reproduction, that's all we can do here.

This is not normal. See [here](https://linuxmirrors.cn/use/#%E5%B8%B8%E8%A7%81%E9%97%AE%E9%A2%98)

![Image](https://github.com/user-attachments/assets/aaa9a86e-80e8-425b-bcd0-077b4de2a189)

SuperManito:
[...]

PR #8286, opened 2025-06-29
Updated info plugin to exclude guessed venv paths

kamilkrzyskow:
Hi 👋,
since the amount of cases with minimal reproduction containing an inactive Virtual Environment reached a comfortable threshold I decided to update it and detect/guess possible venv paths.

To guess the paths I use the `pyvenv.cfg` file, which is added in the default `python -m venv` call:
- https://github.com/python/cpython/blob/76dd0eee056fbb2a6dc5c3dbb52568ba61d04994/Lib/venv/__init__.py#L220

Perhaps other "venv" solutions don't do that, and a better alternative would be to use `site-packages` or look for "python" executables, but Linux might not use `python` but specifically some `python3.12` name variant, so I picked the simplest solution for now with the `pyvenv.cfg` :v:

[9.6.14-guess.zip](https://github.com/user-attachments/files/20968378/9.6.14-guess.zip)

```sh
$ mkdocs build
INFO    -  Started archive creation for bug report

Please name your bug report (2-4 words): guess
Possible inactive venv: C:\Users\...\material-info-venv-test\venv2
INFO    -  Archive successfully created:

  9.6.14-guess/docs/index.md 287.0 B
  9.6.14-guess/mkdocs.yml 61.0 B
  9.6.14-guess/platform.json 352.0 B
  9.6.14-guess/requirements.lock.txt 332.0 B

  9.6.14-guess.zip 1.6 kB
```

In the final `platform.json` I decided to hide the username as it's a potential breach of privacy, a bit late after months but better late than never :v:
`...` - in this case I obfuscated the username in the text here, however it would display for the user in the terminal.

Additionally, I added the output of the `VIRTUAL_ENV` environmental variable to make sure we see the activated venv and don't have to guess based on the `sys.path`

Ref:
- https://github.com/squidfunk/mkdocs-material/issues/8284#issuecomment-3015598360
- https://github.com/squidfunk/mkdocs-material/discussions/8198#discussioncomment-13042253

### DIFF
--- a/material/plugins/info/plugin.py
+++ b/material/plugins/info/plugin.py
@@ -282,6 +282,7 @@ class InfoPlugin(BasePlugin[InfoConfig]):
                             *sys.argv[1:]
                         ]),
                         "env:$PYTHONPATH": os.getenv("PYTHONPATH", ""),
+                        "env:$VIRTUAL_ENV": os.getenv("VIRTUAL_ENV", ""),
                         "sys.path": sys.path,
                         "excluded_entries": self.excluded_entries
                     },
--- a/material/plugins/info/plugin.py
+++ b/material/plugins/info/plugin.py
@@ -280,8 +280,13 @@ class InfoPlugin(BasePlugin[InfoConfig]):
                 ]))
             )
 
+            # Try to get login to replace it with USERNAME placeholder
+            try:
+                username = getpass.getuser()
+            except Exception:
+                username = ""
+
             # Add information on platform
-            # Replace login with USERNAME placeholder
             f.writestr(
                 os.path.join(example, "platform.json"),
                 json.dumps(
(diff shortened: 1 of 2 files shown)


squidfunk:
Thanks for the PR. As always, I trust you on this one, since you're the one championing most of the improvements and bug fixes in the plugin. We can fast track this by our feature freeze, since it's not a feature.

kamilkrzyskow:
Thanks, I did test on Windows and WSL, and it's working as I would expect. 🙂
It should work on macOS as well, but I didn't go the extra mile to test it on GitHub Actions this time around.

Minor bug was on WSL, with the username substitution, as WSL loads the name as lowercase and Windows itself uses uppercase for the directory. So I would have to replace it with regex to make it work in this scenario of mixed case between Windows and WSL, however I think that's a very rare scenario, so I didn't optimise for it. If somebody uses Docker with WSL they likely don't put files in any personal path either 🤔

squidfunk:
Released as part of [9.6.15](https://squidfunk.github.io/mkdocs-material/changelog/#9.6.15).

PR #8327, opened 2025-07-11
Replace .cache with ~/.cache in "Publishing your site" documentation

ChrisMayfield:
When using the GitHub Actions workflow provided in documentation, the following warning occurs:

```
Post job cleanup.
Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.
```

Adding `~/` to the path resolves this issue. For reference, see `Post Run actions/cache/@v4` in these logs:

* [Before the change][1] (`path: .cache`)
* [After the change][2] (`path: ~/.cache`)

[1]: https://github.com/cspogil/cspogil.github.io/actions/runs/16211472269/job/45772312586
[2]: https://github.com/cspogil/cspogil.github.io/actions/runs/16220862593/job/45800795438

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -46,7 +46,7 @@ contents:
           - uses: actions/cache@v4
             with:
               key: mkdocs-material-${{ env.cache_id }}
-              path: .cache # (4)!
+              path: ~/.cache # (4)!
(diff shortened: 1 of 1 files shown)


squidfunk:
Good catch, thanks!

Issue #8349, opened 2025-07-22
Allow .venv as folder name and not just venv when creating minimal reproductions using info plugin


schneekluth:
### Context

When using the [Python Extension](https://marketplace.visualstudio.com/items?itemName=ms-python.python) in vscode the command palette allows to easily create Virtual Environments but they have a fixed folder name `.venv/`.

<img width="612" height="214" alt="Image" src="https://github.com/user-attachments/assets/addc9c09-fc2b-4e4c-982c-0b354ba5e50a" />

### Description

Currently the [Info Plugin](https://squidfunk.github.io/mkdocs-material/plugins/info/) only respects folders for Virtual Environments named `venv/`. If you create a minimal reproduction example the created `zip` is less than a megabyte in size:

<img width="334" height="37" alt="Image" src="https://github.com/user-attachments/assets/ac27ede1-5330-43a5-aa86-d9645be96dfc" />

If the folder name is `.venv`  the content of this folder is included in the generated `zip` which results in a filesize of roughly 30 megabytes and a warning is shown:

<img width="881" height="115" alt="Image" src="https://github.com/user-attachments/assets/eee05eff-1770-4f64-8fcb-0c2c80f1a45c" />

Would it be possible to allow virtual environment folders to be names `.venv`?

### Related links

- [Info Plugin](https://squidfunk.github.io/mkdocs-material/plugins/info/)
- [Creating a reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/#environment)

### Use Cases

[Creating a virtual environment in vscode](https://code.visualstudio.com/docs/python/environments#_creating-environments) is way quicker and less prone to errors as using the CLI. vscode provides handy commands that even novice users can easily understand.

### Before submitting

- [x] I have read and followed the [change request guidelines](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/).
- [x] I have verified that [my idea is a change request and not a bug report](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/#its-not-a-bug-its-a-feature).
- [x] I have ensured that, to the best of my knowledge, [my idea will benefit the entire community](https://squidfunk.github.io/mkdocs-material/contributing/requesting-a-change/#benefit-for-the-community).
[...]

PR #8352, opened 2025-07-23
Fixed info venv guessing again

kamilkrzyskow:
Fixes: #8349 

Hi 👋 ,

this PR fixes the shortcomings of the previous #8286 PR, as it mainly adds the code that should've been there previously😅 
We can't use `glob.iglob` to look for hidden paths, as this was added in Python 3.11, so the best choice is `os.walk`

Change list:
- [x] .dotpath support in the venv guess logic
- [x] fixed empty username, str.replace("", "123") inputs 123 between every character
- [x] os.getcwd() usage stayed mostly as-is, because we want to support config file in another directory

However, the info plugin uses MkDocs paths, which could also have inconsistent drive letters. Therefore, I introduced a function, which slightly decreases the readability of the code, but hopefully there won't be too many issues ✌️

Works for me on both Windows and WSL paths.
Maybe @schneekluth could double check on both Windows and macOS

To install this branch run this command (edited)
```
pip install git+https://github.com/kamilkrzyskow/mkdocs-material.git@b1863de7271a76fadd4abbd18b9295845f6c5958#egg=mkdocs-material
```

### DIFF
--- a/material/plugins/info/plugin.py
+++ b/material/plugins/info/plugin.py
@@ -284,7 +284,7 @@ class InfoPlugin(BasePlugin[InfoConfig]):
             try:
                 username = getpass.getuser()
             except Exception:
-                username = ""
+                username = "USERNAME"
 
             # Add information on platform
             f.writestr(
--- a/material/plugins/info/plugin.py
+++ b/material/plugins/info/plugin.py
@@ -204,15 +204,16 @@ class InfoPlugin(BasePlugin[InfoConfig]):
         # Guess other Virtual Environment paths in case we forget to activate
         # them or in case we have multiple. Making sure which venv is activated
         # is not necessary, as it is an optional step in the guidelines.
-        for path in glob.iglob(
-            pathname = "**/pyvenv.cfg",
-            root_dir = os.getcwd(),
-            recursive = True
-        ):
-            path = os.path.join(os.getcwd(), os.path.dirname(path))
-            if path not in site.PREFIXES:
-                print(f"Possible inactive venv: {path}")
-                self.exclusion_patterns.append(_resolve_pattern(path))
+        for abs_root, dirnames, filenames in os.walk(os.getcwd()):
+            for filename in filenames:
+                if filename.lower() != "pyvenv.cfg":
+                    continue
+
+                path = abs_root[0].upper() + abs_root[1:]
+
+                if path not in site.PREFIXES:
+                    print(f"Possible inactive venv: {path}")
+                    self.exclusion_patterns.append(_resolve_pattern(path))
 
         # Exclude site_dir for projects
         if projects_plugin:
--- a/material/plugins/info/plugin.py
+++ b/material/plugins/info/plugin.py
@@ -151,16 +151,22 @@ class InfoPlugin(BasePlugin[InfoConfig]):
         if not isinstance(loaded_configs, list):
             loaded_configs = [loaded_configs]
 
+        # It can happen that the drive letter case is inconsistent on Windows.
+        # Therefore, assure first character to be uppercase for the following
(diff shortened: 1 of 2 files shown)


kamilkrzyskow:
I'm not sure if the edit `@` mention works, so I'll do another one :v: @schneekluth

schneekluth:
I just tried and it worked. The folder `.venv` was not included in the generated `zip` file. Good Job @kamilkrzyskow 

grafik

Content of `platform.json`:

```json
{
  "system": "Windows-11-10.0.26100-SP0",
  "architecture": [
    "64bit",
    "WindowsPE"
  ],
  "python": "3.13.5",
  "cwd": "C:\\Users\\USERNAME\\Desktop\\mkdocs-issue-reproduction",
  "command": "mkdocs build",
  "env:$PYTHONPATH": "",
  "env:$VIRTUAL_ENV": "C:\\Users\\USERNAME\\Desktop\\mkdocs-issue-reproduction\\.venv",
  "sys.path": [
    "C:\\Users\\USERNAME\\Desktop\\mkdocs-issue-reproduction\\.venv\\Scripts\\mkdocs.exe",
    "C:\\Users\\USERNAME\\AppData\\Local\\Programs\\Python\\Python313\\python313.zip",
    "C:\\Users\\USERNAME\\AppData\\Local\\Programs\\Python\\Python313\\DLLs",
    "C:\\Users\\USERNAME\\AppData\\Local\\Programs\\Python\\Python313\\Lib",
    "C:\\Users\\USERNAME\\AppData\\Local\\Programs\\Python\\Python313",
    "c:\\Users\\USERNAME\\Desktop\\mkdocs-issue-reproduction\\.venv",
    "c:\\Users\\USERNAME\\Desktop\\mkdocs-issue-reproduction\\.venv\\Lib\\site-packages"
  ],
  "excluded_entries": [
    "^/.venv/ - /.venv/",
    "^/site/ - /site/",
    "^/venv/ - /venv/"
  ]
}
```

---

Just for completeness: There seems to be a typo in the initial URL you posted:

```
pip install git+https://github.com/kamilkrzyskow/mkdocs-material.git@b1863de7271a76fadd4abbd18b9295845f6c5958#egg=material
```

should be

```
pip install git+https://github.com/kamilkrzyskow/mkdocs-material.git@b1863de7271a76fadd4abbd18b9295845f6c5958#egg=mkdocs-material
```

kamilkrzyskow:
Thanks, it wasn't a typo, but you're right that `material` isn't enough.
I always assumed the `egg` refers to the root entry point directory, so in this case `material`.
Now I need to check other places I provide such info, and reexamine the value of egg.

EDIT:

Ok, my confusion came from other packages that are called `some-package` and have the `some_package` directory name.
In example `pip install` to specific commits, I've seen `egg=some_package`, so I understood it as the name of the directory / what you import.

Instead `egg` resolves the package name that is being installed, and the resolved name is being normalized:
https://peps.python.org/pep-0503/#normalized-names

So anything would work `mkdocs-material`, `mkdocs_material`, `mkdocs.material`, but not `material` 😄

kamilkrzyskow:
@squidfunk The PR is ready ✌️ 

The previous "guess" logic didn't check .dotpaths, which is less than ideal, because there might be more cases like explained here that use `.venv` as their default from VS Code / `uv`, but later they would follow the [reproduction guide](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/) and create `venv`.

So `venv` would be activated, `.venv` would not be detected, so we would be in the same boat.
Now it should work 🙏

squidfunk:
Perfect, thank you for investing the time! I currently have no time to check this in depth, but we fix errors with the info plugin as they appear, since it's so entangled with the environment, so bug are hard to predict anyway.

Issue #6999, opened 2024-04-01
Update Serbian translations


stcksmsh:
### Translations

```Jinja
{% macro t(key) %}{{ {
  "language": "sr",
  "action.edit": "Ажурирај страницу",
  "action.skip": "Иди на текст",
  "action.view": "Погледајте изворни код ове странице",
  "announce.dismiss": "Не приказуј поново",
  "blog.archive": "Архива",
  "blog.categories": "Категорије",
  "blog.categories.in": "у",
  "blog.continue": "Настави са читањем",
  "blog.draft": "Нацрт",
  "blog.index": "Назад на индекс",
  "blog.meta": "Метаподаци",
  "blog.references": "Сродне везе",
  "clipboard.copy": "Копирај у клипборд",
  "clipboard.copied": "Ископирано у клипборд",
  "consent.accept": "Прихвати",
  "consent.manage": "Управљај подешавањима",
  "consent.reject": "Одбиј",
  "footer": "Подножје",
  "footer.next": "Следеће",
  "footer.previous": "Претходно",
  "header": "Заглавље",
  "meta.comments": "Коментари",
  "meta.source": "Извор",
  "nav": "Навигација",
  "readtime.one": "1 минут читања",
  "readtime.other": "# минута читања",
  "rss.created": "РСС феед",
  "rss.updated": "РСС феед ажурираног садржаја",
  "search": "Претрага",
  "search.placeholder": "Претрага",
  "search.share": "Дељење",
  "search.reset": "Очисти",
  "search.result.initializer": "Иницијализујем претрагу",
  "search.result.placeholder": "Унесите појам претраге",
  "search.result.none": "Ништа није пронађено",
  "search.result.one": "1 резултат претраге",
  "search.result.other": "# резултата претраге",
  "search.result.more.one": "још 1 на овој страни",
  "search.result.more.other": "још # на овој страни",
  "search.result.term.missing": "Недостаје",
  "select.language": "Изабери језик",
  "select.version": "Изабери верзију",
  "source": "Иди у репозиторијум",
  "source.file.contributors": "Доприноситељи",
  "source.file.date.created": "Креиран",
  "source.file.date.updated": "Ажуриран",
  "tabs": "Табови",
  "toc": "Садржај",
  "top": "Назад на врх"
}[key] }}{% endmacro %}
```

### Country flag

:flag_rs:

### Before submitting

[...]

PR #3009, opened 2021-09-08
Improve serbian translation (cyr and latin)

majkinetor:
This commit provides improved translation for Serbo-Croatain (sh) and Serbian (sr) languages.

I added translation for missing fields (such as recent _Back to top_) and changed translation of few another fields to make it more consistent.

Issue #835, opened 2018-08-03
New translation: Serbian


taocnrut:
## Instructions

1. Check, if your language is already available: [here](http://bit.ly/2DCzaL0)
2. If it isn't, please translate the labels on the right:

``` jinja
{% macro t(key) %}{{ {
  "language": "en",
  "clipboard.copy": "Копирај у клипборд",
  "clipboard.copied": "Ископирано у клипборд",
  "edit.link.title": "Уреди страницу",
  "footer.previous": "Претходно",
  "footer.next": "Следеће",
  "meta.comments": "Коментари",
  "meta.source": "Извор",
  "search.language": "sr",
  "search.placeholder": "Претрага",
  "search.result.placeholder": "Унесите појам претраге",
  "search.result.none": "Нису пронађени документи",
  "search.result.one": "1 резултат претраге",
  "search.result.other": "# резултата претраге",
  "search.tokenizer": "[\s\-]+",
  "skip.link.title": "Иди на текст",
  "source.link.title": "Иди у репозиторијум",
  "toc.title": "Садржај"
}[key] }}{% endmacro %}
```

Thanks!

squidfunk:
Thank you for your contribution! Added both in fd75fa8eb71f1ffa8282b36dced8948de00fc0fa and will be part of the next release.

squidfunk:
Released as part of 3.0.0.

PR #8445, opened 2025-09-14
Update Serbian languages

rammba:
Hello @squidfunk, as a native Serbian speaker and contributor to few open-source projects (eg. [official React documentation](https://github.com/reactjs/sr.react.dev)) I wanted to improve Serbian translations a bit.
Of course, I'm kindly asking previous translators @taocnrut (#835),  @majkinetor (#3009) & @stcksmsh (#6999) to take a look and give their opinion about the changes.

Another thing is regarding currently supported [Serbo-Croatian (`sh`) language](https://en.wikipedia.org/wiki/Serbo-Croatian) which is deprecated and unused I suppose. Is there a way to actually add Serbian Latin (as proposed in #1480) instead of that? Maybe I should move this discussion to the issue?


src/templates/partials/languages/sh.html
```diff
@@ -23,22 +23,23 @@
 <!-- Translations: Serbo-Croatian -->
 {% macro t(key) %}{{ {
   "language": "sh",
+  "direction": "ltr",
```
squidfunk, line 26: This is not needed – it's the default and given by the `en` translation, please remove.

src/templates/partials/languages/sr.html
```diff
@@ -23,9 +23,10 @@
 <!-- Translations: Serbian -->
 {% macro t(key) %}{{ {
   "language": "sr",
+  "direction": "ltr",
```
squidfunk, line 26: See above.

squidfunk (changes requested):
Thanks for the PR – only minor remarks.

squidfunk:
> Another thing is regarding currently supported [Serbo-Croatian (sh) language](https://en.wikipedia.org/wiki/Serbo-Croatian) which is deprecated and unused I suppose. Is there a way to actually add Serbian Latin (as proposed in https://github.com/squidfunk/mkdocs-material/pull/1480) instead of that? Maybe I should move this discussion to the issue?

Could you please expand on that? Why is it _deprecated_? Also, we currently only support HTML5 language codes, which the language code mentioned in the issue is not. We might turn the direction around in the future (allow to provide any translations and map back into HTML language), but that's currently not a priority.

rammba:
> Could you please expand on that? Why is it deprecated?

It's not actually depracated, but it's an old version of current languages which were used mostly in former Yugoslavia. I'm fine with using it if there is no official code (as mentioned in [this comment](https://github.com/squidfunk/mkdocs-material/pull/1480#issuecomment-594126534)) for Serbian with Latin script.

squidfunk:
> It's not actually depracated, but it's an old version of current languages which were used mostly in former Yugoslavia. I'm fine with using it if there is no official code (as mentioned in https://github.com/squidfunk/mkdocs-material/pull/1480#issuecomment-594126534) for Serbian with Latin script.

I can't really decide that, because I don't speak any kind of Serbian. Maybe the previous translators can comment, so we can reach a decision if that is a sensible path to take. Other than that it's not possible for me/us to judge, and please keep in mind that this might break expectations of users that currently use the code.

squidfunk (approved):
Looks good to merge now, thanks again!

PR #8447, opened 2025-09-14
Replace codecs.open() with open() since codecs.open() deprecated in 3.14

### LINKED ISSUE #8446: codecs.open DeprecationWarning when using Python 3.14
### Context

When building documentation using MkDocs with the Material Theme in virtual environment using Python 3.14.0rc2, I got the following warning:

```sh
INFO    -  DeprecationWarning: codecs.open() is deprecated. Use open() instead.
             File ".venv/lib/python3.14/site-packages/material/extensions/emoji.py", line 53, in to_svg
               el.text = md.htmlStash.store(_load(icons[shortname]["path"]))
             File ".venv/lib/python3.14/site-packages/material/extensions/emoji.py", line 68, in _load
               with codecs.open(file, encoding = "utf-8") as f:
```

This is the line causing the problems in [emoji.py](https://github.com/squidfunk/mkdocs-material/blob/master/material/extensions/emoji.py#L68):

```py
def _load(file: str):
    with codecs.open(file, encoding = "utf-8") as f:
        return f.read()
```

Proposed solution, just replace `codecs.open` with `open` and remove the import for `codecs`:
```py
def _load(file: str):
    with open(file, encoding = "utf-8") as f:
        return f.read()
```

I don't understand why your code tree duplicates files, but the file appears to be located at these two locations:
- material/extensions/emoji.py
- src/extensions/emoji.py

### Bug description

`codecs.open` is deprecated as of Python 3.14 and `open` should be used instead.

### Related links

- [codecs.open is officially deprecated in Python 3.14](https://docs.python.org/3.14/library/codecs.html)

### Reproduction

`mkdocs build -s` using latest version of MkDocs and Material Theme with Python 3.14

### Steps to reproduce

`mkdocs build -s` using latest version of MkDocs and Material Theme with Python 3.14

Note: I tried building the zip-file you want, but couldn't.  This is a very straightforward issue with a clear 1 or 2 line solution.

### Browser

_No response_

### Before submitting

- [x] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
- [x] I have attached links to [the documentation](https://squidfunk.github.io/mkdocs-material/), and possibly related [issues](https://github.com/squidfunk/mkdocs-material/issues) and [discussions](https://github.com/squidfunk/mkdocs-material/discussions).
- [x] I assure that I have [removed all customizations](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/#remove-customizations) before submitting this bug report.
- [x] I have attached a __.zip file__ with a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/) using the [built-in info plugin](https://squidfunk.github.io/mkdocs-material/plugins/info/).

tleonhardt:
Replace `codecs.open()` with `open()` since the former is [deprecated in Python 3.14](https://docs.python.org/3.14/library/codecs.html#codecs.open)

`codecs.open` was necessary in Python 2 because the built-in `open()` function could not handle different text encodings like UTF-8. It provided a way to read and write text files containing Unicode characters, which the default open() could not do without causing errors.  With Python 3, the built-in `open()` function was changed to take an encoding argument and to handle text files as Unicode by default

Also:
- Added `uv.lock` to `.gitignore` to make it easier to test with uv virtual environments using things like a release candidate version of Python

Closes #8446

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,6 @@ tmp
 # IDEs & Editors
 .idea
 *~
+
+# uv Python package manager
+uv.lock
--- a/material/extensions/emoji.py
+++ b/material/extensions/emoji.py
@@ -20,7 +20,6 @@
 
 from __future__ import annotations
 
-import codecs
 import functools
 import material
 import os
@@ -65,7 +64,7 @@ def to_svg(
 # Load icon
 @functools.lru_cache(maxsize = None)
 def _load(file: str):
-    with codecs.open(file, encoding = "utf-8") as f:
+    with open(file, encoding = "utf-8") as f:
         return f.read()
 
 # Load twemoji index and add icons
--- a/src/extensions/emoji.py
+++ b/src/extensions/emoji.py
@@ -20,7 +20,6 @@
 
 from __future__ import annotations
 
-import codecs
 import functools
 import material
 import os
@@ -65,7 +64,7 @@ def to_svg(
 # Load icon
 @functools.lru_cache(maxsize = None)
 def _load(file: str):
-    with codecs.open(file, encoding = "utf-8") as f:
+    with open(file, encoding = "utf-8") as f:
         return f.read()
 
 # Load twemoji index and add icons


squidfunk:
Thanks for the PR. The code was adapted from @facelessuser's extension. Maybe he can shortly comment on why `codecs` was necessary in the first place? Does this still work in Python 3.8?

tleonhardt:
@squidfunk I updated the description with this text explaining that Python 2 required using `codecs.open` for unicode files, but in Python 3 this is supported out-of-the-box with the built-in `open` function:
>codecs.open was necessary in Python 2 because the built-in open() function could not handle different text encodings like UTF-8. It provided a way to read and write text files containing Unicode characters, which the default open() could not do without causing errors. With Python 3, the built-in open() function was changed to take an encoding argument and to handle text files as Unicode by default

And yes this still works with I believe all versions of Python 3, certainly including 3.8.

tleonhardt:
@squidfunk I updated the PR description with this text explaining that Python 2 required using codecs.open for unicode files, but in Python 3 this is supported out-of-the-box with the built-in open function:

codecs.open was necessary in Python 2 because the built-in open() function could not handle different text encodings like UTF-8. It provided a way to read and write text files containing Unicode characters, which the default open() could not do without causing errors. With Python 3, the built-in open() function was changed to take an encoding argument and to handle text files as Unicode by default

tleonhardt:
For `cmd2` that I help maintain, we had to make an identical change a couple months ago when we began testing on Python 3.14: https://github.com/python-cmd2/cmd2/pull/1470

squidfunk:
Thanks, that makes sense! Python 2, wow, that feels distant. I think this should be good to merge then.

tleonhardt:
@squidfunk Thank you for creating and maintaining such a wonderful documentation system! I've been using it for a bit under a year now after moving to it from Sphinx and I really love it.

facelessuser:
I don't know exactly what MkDocs Material relies on, but I will note that `codecs.open()` is not exactly equivalent when using `open()`. `open(..., errors='strict')` would be equivalent. It may be that Material does not require this the "strict" parameter, but something to note that I discovered in another project where I did require strict error handling in relation to encoding, etc. Just an FYI, especially if you notice different behavior.

squidfunk:
@facelessuser perfect, thanks for noting! If we discover problems along the way, we can add the strict parameter.

PR #8455, opened 2025-09-22
Carefully worded link to pre-release banner discussion

staadecker:


squidfunk:
Thanks for the PR! However, as mentioned in https://github.com/squidfunk/mkdocs-material/discussions/5803#discussioncomment-14477520, we do not want to endorse this _right now_. As also said in this comment, we're working on making such overrides much simpler in the future as part of our [transformational work](https://squidfunk.github.io/mkdocs-material/blog/2024/08/19/how-were-transforming-material-for-mkdocs/).

PR #8494, opened 2025-10-16
Make timeout a variable

baumkuchen42:
Made the default timeout variable that was previously defined in source code a proper config variable. This way, users of the plugin can increase the timeout if their servers are too slow for mkdocs


squidfunk:
Thanks for the PR! We're currently in feature freeze. We'll evaluate this PR once we finish our [foundational work](https://github.com/squidfunk/mkdocs-material/discussions/8461).

squidfunk:

PR #8502, opened 2025-10-30
docs: use a CSS variable for custom admonition color

jhermann:
Shows how you can make the admonition color easily customizable via a CSS variable (similar to the icon).


squidfunk:
Thank for the PR. [Relative RGB values](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb#browser_compatibility) are a recent addition and do not match the [browser range](https://squidfunk.github.io/mkdocs-material/browser-support/) we commit to support. Thus, we're currently not considering advertising this on our documentation. Thank you for your understanding.

PR #8522, opened 2025-11-06
Fix some urls

Echo-Nie:
fix some 404 urls

image


squidfunk:
Thanks for the PR! However, the social card links are broken after you replaced them? I've fixed the lunr.js link in 9853cc3a1.

PR #8537, opened 2025-11-13
Fix two url

Echo-Nie:
There are two broken urls.
https://pmarsceill.github.io/just-the-docs/
https://fontawesome.com/docs/web/setup/upgrade/whats-changed#icons-renamed-in-version-6


squidfunk:
Thanks! Just the Docs was not correct, but I fixed it.

Issue #8523, opened 2025-11-06
⚠️ We're now in maintenance mode


squidfunk:
> [!WARNING]
>
> __Material for MkDocs is now in maintenance mode.__ This means:
>
> - We will continue to provide __critical bug fixes__ and __security updates__ for __12 months__, at least
> - We will not add new features to Material for MkDocs, as Zensical is our focus from now on
> - We recommend switching to [Zensical] once it has all features you depend on

  [Zensical]: https://zensical.org  

### Official announcement

Before continuing, we recommend reading our announcement blog post. It explains what [Zensical] is, why we chose to build it, why you might consider switching to it, compatibility with Material for MkDocs, our new offering for professional users, and our vision for it:

→ [Read the full announcement on our blog](https://squidfunk.github.io/mkdocs-material/blog/2025/11/05/zensical/)

### Transition from Material for MkDocs

We designed Zensical to be compatible with Material for MkDocs. We recommend reading the following resources:

- [How Zensical is compatible with Material for MkDocs](https://zensical.org/compatibility/)
- [How you can build your Material for MkDocs projects with Zensical now](https://zensical.org/compatibility/configuration/)
- [Which features of Material for MkDocs are already supported](https://zensical.org/compatibility/features/) (spoiler: __a whole lot__)
- [Which plugins from the MkDocs ecosystem will be natively supported](https://zensical.org/compatibility/plugins/)
- [What MkDocs 2.0 means for your documentation projects](https://squidfunk.github.io/mkdocs-material/blog/2026/02/18/mkdocs-2.0/)

__You can still use Material for MkDocs__ – we will provide automated tools to make the jump to Zensical as simple as possible.

### Stay in the loop

- You can star and watch the [Zensical GitHub repository](https://github.com/zensical/zensical) to receive updates
- You might consider subscribing to [the Zensical newsletter](https://zensical.org/about/newsletter/) to stay updated
- You might consider following Zensical on [X](https://x.com/zensical), [BlueSky](https://bsky.app/profile/zensical.org) or [Fosstodon](https://fosstodon.org/@zensical)

> [!NOTE]
>
> __Using Material for MkDocs in a professional context?__
>
> If documentation plays a critical role in your organization, and you're worried how this might affect your business, or to get a tour of [Zensical Spark](https://zensical.org/spark/), feel free to reach out to [hello@zensical.org](mailto:hello@zensical.org).

### Changes to the issue tracker

[...]

PR #8542, opened 2025-12-01
Privacy plugin: more accurate pattern to capture external links

emaheuxPEREN:
Use a more accurate pattern to capture external links in privacy plugin.

Fix is two-fold:
- avoid false-positive for internal links that starts with `http`
- avoid false-negative for scheme-less links, e.g. `//x.y.z/`

### DIFF
--- a/src/plugins/privacy/plugin.py
+++ b/src/plugins/privacy/plugin.py
@@ -70,8 +70,8 @@ class PrivacyPlugin(BasePlugin[PrivacyConfig]):
         self.assets = Files([])
         self.assets_done: list[File] = []
         self.assets_expr_map = {
-            ".css": r"url\(\s*([\"']?)(?P<url>http?[^)'\"]+)\1\s*\)",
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR. Please provide a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/#reproduction), so we can observe the errors this PR is trying to fix + decide whether it is worth integrating. Please note that this project is now in [maintenance mode](https://github.com/squidfunk/mkdocs-material/issues/8523).

emaheuxPEREN:
Here you are : 
[9.7.0-privacy-external-links-pr8542.zip](https://github.com/user-attachments/files/24100118/9.7.0-privacy-external-links-pr8542.zip)

squidfunk:
Thanks for providing the reproduction, looks good!

squidfunk:
Another fix was necessary to make protocol-relative URLs work. I'm not sure why this worked on your machine, but requests failed with being able to fetch the URL. Fixed in 79ba428ab.

PR #8550, opened 2025-12-29
fix: Gitlab CI configuration was updated for site publishing

carloseguevara:
From version 17.10 the gitlab pages have a different configuration way:

> In GitLab 17.10 and later, for Pages jobs only, the public directory is appended automatically to [artifacts:paths](https://docs.gitlab.com/ci/yaml/#artifactspaths) when the [pages.publish](https://docs.gitlab.com/ci/yaml/#pagespublish) path is not explicitly specified

Reference: https://docs.gitlab.com/user/project/pages/getting_started/pages_from_scratch/#specify-the-public-directory-for-artifacts

Also, a more simple way (following the documentation) is just define `pages: true` but for me is more clear in this way, instead to know the default value reading the documentation. The match between `--site-dir public` and `pages: publish: public` is more intuitive.

### DIFF
--- a/docs/publishing-your-site.md
+++ b/docs/publishing-your-site.md
@@ -135,9 +135,8 @@ pages:
     key: ${CI_COMMIT_REF_SLUG}
     paths:
       - ~/.cache/ # (1)!
-  artifacts:
-    paths:
-      - public
+  pages:
+    publish: public
   rules:
     - if: '$CI_COMMIT_BRANCH == $CI_DEFAULT_BRANCH'
 ```


squidfunk:
Thanks for the PR. So, the old way still works? Additionally, as I understand, the changes require the new version of GitLab (Im not a user)? If so, we're not intending to merge this, because we're in [maintenance mode](https://github.com/squidfunk/mkdocs-material/issues/8523), and don't change things that might make more work for us, e.g., when user's don't realize they need to upgrade to the latest version to get going.

carloseguevara:
Got it! I wasn't aware of the maintenance mode.

GitLab supports the last three major versions (currently 18.x, 17.x, and 16.x). However, this applies only to Self-Managed users (Premium and Ultimate customers). You can find more information [here](https://about.gitlab.com/support/statement-of-support/) .

Consequently, for any other users who do not use Self-Managed (the majority of them), they are on the latest major version, and the current code example doesn't work.

squidfunk:
> Consequently, for any other users who do not use Self-Managed (the majority of them), they are on the latest major version, and __the current code example doesn't work.__

Thanks for the explanation. Okay, then we should merge this.

PR #8552, opened 2026-01-01
Update optimize.md with version requirement note

ab12gu:
Added note about Material for MkDocs version requirement. When trying to use plugin, building material errors out stating no optimize plugin found (not super helpful). The error probably occurs a lot because the release was just a couple months ago.

I might look into updating the error code later in build, but good to note in the documentation at least :).

### DIFF
--- a/docs/plugins/optimize.md
+++ b/docs/plugins/optimize.md
@@ -103,6 +103,8 @@ and mentions some alternative environments.
   [built-in plugins]: index.md
   [image processing]: requirements/image-processing.md
 
+And, ensure that Material for MkDocs version is at least the minimum listed above.
+
 ### General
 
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! We already list the version for each and every feature in each section (see the 9.7.0 below the headline). The paragraph this PR adds doesn't add more value, so please understand that we're not considering adding it.

Bildschirmfoto 2026-01-02 um 09 29 36

ab12gu:
Yeah thats fine. I do understand that you have the version number above that does imply it and it is a little annoying to have overly verbose docs. 

I'll see if there is a way to update the error code during build later (more useful than being directly in the docs)

Thanks for quick response tho!

PR #8561, opened 2026-01-27
fix: source file (raw/blob) view button generates invalid file path

pencilnav:
Discoved in https://github.com/Atlas-OS/docs/issues/281 

> ### [BUG]: Page Source Link is Wrong on Text Editors
> 
> On the [Text Editors page](https://docs.atlasos.net/getting-started/post-installation/software/text-editors/), clicking the view source link leads to a 404. The link looks to be changed from text-editors.md to text-rawors.md with edit -> raw.
> 

## fix: source file (raw/blob) view button generates invalid file path

This PR is to fix the gen logic for the view buttons, which wrongfully converted all text in the **docs edit link** with the word `edit` to `raw`. PR tested locally with `mkdocs-serve` and confirm fixes the issue.

e.g. for http://docs.atlasos.net/post-install/software/text-editors/

#### Edit link: 

github.com/Atlas-OS/docs/**edit**/main/docs/post-install/software/text-**edit**ors.md

#### What its supposed to convert to: 

github.com/Atlas-OS/docs/**raw**/main/docs/post-install/software/text-**edit**ors.md

#### But it was presented as:

github.com/Atlas-OS/docs/**raw**/main/docs/post-install/software/text-**raw**ors.md

### Affected files:
https://github.com/squidfunk/mkdocs-material/blob/0d11a2d01174a0ab3bec97300c4432da44128253/src/templates/partials/actions.html#L47

### DIFF
--- a/src/templates/partials/actions.html
+++ b/src/templates/partials/actions.html
@@ -39,17 +39,17 @@
   <!-- View button -->
   {% if "content.action.view" in features %}
     {% if "/blob/" in page.edit_url %}
-      {% set part = "blob" %}
+      {% set part = "/blob/" %}
     {% else %}
-      {% set part = "edit" %}
-    {% endif %}
-    <a
-      href="{{ page.edit_url | replace(part, 'raw') }}"
-      title="{{ lang.t('action.view') }}"
-      class="md-content__button md-icon"
-    >
-      {% set icon = config.theme.icon.view or "material/file-eye-outline" %}
-      {% include ".icons/" ~ icon ~ ".svg" %}
-    </a>
+      {% set part = "/edit/" %}
+    {% endif %}      
+      <a
+        href="{{ page.edit_url | replace(part, '/raw/') }}"
+        title="{{ lang.t('action.view') }}"
+        class="md-content__button md-icon"
+      >
+        {% set icon = config.theme.icon.view or "material/file-eye-outline" %}
+        {% include ".icons/" ~ icon ~ ".svg" %}
+      </a>
   {% endif %}
 {% endif %}


squidfunk:
Thanks for the PR! Material for MkDocs is in [maintenance mode](https://github.com/squidfunk/mkdocs-material/issues/8523), and we're only considering merging and releasing critical bugfixes. I understand that this is causing you trouble, but it can be fixed with a single [template override](https://squidfunk.github.io/mkdocs-material/customization/#overriding-partials), so we don't consider it critical. Please also consider reading the announcement on [why we're discontinuing Material for MkDocs](https://squidfunk.github.io/mkdocs-material/blog/2025/11/05/zensical/).

If you wish to follow up, feel free to send a [PR to Zensical](https://github.com/zensical/ui), so we can fix it there. Zensical currently uses the same structure of templates as Material for MkDocs ☺️

pencilnav:
IMO putting the project in maintenance mode should still prioritize user experience (in comparison it should be way easier then letting someone to change their template override), especially under the condition where zensical isn't yet providing the same level of functionality with mkdocs's plugins.

It may be small, but it could be annoying. At least consider keeping it open and if a future “critical bugfix" happens, merge it together. Or add a notice on your docs on how to fix it (using template override).

Not all docs writers are coders.

squidfunk:
Thanks for sharing your opinion! This behavior has been [in place for 4 years](https://github.com/squidfunk/mkdocs-material/commit/c4d61cdc419838c0c245f63ab6f26a6ca863eddc), and it came up just now – it can't be critical, or it would've been brought up earlier. Thus, we don't consider it critical. An easy and stable workaround exists, since we will not change the templates during maintenance mode. From our perspective, any time invested into the dying MkDocs toolchain is a waste of effort. I kindly ask you to respect our decision.

pencilnav:
It would be nice to keep this open, and if there is some other PR that fixes critical issues (which requires releasing a new version), merge this together with that. 
Or a small notice on your docs mentioning how to fix this. (If my previous wordings aren't clear im sorry)

But anyways, i respect your decision.

PR #8564, opened 2026-02-18
Added blog post on MkDocs 2.0

squidfunk:


bonastreyair:
Why not also limiting the requirements.txt with `mkdocs>=1.6,<2`?

This would mean that your package will make sure that even if mkdocs 2.0 gets released, we would all be safe and no upgrade will ever happen thanks to dependency resolution and your safetynet of compatibility.

WDYT @squidfunk ?

squidfunk:
@bonastreyair Thanks for the suggestion. When we put Material for MkDocs into [maintenance mode](https://github.com/squidfunk/mkdocs-material/issues/8523), we decided to [open up the version ranges of all dependencies](https://github.com/squidfunk/mkdocs-material/commit/b583ea7765ea770505cafea5c0ca931e12289ad1#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552), so that we're not limiting what users could upgrade to.

However, the latest developments over at our central dependency, MkDocs, might be cause for concern, which is why we just limited the version range to `<2` in 10258334e. From our understanding of the situation, it is highly unlikely that there will be any more significant updates to MkDocs 1.x, and MkDocs 2.0 will be an entirely new tool of its own.

Issue #8567, opened 2026-02-19
New version with banner has side effect on other mkdocs plugin, at least mike


jylenhofdecat:
### Context

With the release of https://github.com/squidfunk/mkdocs-material/releases/tag/9.7.2
you added a warning banner that has side effect on mike mkdocs plugin I use

You're using standard stdout to print the banner and this banner is added to its own output...

### Bug description

On mike, I use `mike list` to list versions managed by mike... and now in its output in addition of its standard output there is the warning which breaks my script qui parses the json output

```
│ ⚠ WARNING – MkDocs 2.0 is incompatible with Material for MkDocs 
 │  
 │  MkDocs 1.x is unmaintained. We recommend switching to Zensical, our 
 │  new static site generator, as soon as possible. We're providing an 
 │  analysis of the situation in this article: 
 │  
 │  https://squidfunk.github.io/mkdocs-material/blog/2026/02/18/mkdocs-2.0/ 

[ 
 { 
 "version": "v0.2.1", 
 "title": "v0.2.1", 
 "aliases": [] 
 }, 
 { 
 "version": "v0.1.0", 
 "title": "v0.1.0", 
 "aliases": [] 
 } 
]
```

### Related links

- [Reporting a bug](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/)
-

### Reproduction

See output

### Steps to reproduce

Update the mkdocs-material with last version

### Before submitting

- [x] I have read and followed the [bug reporting guidelines](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/).
- [x] I have attached links to [the documentation](https://squidfunk.github.io/mkdocs-material/), and possibly related [issues](https://github.com/squidfunk/mkdocs-material/issues) and [discussions](https://github.com/squidfunk/mkdocs-material/discussions).
- [x] I assure that I have [removed all customizations](https://squidfunk.github.io/mkdocs-material/contributing/reporting-a-bug/#remove-customizations) before submitting this bug report.
- [x] I have attached a __.zip file__ with a [minimal reproduction](https://squidfunk.github.io/mkdocs-material/guides/creating-a-reproduction/) using the [built-in info plugin](https://squidfunk.github.io/mkdocs-material/plugins/info/).

squidfunk:
[...]

PR #8569, opened 2026-02-26
Document `NO_MKDOCS_2_WARNING=1` environment variable

johnthagen:
Follow on to

- #8564

And as suggested at

- https://github.com/squidfunk/mkdocs-material/issues/8567#issuecomment-3927761209

This is the first step to informing people how to silence the warning during a `build` or `serve`.


squidfunk:
Thanks for the follow-up! Please understand, that listing this at the top is not what we want here – we want users to be aware of what's coming, and showing them the bypass too early is not a good idea. It defeats the purpose. Thus, let's please move it further down, and put it into an admonition, maybe a tip? I'd say moving it above "The story behind Zensical" section is the best position, so we still raise awareness for MkDocs 2.0 and Zensical.

johnthagen:
@squidfunk Sure!

How does 75b33c4 look?

squidfunk:
LGTM, thanks!

Issue #8478, opened 2025-10-02
Live reload in `mkdocs serve` is broken in version 9.6.21


Skaronator:
### Bug description

After upgrading the mkdocs-material Docker image from `ghcr.io/squidfunk/mkdocs-material:9.6.20` to `ghcr.io/squidfunk/mkdocs-material:9.6.21`, the live-reloading feature has stopped working.

When running `mkdocs serve --dev-addr=0.0.0.0:7070`, the server starts and serves the site correctly. However, it no longer detects changes to source files (like Markdown files or `mkdocs.yml`). Consequently, the site does not rebuild, and the browser page does not auto-refresh.

To confirm this was not a container-specific issue, I performed A/B testing with a local Python virtual environment. The same bug occurs when using `mkdocs-material==9.6.21` installed via pip, while version `9.6.20` works as expected.

My environment is Windows 11 with WSL2 (Debian 13), with all files and commands being executed within the Linux environment.

### Related links

N/A

### Reproduction

This is basically just 1:1 the original guide. No special changes required.

[9.6.21-mkdocs-serve-no-longer-watches-for-filechanges.zip](https://github.com/user-attachments/files/22656989/9.6.21-mkdocs-serve-no-longer-watches-for-filechanges.zip)

### Steps to reproduce

1. Run `mkdocs serve --dev-addr=0.0.0.0:7070`
2. Modify index.md
3. Watch Log

### with `mkdocs-material==9.6.21`

```bash
mkdocs serve --dev-addr=0.0.0.0:7070                           
INFO    -  Building documentation...
INFO    -  Cleaning site directory
INFO    -  Documentation built in 0.22 seconds
INFO    -  [11:38:35] Serving on http://0.0.0.0:7070/
```

Nothing happens when modifing my index.md

### with `mkdocs-material==9.6.20`

```bash
mkdocs serve --dev-addr=0.0.0.0:7070                           
INFO    -  Building documentation...
INFO    -  Cleaning site directory
INFO    -  Documentation built in 0.21 seconds
INFO    -  [11:37:24] Watching paths for changes: 'docs', 'mkdocs.yml'
INFO    -  [11:37:24] Serving on http://0.0.0.0:7070/
INFO    -  [11:37:28] Detected file changes
INFO    -  Building documentation...
INFO    -  [11:37:28] Reloading browsers
INFO    -  [11:37:30] Detected file changes
INFO    -  Building documentation...
INFO    -  [11:37:30] Reloading browsers
[...]

PR #8582, opened 2026-04-16
add `--livereload` flag to `mkdocs serve` command

Aevarkan:
Caused me a bit of pain when I was contributing documentation to this project here:
- https://github.com/EndstoneMC/endstone/pull/381#issue-4276183999

I believe it'd be best to not give too much explanation, as to not lead anyone down a rabbit hole, and instead include this flag in the base command.

### DIFF
--- a/docs/creating-your-site.md
+++ b/docs/creating-your-site.md
@@ -196,7 +196,7 @@ write your documentation. The server will automatically rebuild the site upon
 saving. Start it with:
 
 ``` sh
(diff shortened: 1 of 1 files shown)


squidfunk:
Thanks for the PR! More than half a year after this bug was first reported in #8478, it's still not fixed upstream, so I agree - let's make it simple, not sending everybody down this rabbit hole.

PR #8583, opened 2026-04-22
docs: add section-wide comments example using .meta.yml

MMoney1988:
## Summary
Adds a small example to the comment system documentation showing how to enable comments for a whole folder using `.meta.yml` and how to override that on a specific page.

## Context
The page already shows single-page comment setup and links to the meta plugin, but not a compact example for this specific folder-wide use case. This is a small clarification for a common usage pattern.

## Scope
- documentation-only change
- no behavior change
- no restructuring
- no feature addition


squidfunk:
Thanks for the PR!

PR #8589, opened 2026-05-12
Update document for lists, to mention sane_lists extension

MaddyGuthridge:
This is similar to #7373, but far less broad. Instead of adding the extension by default, or giving an enormous amount of detail, it just mentions the existence of the extension.

This should provide a far better compromise between preventing users from option-hunting, and ensuring that `mkdocs-material` doesn't face increased maintenance burden by not implying stronger support for this extension than is actually offered.

I think the existence of this option should be documented, if only to prevent time wasted hunting through issues and discussions with GitHub's awful search UI.

### DIFF
--- a/docs/reference/lists.md
+++ b/docs/reference/lists.md
@@ -102,6 +102,14 @@ be re-numbered when rendered:
 
 </div>
 
+If you don't want your lists to be automatically re-numbered, you should enable
+the [`sane_lists` markdown extension](https://python-markdown.github.io/extensions/sane_lists/).
+
+```yml
+markdown_extensions:
+  - sane_lists
+```
+
 ### Using definition lists
 
 When [Definition Lists] is enabled, lists of arbitrary key-value pairs, e.g. the


squidfunk:
Thanks for the PR. We don't consider updating the docs in maintenance mode, unless absolutely necessary. I hope you can understand, that we're not interested in increasing the maintenance burden, and updating the docs to promote new options is the wrong signal. The MkDocs ecosystem is dead.

> This should provide a far better compromise between preventing users from option-hunting, and ensuring that mkdocs-material doesn't face increased maintenance burden by not implying stronger support for this extension than is actually offered.

[Zensical](https://zensical.org) will make hunting for extensions and plugins obsolete. This is our way forward.

MaddyGuthridge:
Thanks, I wasn't aware of [the absolute state of mkdocs](https://squidfunk.github.io/mkdocs-material/blog/2026/02/18/mkdocs-2.0/). This and the fact that `mkdocs-material` is in maintenance mode could be worth documenting on the `README.md` of this project, to make it more visible.

squidfunk:
Thanks for noting! We have a non-dismissable banner in our documentation since November 5 announcing Zensical. The issue tracker mentions that this project is in maintenance mode, including all releases. Recent releases of Material for MkDocs print a warning to the terminal about MkDocs 2.0. How could you miss it? 😉

MaddyGuthridge:
Ohhhhhhhhh it's probably that I haven't needed to update any documentation backend stuff (just been editing markdown files) for the docs of all my projects, so the warnings have been getting swallowed by CI. The one repo where I have been running `mkdocs` locally hasn't been updated for a few months, since I moved it to my workplace GitLab, meaning dependabot hasn't been doing it's thing (need to switch it to renovate I think).

As for how I missed the banner, I think I saw it and thought it was an ad, rather than a warning of the deprecation. Perhaps rephrasing it to say something like "`mkdocs-material` is in maintenance mode. Learn more about this decision." would make it clear that this info is important for anyone reading the documentation.

lovelydinosaur:
> Thanks, I wasn't aware of [the absolute state of mkdocs](https://squidfunk.github.io/mkdocs-material/blog/2026/02/18/mkdocs-2.0/).

You might want to take that with more than a little pinch of salt...

> I think I saw it and thought it was an ad

That's.... entirely right, yep.

The breathless nonsense in "[What MkDocs 2.0 means for your documentation projects](https://squidfunk.github.io/mkdocs-material/blog/2026/02/18/mkdocs-2.0/)" is the Zensical team hyping up their product with a bit of red-hatting silliness.

I didn't think that was how we operated in the open-source world. And yet, here we are.

PR #8590, opened 2026-05-13
Fixed typo: formely → formerly

mokashang:
Fixes a small typo (`formely` → `formerly`) in the deprecation notices for the built-in `projects` and `typeset` plugins.


squidfunk:
Thanks!

PR #8601, opened 2026-07-28
ci: attach provenance and SBOM attestations to the published Docker image

kobihikri:
Hi, and thanks for Material for MkDocs.

Your PyPI side is already covered and I want to say so first, because it is why this PR is narrow. The `python` job publishes through `pypa/gh-action-pypi-publish` with `id-token: write`, so Trusted Publishing is in use and PEP 740 attestations are being produced — I checked the current release against PyPI's integrity endpoint and it returns a real attestation bundle:

```
$ curl -s -o /dev/null -w '%{http_code}' \
    https://pypi.org/integrity/mkdocs-material/9.7.7/mkdocs_material-9.7.7-py3-none-any.whl/provenance
200
```

The Docker image is the half that has nothing. The "Publish Docker image" step pushes to both Docker Hub and `ghcr.io` on release, and those manifests carry no provenance or SBOM.

The change is two lines on that one step:

```yaml
      - name: Publish Docker image
        uses: docker/build-push-action@v6
        with:
          context: .
          platforms: ${{ env.PLATFORMS }}
          push: ${{ github.event_name == 'release' }}
          provenance: mode=max
          sbom: true
```

I put it only on the publish step, not the earlier "Build Docker image" step — that one is `load: true` for the local `Check Docker image` run, so attesting it would cost build time for something never pushed.

BuildKit attaches the attestation to the manifest, so both registries get it from the single build, and **no permissions change is needed**. Your `Check manifest` step afterwards will show the extra attestation manifest in the index, which is expected.

Two caveats: `mode=max` records build args (`provenance: true` is the smaller option), and the extra manifest is handled by Docker Hub and GHCR but worth knowing about if a mirror sits in front.

No SLSA level claimed.

Disclosure: I used AI assistance to help spot this and prepare the change, and I checked the PyPI attestation state and the workflow myself.


kobihikri:
**Correction — I got a fact wrong in this PR, and I would rather flag it myself than let it sit.**

I wrote that the pushed manifest "carries no provenance or SBOM attestation". That is half wrong, and the wrong half matters.

**Provenance is already there.** For *public* repositories, `docker/build-push-action` adds provenance attestations with `mode=max` by default — [Docker's documentation](https://docs.docker.com/build/ci/github-actions/attestations/) states it plainly: *"Public repos: provenance attestations with `mode=max` are automatically added"*. I checked published images and they do already carry attestation manifests. So the `provenance: mode=max` line in my diff makes existing behaviour **explicit**; it does not add anything new.

**The SBOM is genuinely new.** That part stands — the same page says *"SBOM attestations aren't automatically added to the image"*, and `sbom: true` is what enables them.

I also wrote in the caveats that `provenance: true` gives "a smaller record". That is wrong as well: `true` resolves to max on a public repo, and the smaller setting is `provenance: mode=min`.

**So the honest description of this PR is:** it adds an SBOM attestation, and pins the provenance mode explicitly instead of relying on the default. Both are still defensible — an explicit line means the behaviour will not change quietly if the default ever does — but it is a smaller change than my description implied, and you should judge it on that basis rather than on what I originally wrote.

Happy to retitle and rewrite the description accordingly, or to close this if the SBOM alone is not worth the diff to you. Either is fine — just say which and I will act on it.

Apologies for the inaccuracy. It was caught by a maintainer reviewing the same change on another project, and they were right to.

squidfunk:
Thanks for the PR! Material for MkDocs is in maintenance mode and will be [EOL (End of Life) on November 5 this year](https://github.com/squidfunk/mkdocs-material/issues/8523), so in a little more than 3 months. At this time, we do not intend to make changes to our publishing workflows, since it's unclear whether there will be any more releases at all. We recently issued a fix for a CVE, but once EOL, there will be no more security update. For this reason, we don't consider changing proven workflows to be a good invest of our time.

I hope that's understandable from a maintainer's perspective. We recommend to either switch to a maintained fork, or check out our new static site generator [Zensical](https://squidfunk.github.io/mkdocs-material/blog/2025/11/05/zensical/) that is in large parts already compatible with Material for MkDocs.

